PR #190, opened 2015-05-29
add group update

tamarrow:
Also add 'Created deployment ' message to all updates.


tamarrow:
thanks for the review!

PR #209, opened 2015-06-05
fix analytics to use actual subcommand

mgummelt:
bug fix


ssk2:
:+1: merge once built successfully

PR #212, opened 2015-06-05
truncated test output

mgummelt:
This is trivial, but some of the test output is so long, it's hard to get back up to the stack trace in my shell.  This truncates it.


tamarrow:
lgtm

PR #213, opened 2015-06-05
add HOST column to task table

mgummelt:


dcos/mesos.py
```diff
@@ -305,17 +305,27 @@ def dict(self):
...
         :rtype: dict
         """
+
         return self._task
 
     def framework(self):
         """Returns the task's framework
 
-        :returns" task's framework
+        :returns: task's framework
```
ssk2: Good catch.

dcos/mesos.py
```diff
@@ -305,17 +305,27 @@ def dict(self):
...
         :rtype: Framework
         """
 
         return self._master.framework(self["framework_id"])
 
+    def slave(self):
+        """Returns the task's slave
```
ssk2: Just to make this obvious - can we say `Returns the task's slave ID`?

ssk2:
Looks fine, merge when it builds!

PR #216, opened 2015-06-08
Update readme to use docs site instead of beta-docs.

ssk2:


ssk2:
Merge after new site goes liev.

PR #218, opened 2015-06-10
[ADD] DCOS-1002  Create a TeamCity integrated test for Windows version of DCOS CLI

YuraTolochkevich:


ssk2:
Let's squash all of these commits please :)

YuraTolochkevich:
@ssk2  Thanks, ready

swartzrock:
Squashed, tests passed. Sunil can you please merge if it looks good? LGTM.

mgummelt:
Out of curiosity, is anyone actually using Windows?

cli/tests/integrations/common.py
```diff
@@ -30,7 +30,7 @@ def exec_command(cmd, env=None, stdin=None):
         stderr=subprocess.PIPE,
         env=env)
 
-    stdout, stderr = process.communicate()
+    stdout, stderr = [_.replace(b'\r', b'') for _ in process.communicate()]
```
mgummelt: underscore is for variables you don't use.  you use this one.  name it something else

bin/clean.ps1
```diff
@@ -0,0 +1,5 @@
+﻿$BaseDir = (Get-Location).Path
```
mgummelt: Let's put these windows build files in their own subdirectory

mgummelt:
For the existing tests, our entry point was bin/start_tests.sh  What is the entry point for this new test suite?

ssk2:
@mgummelt, a not-insignificant proportion of enterprise users use Windows, including some of Mesosphere's customers. It was an important requirement for the CLI.

cli/tests/integrations/test_analytics.py
```diff
@@ -115,64 +173,6 @@ def test_config_reporting_false():
         assert requests.post.call_count == 0
 
 
-@_mock
-def test_production_setting_true():
```
mgummelt: If you've just moved these, please move them back so we know nothing has changed

mgummelt:
Which customers?

cli/tests/integrations/test_config.py
```diff
@@ -12,20 +12,24 @@
 
 @pytest.fixture
 def env():
-    return {
+    r = os.environ.copy()
```
mgummelt: why this change?

cli/tests/integrations/common.py
mgummelt: Please add a comment here describing why you're doing this.

cli/tests/integrations/test_dcos.py
```diff
@@ -71,9 +71,11 @@ def test_version():
 
 
 def test_missing_dcos_config():
-    env = {
+    env = os.environ.copy()
```
mgummelt: why this change?

cli/tox.win.ini
```diff
@@ -0,0 +1,41 @@
+[tox]
```
mgummelt: why do we need a separate tox for windows?

dcos/package.py
```diff
@@ -943,6 +944,25 @@ def copy_to_cache(self, target_dir):
                 'Unable to fetch packages from [{}]'.format(self.url))
 
 
+def onerror(func, path, exc_info):
```
mgummelt: please follow the pydoc documentation format we use for all other functions.  Also, if this function is private to this module, prefix it with a `_`

mgummelt: why did you create this function?

dcos/package.py
```diff
@@ -985,8 +1005,14 @@ def validate(self):
 
         # TODO(CD): implement these checks in pure Python?
         scripts_dir = os.path.join(self._base_path, 'scripts')
-        validate_script = os.path.join(scripts_dir, '1-validate-packages.sh')
-        result = subprocess.call(validate_script)
+        suffix = 'ps1' if util.is_windows_platform() else 'sh'
+        validate_script = os.path.join(scripts_dir,
+                                       '1-validate-packages.' + suffix)
+        cmd = [validate_script]
```
mgummelt: do an if..else here to be more functional

cli/tests/integrations/common.py
ytalashkocv: This is needed to get rid of '\r' from Windows's lines endings.

cli/tox.win.ini
ytalashkocv: Because Windows requires additional library pypiwin32

cli/tests/integrations/test_config.py
ytalashkocv: Because previous version of code does not worked for Windows

cli/tests/integrations/test_dcos.py
ytalashkocv: same as above

bin/clean.ps1
ytalashkocv: done

cli/tests/integrations/test_analytics.py
ytalashkocv: Because of tests order is very important for Windows. With old order OSError raises (reproducible only on py3.4).

dcos/package.py
ytalashkocv: yes, done

dcos/package.py
ytalashkocv: This function for adding write permission to files on Windows platform. Because exception raised when trying to delete read-only files on Windows.

ytalashkocv:
@mgummelt Please, re review

cli/tests/integrations/test_analytics.py
mgummelt: Why does an OSError raise?  The tests should behave independently, so if they're not, we need to fix that.

cli/tests/integrations/test_config.py
mgummelt: Yes, I assume that's the case for all the changes in this PR.  My question is *why* did it not work on Windows?

cli/tests/integrations/common.py
```diff
@@ -30,7 +30,8 @@ def exec_command(cmd, env=None, stdin=None):
         stderr=subprocess.PIPE,
         env=env)
 
-    stdout, stderr = process.communicate()
+    stdout, stderr = [std_stream.replace(b'\r', b'')
```
mgummelt: I mean put a comment in the code

mgummelt:
I still need the answer to this question:

For the existing tests, our entry point was bin/start_tests.sh What is the entry point for this new test suite?

dcos/package.py
```diff
@@ -943,6 +944,25 @@ def copy_to_cache(self, target_dir):
                 'Unable to fetch packages from [{}]'.format(self.url))
 
 
+def _onerror(func, path, exc_info):
```
mgummelt: This still doesn't follow the pydoc format used in other functions.  Please follow that format.

mgummelt: Also this needs to be named something more specific such as `_rmtree_on_error`

mgummelt: Why is the ~/.git directory created read-only?  If our user clones the repo on windows, shouldn't the user have write access?

dcos/package.py
```diff
@@ -943,6 +944,25 @@ def copy_to_cache(self, target_dir):
...
+    If the error is for another reason it re-raises the error.
+
+    Usage : ``shutil.rmtree(path, onerror=onerror)``
+    """
+    import stat
+    if not os.access(path, os.W_OK):
+        os.chmod(path, stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO)
+        func(path)
+    else:
+        raise
```
mgummelt: Are you sure this works?  I think calling `raise` without an argument only works inside an `except` block

cli/tests/integrations/test_analytics.py
ytalashkocv: We doesn't know exact reason, only think that this connected with some state on Windows. Fails in winapi function call:

mgummelt: This is an error in `test_marathon`, so why did you move tests in `test_analytics`?

ytalashkocv: This error raises in every test after starts at `test_analytics` module: 
https://teamcity.mesosphere.io/viewLog.html?buildId=21743&buildTypeId=ClosedSource_DcosCli_WindowsTests_WindowsTest&tab=buildLog#_focus=23637

ytalashkocv:
About this question https://github.com/mesosphere/dcos-cli/pull/218#issuecomment-111558163:
Added `win_bin\start_tests.ps1` script

cli/tests/integrations/test_config.py
YuraTolochkevich: Windows requires full copy of the ENV, with next its update. If we put new  poor Env like: {
        constants.PATH_ENV: os.environ[constants.PATH_ENV],
        constants.DCOS_CONFIG_ENV: os.path.join("tests", "data", "dcos.toml"),
        cli_constants.DCOS_PRODUCTION_ENV: 'false'
    } it doesn't work for Windows, and we got next error: 
STDOUT:
STDERR: Traceback (most recent call last):
 File "C:\Users\admin\tmp\dcos-cli\cli\.tox\py27-integration\Scripts\dcos-script.py", line 5, in <module>
   from pkg_resources import load_entry_point
 File "C:\Users\admin\tmp\dcos-cli\cli\.tox\py27-integration\lib\site-packages\pkg_resources\__init__.py", line 37, in <module>
   import email.parser
 File "c:\python27\Lib\email\parser.py", line 12, in <module>
   from email.feedparser import FeedParser
 File "c:\python27\Lib\email\feedparser.py", line 27, in <module>
   from email import message
 File "c:\python27\Lib\email\message.py", line 16, in <module>
   import email.charset
 File "c:\python27\Lib\email\charset.py", line 13, in <module>
   import email.base64mime
 File "c:\python27\Lib\email\base64mime.py", line 40, in <module>
   from email.utils import fix_eols
 File "c:\python27\Lib\email\utils.py", line 27, in <module>
   import random
 File "c:\python27\Lib\random.py", line 885, in <module>
   _inst = Random()
 File "c:\python27\Lib\random.py", line 97, in __init__
   self.seed(x)
 File "c:\python27\Lib\random.py", line 113, in seed
   a = long(_hexlify(_urandom(2500)), 16)
WindowsError: [Error -2146893795] Provider DLL failed to initialize correctly

dcos/package.py
ytalashkocv: `raise` works fine. But I changed `raise` to `raise exc_info[0], exc_info[1], exc_info[2]`. This looks more obvious.

ytalashkocv:
@mgummelt All comments have been addressed. Please review.

dcos/package.py
```diff
@@ -943,6 +946,25 @@ def copy_to_cache(self, target_dir):
                 'Unable to fetch packages from [{}]'.format(self.url))
 
 
+def _rmtree_on_error(func, path, exc_info):
```
mgummelt: This *still* doesn't follow our pydoc format.   Please look at the docstrings for the other functions.

dcos/package.py
mgummelt: I just tried it in python2.7, and got this exception:

`TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType`

Let's be explicit about what we're raising

cli/tests/integrations/test_analytics.py
mgummelt: Why do you think moving the tests fixes it?  Why did you try that?

I'd really like to know why this is happening before we commit this.  If we don't understand this error, then in the likely case that it comes up again, we're not going to be able to fix it.

dcos/package.py
ytalashkocv: @mgummelt Improve `_rmtree_on_error` pydoc

cli/tests/integrations/test_analytics.py
ytalashkocv: When this failure happened (tests failed on py3.4 but works fine for py2.7), we noticed that tests have different order for py2.7 and py3.4(py3.4 runs tests in same order as they are ordered in python module). So we tried to change tests order (as they runned by py2.7). And this helped. Also we try to use py.test annotations like `@pytest.mark.first` but they doesn't work. As I mentioned here https://github.com/mesosphere/dcos-cli/pull/218#discussion_r32337010, we doesn't know exact reason of this failure, but if needed we can research this problem (we also tried, but without any success).

dcos/package.py
ytalashkocv: You got such `TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType` message, because, in current call stack, exception raising not happened (`raise` without arguments re-raises last raised exception). In our case `_rmtree_on_error` is an onerror callback, that called only after exception raised.

ytalashkocv:
@mgummelt All comments addressed. Please review.

cli/tests/integrations/test_analytics.py
mgummelt: Yes, please research the cause.  This is likely to come up again as we evolve the test suite.

cli/tests/integrations/test_config.py
mgummelt: Hmm.   This might be OK, but can you first figure out which missing env var is causing this failure?

dcos/package.py
mgummelt: Ah, got it.  Let's go ahead and raise exc_info anyway, so this function doesn't have to depend on global state (like the exception stack)

ytalashkocv: `raise` without arguments - is a common way for re-raising exceptions. One more reason, is changes in py3 for exception raising. If raise exception manually using exc_info, then such `During handling of the above exception, another exception occurred` message will be shown. Py3 identifying this exception like new, and not like re-raising old one.

mgummelt: OK

cli/tests/integrations/test_config.py
ytalashkocv: It's `SYSTEMROOT` env variable. Value of this variable is `C:\Windows`.

cli/tests/integrations/test_analytics.py
ytalashkocv: I do not found any additional information about this issue.

ytalashkocv:
@mgummelt Comments addressed. Please review.

mgummelt:
OK, one last thing: Who setup the teamcity configuration for this?

I'm very concerned that the failing windows test doesn't actually have anything to do with test ordering, and might just be a non-deterministic bug.  Regardless, it's a failure that we don't understand, and it's likely to come up again.

ytalashkocv:
@mgummelt I have setup teamcity configuration and Yurii Tolochkevych create ami image (Windows server 2012).

mgummelt:
How is DCOS installed on that machine?  From the TC config, it looks to be a long-running DCOS instance that you've preinstalled.  I think this might cause problems with frameworks re-registering in subsequent test runs because ZooKeeper state isn't cleared.

We run DCOS in a vagrant image from the dcos-image project for our other tests.  Have you looked into that?

mgummelt:
Regardless, this LGTM.  You can merge.  I'll leave it up to you to manage that TC config.  Let me know if you need any help.

YuraTolochkevich:
@mgummelt, Right, we use preinstalled  DCOS cluster for tests. Regarding DCOS vagrant  image,  we tried to deploy DCOS with vagrant, but our AWS  Windows machine is Virtual machine itself and it is a case of the nested visualization, and only allows deploying  32-bit  VMs with Virtual Box, VmWare or other provider. If there is an possibility to provide Windows machine installed on bare metal, we can try run tests with vagrant

mgummelt:
@spahl ^^

PR #224, opened 2015-06-16
fix group schema.  fixes #219

### LINKED ISSUE #219: 'dcos marathon groups add' error when dependencies defined
I'm testing marathon's application groups via dcos and created a sample json. When I'm posting it via curl to Marathons REST api, it is successfully started. However via dcos cli I get errors.

If I add a toplevel empty
```"apps": [ ]```
property into the json, then I get:
```RefResolutionError: Unresolvable JSON pointer: u'definitions/pathType'```

Without the ```apps``` property just a lengthy stacktrace.

This is the command I'm using:
```dcos marathon group add multi.json```

The json:
```
{
  "id": "/multi-test7",
  "groups": [ {
    "id": "/multi-test7/persistence",
    "apps": [ {
      "id": "/multi-test7/persistence/mongo",
      "cmd": "while [ true ] ; do echo 'Fake MongoDB server' ; sleep 5 ; done",
      "cpus": 0.1,
      "mem": 64.0,
      "instances": 1
    }]
  },{
    "id": "/multi-test7/persistence2",
    "apps": [ {
      "id": "/multi-test7/persistence2/mongo",
      "dependencies": ["/multi-test7/persistence/mongo"],
      "cmd": "while [ true ] ; do echo 'Fake MongoDB server' ; sleep 5 ; done",
      "cpus": 0.1,
      "mem": 64.0,
      "instances": 1
    }]
  } ]
}
```
After some testing it turned out that if I remove the ```dependencies``` property from the json, it will be posted successfully to Marathon via dcos cli.

### PR DESCRIPTION
**Pull request by mgummelt**
fixes #219

### DIFF
--- a/cli/tests/data/marathon/apps/zero_instance_sleep.json
+++ b/cli/tests/data/marathon/apps/zero_instance_sleep.json
@@ -4,6 +4,7 @@
   "cpus": 0.1,
   "mem": 16,
   "instances": 0,
+  "dependencies": [],
   "labels": {
     "PACKAGE_ID": "zero-instance-app",
     "PACKAGE_VERSION": "1.2.3"
--- a/cli/dcoscli/data/marathon-group-schema.json
+++ b/cli/dcoscli/data/marathon-group-schema.json
@@ -143,7 +143,7 @@
                 },
                 "dependencies": {
                     "items": {
-                        "$ref": "#/definitions/pathType"
+                        "$ref": "#/definitions/app/definitions/pathType"
                     },
                     "type": "array"
                 },

### CODE REVIEW

**Comment by mgummelt** [comment]
@ConnorDoyle plz review

**Comment by ConnorDoyle** [comment]
Thanks @mgummelt.  Please add a test case to exercise this part of the schema (group with dependencies).

**Comment by mgummelt** [comment]
@ConnorDoyle done

File: cli/tests/data/marathon/apps/zero_instance_sleep.json [hunk 1]
```diff
@@ -4,6 +4,7 @@
   "cpus": 0.1,
   "mem": 16,
   "instances": 0,
+  "dependencies": [],
```
ConnorDoyle: In order to test the item schema, we should add at least one dependency.

mgummelt:
I'll go ahead and add one, but I'm not really interested in testing every aspect of this schema.  that would be a lot of tests.

cli/tests/data/marathon/apps/zero_instance_sleep.json
mgummelt: I'll go ahead and add one, but I'm not really interested in testing every aspect of this schema. that would be a lot of tests.

ConnorDoyle:
Sure, it's just that without it the new dependency field doesn't exercise the change you made to the schema.

ConnorDoyle:
I think adding them iteratively like this as anti-regression measure makes sense.

ConnorDoyle:
Please :ship: once tests are green with the dependency change.

mgummelt:
Actually it looks as if our marathon schema logic is a bit more broken than I thought.  I assigned @tamarrow a JIRA

mgummelt:
Closing for now

PR #181, opened 2015-05-22
dcos tail -- DCOS-1133

mgummelt:
This is missing tests.  I figured I'd go ahead and get the conversation started while I write the tests, then we can review tests separately later (part of the same PR, of course).  

Tomorrow (5/22), I'll mark this up with comments describing some of my design decisions.

PR #229, opened 2015-06-26
use contextmanager for marathon app tests

tamarrow:
Tests should clean up state even if they fail.


mgummelt:
This is already done in #181

PR #230, opened 2015-06-26
group rollbar messages around just the exception title

mgummelt:


ssk2:
How would this be different for an exception like [this](https://rollbar.com/thomas-mesosphere/dcos/items/1187/)?

mgummelt:
Entire message would be just "IOError: [Errno 2] No such file or directory: 'vangogh'" and the stacktrace is in the "stderr" property of the occurrence

ssk2:
Hopefully still an improvement! Thanks!

PR #233, opened 2015-06-27
dcos-1886 don't crash if invalid json commands

tamarrow:
Instead of leaving a stack trace, show more helpful message. 
```
Tamars-MacBook-Pro:dcos-cli tamar$ dcos marathon app update {"instances": 2}
Error: No JSON object could be decoded
```

In this command, app-id wasn't specified, and we parsed it as `"{"` with properties `'"instances": 2}'`, which is what caused the stacktrace.

### DIFF
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -361,6 +361,16 @@ def test_update_bad_type():
         assert stdout == b''
 
 
+def test_update_invalid_request():
+    returncode, stdout, stderr = exec_command(
+        ['dcos', 'marathon', 'app', 'update', '{', 'instances'])
+    assert returncode == 1
+    assert stdout == b''
+    stderr = stderr.decode()
+    assert stderr.startswith('Error while fetching')
+    assert stderr.endswith('HTTP 400: Bad Request\n')
+
+
 def test_update_app():
     with _zero_instance_app():
         returncode, stdout, stderr = exec_command(
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -54,6 +54,11 @@ def _to_exception(response):
     if isinstance(response, Exception):
         return DCOSException(_default_marathon_error(str(response)))
 
+    if response.status_code == 400:
+        return DCOSException(
+            'Error while fetching [{0}]: HTTP {1}: {2}'.format(
+                response.request.url, response.status_code, response.reason))
+
     message = response.json().get('message')
     if message is None:
         errs = response.json().get('errors')


dcos/marathon.py
```diff
@@ -54,7 +54,11 @@ def _to_error(response):
     if isinstance(response, Error):
         return DefaultError(_default_marathon_error(response.error()))
 
-    message = response.json().get('message')
+    try:
```
mgummelt: can you add a regression test?

dcos/marathon.py
```diff
@@ -54,7 +54,11 @@ def _to_error(response):
...
+        message = response.json().get('message')
+    except ValueError as e:
```
mgummelt: When does marathon return a json object with message field?  What kind of response?

tamarrow: http://docs.python-requests.org/en/latest/user/quickstart/#json-response-content

tamarrow: returned from http.request

mgummelt: I know what json() does

tamarrow: When marathon returns an error with what the user is trying to do, update app that doesn't exist, etc

dcos/marathon.py
tamarrow: done

dcos/marathon.py
mgummelt: That isn't true for the user error that you're addressing here.  There is no 'message' field.  That's the source of the bug.  Can you point me to the marathon docs for this 'message' field?

tamarrow: There aren't really docs about the message spec, but its mentioned here: https://mesosphere.github.io/marathon/docs/rest-api.html

cli/tests/integrations/test_marathon.py
```diff
@@ -388,6 +388,14 @@ def test_update_bad_type():
     _remove_app('zero-instance-app')
 
 
+def test_update_invalid_json():
+    returncode, stdout, stderr = exec_command(
+        ['dcos', 'marathon', 'app', 'update', '{', 'instances'])
+    assert returncode == 1
+    assert stdout == b''
+    assert stderr.decode() == 'Error: No JSON object could be decoded\n'
```
mgummelt: This isn't a great error message

dcos/marathon.py
mgummelt: Can you tell the marathon team that this is missing in the docs?  Or file a Github issue in the marathon project?

We should be checking the response status here, and forming the appropriate error message.  Catching this exception is a hack.  Also, you're surfacing the JSON parsing error, which is not what the user cares about.

tamarrow: filed a github issue. 

Why should we check the response code? In this case it would be 400, and we could just say `Bad Request`, but isn't it more helpful for the user that they know what the error is (ie bad json)?

mgummelt: That's not actually what the error is.  The JSON error you're surfacing is the exception in response.json() where requests finds that it can't parse the response as JSON.  It's only a coincidence that this has something to do with the JSON-formatting of that command line argument.

tamarrow:
@mgummelt bump

dcos/marathon.py
mgummelt: This concern still applies

mgummelt:
See my comment above.  It still applies.

dcos/marathon.py
tamarrow: The actual bug is in marathon. But until it's fixed, we should still catch the error.

mgummelt: I don't understand.  We'll have to talk in person tomorrow.

tamarrow:
@mgummelt:
request: /v2/apps/%7Binstances%3A
error_code: 400
Bad Request

dcos/marathon.py
mgummelt: OK, I understand now.  We should first check the return code, and only try to parse the 'message' if it's a 404.  This 400 should result in the Bad Request message.  See `http._default_to_exception` for an example

tamarrow: updated

dcos/marathon.py
```diff
@@ -54,6 +54,11 @@ def _to_exception(response):
     if isinstance(response, Exception):
         return DCOSException(_default_marathon_error(str(response)))
 
+    if response.status_code == 400:
```
mgummelt: We only know the `message` field to be in a 404 response, right?  So this should be `!= 404`

tamarrow: no, in `test_show_bad_app_version` the resulting error is from status_code `422`

tamarrow:
bump

mgummelt:
LGTM

PR #234, opened 2015-06-29
DCOS-1762 hopefully this fixes CLI commands hanging

mgummelt:
set read timeout and connect timeout

### DIFF
--- a/cli/dcoscli/analytics.py
+++ b/cli/dcoscli/analytics.py
@@ -103,10 +103,12 @@ def _segment_request(path, data):
         SEGMENT_IO_WRITE_KEY_DEV
 
     try:
+        # Set both the connect timeout and the request timeout to 1s,
+        # to prevent rollbar from hanging the CLI commands
         http.post('{}/{}'.format(SEGMENT_URL, path),
                   json=data,
                   auth=HTTPBasicAuth(key, ''),
-                  timeout=1)
+                  timeout=(1, 1))
     except Exception as e:
         logger.exception(e)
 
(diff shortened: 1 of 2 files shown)


mgummelt:
@jsancio plz review

cli/dcoscli/analytics.py
```diff
@@ -107,7 +107,7 @@ def _segment_request(path, data):
         requests.post('{}/{}'.format(SEGMENT_URL, path),
                       json=data,
                       auth=HTTPBasicAuth(key, ''),
-                      timeout=1)
```
jsancio: Can you write a small comment as to why we are doing this? Also, can we change this to use our `dcos.http` package? I remember we didn't get decent debug output (with `--log-level=...`) because we weren't using `dcos.http`.

Also, should we change `dcos.http` to do the same thing?

jsancio:
@mgummelt Let me know if you are waiting on me for this PR.

cli/dcoscli/analytics.py
jsancio: I don't think we need to change it in this PR. I think #243 is already doing this.

mgummelt: ok, I'll wait for #243

mgummelt:
blocked on #243

jsancio:
@mgummelt FYI, we merged #243. Looks like it needs a rebase.

mgummelt:
@jsancio: updated

jsancio:
Looks good. Should we change the default and all of the call sites to set both read and connect timeout `(x, x)` (connect and read timeouts)? I think these are all of the places we should change:

1. https://github.com/mesosphere/dcos-cli/blob/master/dcos/http.py#L49
1. https://github.com/mesosphere/dcos-cli/blob/master/dcos/mesos.py#L12

PR #236, opened 2015-06-29
DCOS-1455 better app status info in dcos marathon app list

mgummelt:


cli/dcoscli/tables.py
```diff
@@ -48,13 +61,40 @@ def get_container(app):
...
+
     fields = OrderedDict([
         ("ID", lambda a: a["id"]),
         ("MEM", lambda a: a["mem"]),
         ("CPUS", lambda a: a["cpus"]),
-        ("DEPLOYMENTS", lambda a: len(a["deployments"])),
         ("TASKS", lambda a: "{}/{}".format(a["tasksRunning"],
                                            a["instances"])),
+        ("HEALTH", get_health),
+        ("DEPLOYMENT", get_deployment),
```
ssk2: I'm not sure I like "DEPLOYMENT" as the title here. How about "ACTION"?

mgummelt: "deployment" is a marathon term.  we should stay consistent.

ssk2: Got it!

PR #237, opened 2015-06-30
dcos node

mgummelt:
I was about to implement `dcos node log`, and figured we probably need a top level `dcos node` first.  It only lists slaves, because DCOS does not provide an endpoint to get the masters AFAIK.


mgummelt:
@jsancio tests pass.  plz review

cli/dcoscli/node/main.py
```diff
@@ -0,0 +1,92 @@
...
+
+def main():
+    try:
+        return _main()
+    except DCOSException as e:
+        emitter.publish(e)
+        return 1
+
+
+def _main():
```
jsancio: Out of scope for this PR but should we refactor this dance (`main` and `_main`) and move it to `dcos`? I haven't thought it much but it feels useful for other sub-commands.

cli/tests/unit/data/node.txt
```diff
@@ -0,0 +1,2 @@
+ HOSTNAME       IP                          ID                   
+ dcos-01   172.17.8.101  20150630-004309-1695027628-5050-1649-S0
```
jsancio: Hmm. Would this fail if we run the tests on a developer laptop?

jsancio: I see. Ignore my comment. This is a unit test.

cli/dcoscli/tables.py
```diff
@@ -276,3 +276,21 @@ def package_search_table(search_results):
     tb.align['DESCRIPTION'] = 'l'
 
     return tb
+
+
+def slave_table(slaves):
+    """Returns a PrettyTable representation of the provided DCOS slaves
+
+    :param slaves: slaves to render.  dicts from /mesos/state-summary
+    :type aves: [dict]
```
jsancio: s/aves/slaves/

jsancio:
Looks good to me! Just some minor comments.

cli/dcoscli/node/main.py
mgummelt: we could, yea

PR #235, opened 2015-06-29
add new marathon version in universe to test

tamarrow:
universe #164 added a new marathon version to universe which broke `test_describe` since `dcos package describe marathon` changed. It's not great that universe changes can break our test :disappointed:

PR #238, opened 2015-06-30
Include all marathon versions in the package

jsancio:


cli/tests/integrations/test_package.py
```diff
@@ -207,7 +207,8 @@ def test_describe():
     "framework"
   ],
   "versions": [
-    "0.8.1"
+    "0.8.1",
+    "0.9.0-RC3"
```
jsancio: @BenWhitehead @mgummelt Do we know why the Marathon package has multiple versions? Is this the intended usage?

jsancio: I see. Maybe the issues is in how we generate the version in the CLI. Is the plan for the user to maybe do something like `dcos package install --version=0.8.1 marathon`? or `dcos package install --version=0 marathon`? @ConnorDoyle?

ConnorDoyle: It's intended to work this way.  The index JSON in the universe provides a map for each package from unique software version (like 0.8.1) to the largest ordinal package version that provides that software version (like 0).  There's already code to implement the first example flag you mentioned but it needs to be exposed.  It would be useful to be able to specify either the software version or the package version.  For this reason I recommend somehow making it explicit (e.g. --software-version / --package-version) or some better naming scheme.  There was a JIRA for adding this feature at one point but I can't seem to find it right now.

mgummelt:
@tamarrow already has PR #235 that should fix this

jsancio:
I am closing this PR. The build was already fixed.

PR #242, opened 2015-06-30
dcos-595 give notice if no packages

tamarrow:


cli/dcoscli/package/main.py
```diff
@@ -390,7 +391,12 @@ def _list(json_, endpoints, app_id, package_name):
 
             results.append(pkg_info)
 
-    emitting.publish_table(emitter, results, tables.package_table, json_)
+    if results:
+        emitting.publish_table(emitter, results, tables.package_table, json_)
+    else:
+        msg = ("There are currently no installed packages. "
+               "Use `dcos package install` to install a package.")
```
ssk2: Let's add a please -> "Please use ..."

ssk2:
This looks fine but let's double check what's happening with list-installed (see the comment on the relevant JIRA issue) before merging.

cli/dcoscli/package/main.py

tamarrow:
@jsancio did we settle on `list` instated of `list-installed`?

cli/dcoscli/package/main.py
```diff
@@ -6,7 +6,8 @@
     dcos package describe [--app --options=<file> --cli] <package_name>
     dcos package install [--cli | [--app --app-id=<app_id>]]
                          [--options=<file> --yes] <package_name>
-    dcos package list [--json --endpoints --app-id=<app-id> <package_name>]
+    dcos package list-installed [--json --endpoints --app-id=<app-id>
```
jsancio: We should keep the command as `dcos pacakge list` sorry for the confusion.

jsancio:
> @jsancio did we settle on list instated of list-installed?

We settled on `dcos package list`.

cli/dcoscli/package/main.py
tamarrow: np, changed it back to `list`

cli/dcoscli/package/main.py
```diff
@@ -401,7 +401,12 @@ def _list(json_, endpoints, app_id, package_name):
 
             results.append(pkg_info)
 
-    emitting.publish_table(emitter, results, tables.package_table, json_)
+    if results:
+        emitting.publish_table(emitter, results, tables.package_table, json_)
+    else:
+        msg = ("There are currently no installed packages. "
+               "Please use `dcos package install` to install a package.")
+        emitter.publish(msg)
```
jsancio: This is going to break scripts. E.g. `dcos package list --json | jq ...`. jq will break because it can't parse the string `There are currently...`.

tamarrow: updated

cli/dcoscli/package/main.py
```diff
@@ -401,7 +401,14 @@ def _list(json_, endpoints, app_id, package_name):
...
             results.append(pkg_info)
 
-    emitting.publish_table(emitter, results, tables.package_table, json_)
+    if results:
+        emitting.publish_table(emitter, results, tables.package_table, json_)
+    else:
+        msg = ("There are currently no installed packages. "
+               "Please use `dcos package install` to install a package.")
+        if json_:
+            msg = json.dumps(msg)
```
jsancio: I don't think this is going to work. In the "success" case it will return an JSON array. In the "failure" case it will return a JSON string. Let's try to keep this stuff consistent.

cli/dcoscli/package/main.py
```diff
@@ -401,7 +401,14 @@ def _list(json_, endpoints, app_id, package_name):
 
             results.append(pkg_info)
 
-    emitting.publish_table(emitter, results, tables.package_table, json_)
+    if results:
+        emitting.publish_table(emitter, results, tables.package_table, json_)
+    else:
+        msg = ("There are currently no installed packages. "
+               "Please use `dcos package install` to install a package.")
+        if json_:
```
mgummelt: use a DCOSException for consistency.  Also, don't json format it

tamarrow: see comment above before it was json-ed: 'This is going to break scripts. E.g. dcos package list --json | jq .... jq will break because it can't parse the string There are currently....'

mgummelt: It would break scripts anyway as soon as they try to parse the contents.

Use a DCOSException so it goes to stderr.

mgummelt: oh, wait, no, sorry.  --json should result in `[]` (which `publish_table` should handle I think)

tamarrow:
updated

mgummelt:
LGTM

PR #243, opened 2015-07-01
[ADD] DCOS-2198 add proxy support

YuraTolochkevich:


dcos/package.py
```diff
@@ -845,7 +846,15 @@ def copy_to_cache(self, target_dir):
...
-                urllib.request.urlretrieve(self.url, tmp_file)
+
+                req = requests.get(self.url, stream=True,
+                                   proxies=util.get_proxy_dict())
+                if req.status_code == 200:
+                    with open(tmp_file, 'wb') as f:
+                        for chunk in req.iter_content(1024):
+                            f.write(chunk)
+                else:
+                    raise Exception
```
jsancio: Let's raise a DCOSException with a nice message.

YuraTolochkevich: DCOSException is re-raised below 
        except Exception:
            raise DCOSException(
                'Unable to fetch packages from [{}]'.format(self.url))

YuraTolochkevich: @jsancio  Thanks Jose! Are there any more comments to this PR?

dcos/package.py
```diff
@@ -845,7 +846,15 @@ def copy_to_cache(self, target_dir):
                 tmp_file = os.path.join(tmp_dir, 'packages.zip')
 
                 # Download the zip file.
-                urllib.request.urlretrieve(self.url, tmp_file)
+                requests.packages.urllib3.disable_warnings()
+                req = requests.get(self.url, stream=True,
+                                   proxies=util.get_proxy_dict())
```
jsancio: Can we use `dcos.http` instead of using `requests` directly?

dcos/package.py
```diff
@@ -845,7 +846,15 @@ def copy_to_cache(self, target_dir):
                 tmp_file = os.path.join(tmp_dir, 'packages.zip')
 
                 # Download the zip file.
-                urllib.request.urlretrieve(self.url, tmp_file)
+                requests.packages.urllib3.disable_warnings()
```
jsancio: We have a helper method for this: https://github.com/mesosphere/dcos-cli/blob/master/dcos/http.py#L195-L197.

jsancio: We normally don't disable this warning in the `dcos` package instead we have been doing it at the entry point for each program in the `dcoscli` package. See: https://github.com/mesosphere/dcos-cli/blob/master/cli/dcoscli/config/main.py#L52

cli/dcoscli/analytics.py
```diff
@@ -107,7 +107,8 @@ def _segment_request(path, data):
         requests.post('{}/{}'.format(SEGMENT_URL, path),
                       json=data,
                       auth=HTTPBasicAuth(key, ''),
-                      timeout=1)
+                      timeout=1,
+                      proxies=util.get_proxy_dict())
```
jsancio: To avoid reading the config file 3 times instead of just once. Let's either pass the `conf` object into `_segment_request` or pass the proxy information into it.

dcos/util.py
```diff
@@ -548,3 +548,41 @@ def io_exception(path, errno):
...
+
+
+def _get_http_proxy_url():
+    """ Returns http proxy server url from config
+
+    :returns: Http Proxy server URL
+    :rtype: str
+    """
+    config = get_config()
+    return config.get('core.http_proxy')
```
jsancio: I am not sure that we want to use this pattern. Normally we read and parse the config file once and pass the option or parts of the object down to the functions that need it. I think we should keep that pattern.

jsancio:
Thanks for the PR! I added some comments.

YuraTolochkevich:
@jsancio Thanks  for review. please see updates

dcos/http.py
```diff
@@ -71,7 +71,8 @@ def request(method,
             request.headers)
 
         with requests.Session() as session:
-            response = session.send(request.prepare(), timeout=timeout)
+            response = session.send(request.prepare(), timeout=timeout,
+                                    proxies=util.get_proxy_dict())
```
jsancio: This means that we need to read the config file and parse it every time we do an HTTP request. Can we get around this? Looking at the [documentation](http://docs.python-requests.org/en/latest/user/advanced/#proxies) it looks like `requests` support reading proxy config from environment variables. Should we just overwrite the environment variable if we find those entries in the `dcos.toml` config?

jsancio:
Looks good! Added one last comment. Let's squash the commits. We try to keep PRs to one commit.

YuraTolochkevich:
@jsancio  yes, we can use environment variable with proxy entries from dcos.toml, but when  we should track that proxy values from the dcos.tom and proxy values from environment variables are identical? every time when user execute command `dcos config set http_proxy`? but if htttp_proxy is already exist in the dcos.toml and not  exist in environment variables?

jsancio:
Well, I just want HTTP/HTTPS proxy to work. If the standard way of doing this is through the environment variable then we should just support that. And not support setting it through the `dcos.toml`. @tamarrow, can you work @YuraTolochkevich on this issue?

YuraTolochkevich:
@jsancio Thanks. I have replaced reading proxy values from `dcos.toml` to reading from environment variables

dcos/util.py
```diff
@@ -547,3 +547,19 @@ def io_exception(path, errno):
...
+
+    :returns: Dict with proxy parameters
+    :rtype: dict
+    """
+
+    proxies = dict()
+
+    for name, value in os.environ.items():
+        name = name.lower()
+        if value and name[-6:] == '_proxy':
```
jsancio: Let's be more explicit on the comparison. Easier on the reader later. Can we match for only http_proxy and https_proxy? Thanks!

jsancio:
Looks good! One last comment. Thanks.

YuraTolochkevich:
@jsancio Ready, thanks. looks like, vagrant integration tests are broken now

jsancio:
> @jsancio Ready, thanks. looks like, vagrant integration tests are broken now

Yes. Looks like we changed how we build DCOS Images.

jsancio:
Looks good! :ship:ing it!

dcos/http.py
```diff
@@ -83,7 +83,8 @@ def request(method,
             request.headers)
 
         with requests.Session() as session:
-            response = session.send(request.prepare(), timeout=timeout)
+            response = session.send(request.prepare(), timeout=timeout,
+                                    proxies=util.get_proxy_dict_from_env())
```
mgummelt: I don't think this is ncessary.  `requests` should understand the ENV vars w/o explicitly setting `proxies`

jsancio: @mgummelt I suspect the issue they ran into was case sensitivity. On Windows I think environment variables are case insensitive and on Linux they are case sensitive.

mgummelt: Also that we were using urllib, not requests

jsancio: :+1:

PR #245, opened 2015-07-01
[DCOS-1133] dcos node log

mgummelt:
I don't know why flake8 started complaining about the docopt import, but it did, so I moved those.

I've added `dcos node log`, and moved logging code from task.py into a separate log.py, since it is now used for both `dcos node log` and `dcos task log`

To support accessing the mesos logs, I've updated the `MesosFile` class to support arbitrary mesos files, not just those associated with task.


mgummelt:
@jsancio plz review

cli/dcoscli/log.py
```diff
@@ -0,0 +1,197 @@
...
+    :type follow: bool
+    :param lines: number of lines to print
+    :type lines: int
+    :rtype: None
+    """
+
+    fn = functools.partial(_read_last_lines, lines)
+    curr_header, mesos_files = _stream_files(None, fn, mesos_files)
+    if not mesos_files:
+        raise NO_FILE_EXCEPTION
```
jsancio: Is this correct? Should you raise a global/static object? What happens if you `raise NO_FILE_EXCEPTION` twice?

cli/dcoscli/node/main.py
```diff
@@ -3,19 +3,24 @@
 Usage:
     dcos node --info
     dcos node [--json]
+    dcos node log [--follow --lines=N --master --slave=<slave-id>]
```
jsancio: Should this be `[--follow --lines=N] --master | --slave=<slave-id>`? Not sure how I feel about "required optional" flags. Do we have an example of a CLI that does this? cc @ConnorDoyle @ssk2 @tamarrow .

jsancio: One way to get around this is to make `--master` implicit. The main problem I see with this is how do we document it given how we currently structure the `--help` output.

jsancio:
Do you mind going through this PR in person? Let's schedule 25 min to go through it today.

mgummelt:
OK stop by anytime before 3:30

cli/dcoscli/node/main.py
mgummelt: --master and --slave aren't actually mutually exclusive.  You can log both at the same time.  It's just you have to choose at least one.  I don't think there's any way to express this in docopt.

cli/dcoscli/log.py
mgummelt: I'll change it to be a function

dcos/mesos.py
```diff
@@ -530,9 +530,10 @@ def __getitem__(self, name):
 
 
 class MesosFile(object):
-    """File-like object that is backed by a remote slave file.  Uses the
-    files/read.json endpoint.  This endpoint isn't well documented
-    anywhere, so here is the spec derived from the mesos source code:
+    """File-like object that is backed by a remote slave or master file.
+    Uses the files/read.json endpoint.  This endpoint isn't well
+    documented anywhere, so here is the spec derived from the mesos
+    source code:
```
mgummelt: better document constructor dependencies

dcos/mesos.py
```diff
@@ -692,8 +712,11 @@ def _fetch(self, params):
         :rtype: dict
         """
 
-        read_url = self._mesos_client.slave_url(self._task.slave()['id'],
-                                                'files/read.json')
+        if self._slave:
+            read_url = self._mesos_client.slave_url(self._slave['id'],
+                                                    'files/read.json')
+        else:
+            read_url = self._mesos_client.master_url('files/read.json')
```
mgummelt: look into moving to mesos_client

mgummelt:
@jsancio when do you think you'll get to this?  It's blocking two other PRs

jsancio:
@mgummelt We should always ping the review when we push PR changes. Unfortunately, the reviewer doesn't get notify of those updates.

jsancio:
Looking at it now.

cli/dcoscli/log.py
```diff
@@ -0,0 +1,199 @@
...
+    :type follow: bool
+    :param lines: number of lines to print
+    :type lines: int
+    :rtype: None
+    """
+
+    fn = functools.partial(_read_last_lines, lines)
+    curr_header, mesos_files = _stream_files(None, fn, mesos_files)
+    if not mesos_files:
+        raise _no_file_exception()
```
jsancio: :+1:

cli/dcoscli/log.py
```diff
@@ -0,0 +1,199 @@
...
+
+    while follow:
+        # This flush is needed only for testing, since stdout is fully
+        # buffered (as opposed to line-buffered) when redirected to a
+        # pipe.  So if we don't flush, our --follow tests, which use a
+        # pipe, never see the data
+        sys.stdout.flush()
```
jsancio: :+1:. I think we have a similar problem when we prompt the user if they want to install the package.

cli/dcoscli/node/main.py
```diff
@@ -90,3 +100,50 @@ def _list(json_):
...
+    :type master: bool
+    :param slave: the slave ID to print
+    :type slave: str | None
+    :returns: process return code
+    :rtype: int
+    """
+
+    if not (master or slave):
+        raise DCOSException('You must choose one of --master or --slave.')
+    lines = int(lines)
```
jsancio: This throws a ValueError exception. We have a help method that shows a nicer error: https://github.com/mesosphere/dcos-cli/blob/master/dcos/util.py#L366-L383

On a similar note, can we add tests for all the failure cases like this one?

cli/tests/integrations/test_node.py
```diff
@@ -13,12 +14,17 @@ def test_help():
...
 Options:
-    -h, --help    Show this screen
-    --info        Show a short description of this subcommand
-    --json        Print json-formatted nodes
-    --version     Show version
+    -h, --help            Show this screen
+    --info                Show a short description of this subcommand
+    --json                Print json-formatted nodes
+    --follow              Output data as the file grows
+    --lines=N             Output the last N lines [default: 10]
```
jsancio: Can we add tests for this option? Valid numbers and invalid lines arguments.

dcos/mesos.py
```diff
@@ -100,6 +100,52 @@ def get_state_summary(self):
         url = self.master_url('master/state-summary')
         return http.get(url).json()
 
+    def slave_file_read(self, slave_id, params):
```
jsancio: I know this is not part of this PR so okay. But the argument `param` is too obscure to me as user of this library/module. It requires that we document what keys that are allowed in params. Looking at the rest of the code I see that they are `path`, `length`, `offset`. Why not elevate them to the function? E.g. `def slave_file_read(self, slave_id, path, offset, length)`

dcos/mesos.py
```diff
@@ -100,6 +100,52 @@ def get_state_summary(self):
...
+
+        request format:
+        {
+            path: absolute path to read
+            offset: start byte location, or -1.  -1 means read no data, and
+                    is used to fetch the size of the file in the response's
+                    'offset' parameter.
+            length: number of bytes to read, or -1.  -1 means read the whole
+                    file.
+        }
```
jsancio: I see. They are documented here. If those params are part of the function signature we could have documented it using pydoc's `:param path:` syntax.

dcos/mesos.py
```diff
@@ -530,40 +576,41 @@ def __getitem__(self, name):
...
+    :type slave: Slave | None
+    :param mesos_client: client to use for network requests
+    :type mesos_client: MesosClient | None
+
     """
 
-    def __init__(self, task, path, mesos_client):
+    def __init__(self, path, task=None, slave=None, mesos_client=None):
+        if task and slave:
+            raise ValueError(
```
jsancio: Why ValueError? Why not DCOSException?

jsancio:
Look good. Thanks! Minor comments. Ping me once this PR is passing again.

dcos/mesos.py
mgummelt: Because it indicates a coding error.  i.e. an error that should never happen

mgummelt:
all comments addressed @jsancio

PR #253, opened 2015-07-02
[DCOS-1138] dcos node ssh

mgummelt:


mgummelt:
@tamarrow: plz review

cli/dcoscli/node/main.py
```diff
@@ -4,18 +4,29 @@
...
+    --json                  Print json-formatted nodes
+    --follow                Output data as the file grows
+    --lines=N               Output the last N lines [default: 10]
+    --master                Access the leading master
+    --slave=<slave-id>      Access the slave with the provided ID
+    --version               Show version
+    --option SSHOPT=VAL     SSH option (see `man ssh_config`)
+    --config-file=<path>    Path to ssh config file
+    --user=<user>           SSH user [default: core]
+    --version               Show version
```
tamarrow: `version is listed twice`

cli/dcoscli/node/main.py
```diff
@@ -4,18 +4,29 @@
     dcos node --info
     dcos node [--json]
     dcos node log [--follow --lines=N --master --slave=<slave-id>]
+    dcos node ssh [--option SSHOPT=VAL ...]
```
tamarrow: why are there `...`?

cli/dcoscli/node/main.py
```diff
@@ -148,3 +165,61 @@ def _mesos_files(master, slave_id):
...
         files.append(mesos.MesosFile('/slave/log', slave=slave))
     return files
+
+
+def _ssh(master, slave, option, config_file, user):
+    """SSH into a DCOS node.  Since only the masters are definitely
+    publicly available, we first ssh into an arbitrary master, then
+    hop to the desired node.
+
+    :param master: True if we ssh into the master
```
tamarrow: This description is confusing since the above comment says `we always ssh into an arbitrary master`

cli/dcoscli/node/main.py
mgummelt: whoops

cli/dcoscli/node/main.py
mgummelt: you can specify multiple options

cli/dcoscli/node/main.py
mgummelt: fixed

cli/dcoscli/node/main.py
```diff
@@ -148,3 +165,61 @@ def _mesos_files(master, slave_id):
...
+
+    if master:
+        host = 'leader.mesos'
+    else:
+        summary = mesos.DCOSClient().get_state_summary()
+        slave_obj = None
+        for slave_ in summary['slaves']:
+            if slave_['id'] == slave:
+                slave_obj = slave_
+                break
```
tamarrow: can we do this as a list comprehension?

cli/tests/integrations/test_node.py
```diff
@@ -15,16 +20,24 @@ def test_help():
...
+    --json                  Print json-formatted nodes
+    --follow                Output data as the file grows
+    --lines=N               Output the last N lines [default: 10]
+    --master                Access the leading master
+    --slave=<slave-id>      Access the slave with the provided ID
+    --version               Show version
+    --option SSHOPT=VAL     SSH option (see `man ssh_config`)
+    --config-file=<path>    Path to ssh config file
+    --user=<user>           SSH user [default: core]
+    --version               Show version
```
tamarrow: `version` twice here too

cli/dcoscli/node/main.py
tamarrow: mkay. why do we have --option and SSHOPT? I'm thinking it should be more analogous to how we specify multiple options in marathon `dcos marathon app update [--force] <app-id> [<properties>...]`

cli/tests/integrations/test_node.py
```diff
@@ -103,13 +116,73 @@ def test_node_log_invalid_lines():
...
+
+def test_node_ssh_user():
+    stdout, stderr = _node_ssh_output(
+        ['--master', '--user=bogus', '--option', 'PasswordAuthentication=no'])
+    assert stdout == b''
+    assert stderr.startswith(b'Permission denied')
+
+
+def test_node_ssh_no_agent():
+    pass
```
tamarrow: This is unused?

dcos/mesos.py
```diff
@@ -40,8 +40,21 @@ def __init__(self):
         else:
             self._mesos_master_url = mesos_master_url
 
+    def get_dcos_url(self, path):
```
tamarrow: Couldn't we just use `util.get_config_vals`

cli/tests/integrations/test_node.py
```diff
@@ -103,13 +116,73 @@ def test_node_log_invalid_lines():
...
+                            stdin=slave,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            preexec_fn=os.setsid,
+                            close_fds=True,
+                            shell=True)
+    os.close(slave)
+
+    # wait for the ssh connection
+    time.sleep(5)
```
tamarrow: can we ping for a certain amount of time instead of sleeping arbitrarily?

cli/dcoscli/node/main.py
mgummelt: Those are required.  These are optional.  I don't think we should have optional positional arguments.

cli/dcoscli/node/main.py
mgummelt: I don't think so

cli/tests/integrations/test_node.py
mgummelt: whoops

cli/tests/integrations/test_node.py
mgummelt: I don't know what we would ping

dcos/mesos.py
mgummelt: we already loaded the config in the constructor, so this is using the already-loaded dcos_url, and failing if we can't find it.  We don't fail in the constructor because you're allowed to instantiate this class without dcos_url set, but you can't use this method.

mgummelt:
@tamarrow: updated

cli/dcoscli/node/main.py
tamarrow: I was thinking we could just do this http://www.goodmami.org/2013/01/30/Getting-only-the-first-match-in-a-list-comprehension.html

cli/dcoscli/node/main.py
tamarrow: ^why both?

cli/tests/integrations/test_node.py
tamarrow: we could wait for a response from ssh `echo hi` or something.

dcos/mesos.py
tamarrow: ok

tamarrow: why do we allow a user to instantiate without dcos_url? What can they do without it?

cli/dcoscli/node/main.py
mgummelt: oh yea, good idea

cli/tests/integrations/test_node.py
mgummelt: I don't know what you mean

dcos/mesos.py
mgummelt: access mesos endpoints, e.g. state.json

cli/dcoscli/node/main.py
mgummelt: both what?

mgummelt:
@tamarrow: updated

cli/tests/integrations/test_node.py
tamarrow: nvm, i don't really care. if this causes problem with the tests we can change later.

tamarrow:
LGTM. merge when tests pass

PR #263, opened 2015-07-08
[ADD] DCOS-646 Add --colors flag to force JSON syntax highlighting.

YuraTolochkevich:


tamarrow:
shouldn't we only allow this if the user also specifies `--json`?

YuraTolochkevich:
@tamarrow yes, we should only allow  --colors only with --json.  PR is updated

ssk2:
@YuraTolochkevich - apologies for the delay. We discussed this flag and decided that we should avoid adding `--colors`. Instead, we should only print colors if this is printing to a TTY or sending to the default pager. If the pager environment variable has been overwrriten or we are not printing to a TTY, we just won't bother with color. 

It looks like this might be the behaviour already - can you confirm this? Apologies for the wasted effort - we had some confusion over what was needed (this is an old ticket!).

YuraTolochkevich:
right, this behavior  is implemented already. Let me to discard this PR?

ssk2:
@YuraTolochkevich - yes, let's close it. Again apologies!

PR #264, opened 2015-07-09
Require changes to run parallel builds in TC

jsancio:


tamarrow:
so exciting!

jsancio:
cc @mgummelt

bin/start_tests.sh
```diff
@@ -11,7 +11,21 @@ deactivate
 # Move down to the dcoscli package
 cd cli
 
+# overwrite core.dcos_url in the dcos.toml
+cp tests/data/dcos.toml $DCOS_CONFIG
+sed -i "s/change.dcos.url/$VBOX_IP/g" $DCOS_CONFIG
```
tamarrow: where does VBOX_IP get set?

bin/start_tests.sh
```diff
@@ -11,7 +11,21 @@ deactivate
 # Move down to the dcoscli package
 cd cli
 
+# overwrite core.dcos_url in the dcos.toml
+cp tests/data/dcos.toml $DCOS_CONFIG
```
mgummelt: $DCOS_CONFIG is set to tests/data/dcos.toml.  Why do we overwrite here?

bin/start_tests.sh
```diff
@@ -11,7 +11,21 @@ deactivate
...
+sed -i "s/change.dcos.url/$VBOX_IP/g" $DCOS_CONFIG
+
+echo 'Config before tests'
+cat $DCOS_CONFIG
```
mgummelt: why print the before and after?

cli/tests/data/missing_params_dcos.toml
```diff
@@ -2,5 +2,5 @@
 reporting = false
 email = "test@mail.com"
 [package]
-cache = "true"
 sources = [ "git://github.com/mesosphere/universe.git", "https://github.com/mesosphere/universe/archive/master.zip",]
+cache = "true"
```
tamarrow: why are we moving this?

tamarrow:
`grep`ing `172.17.8.101` looks like it's also used in `tests/unit/data/node.txt`, `tests/fixtures/node.py`, `tests/fixtures/service.py`

cli/tests/integrations/test_config.py
```diff
@@ -72,7 +72,7 @@ def test_version():
 
 
 def test_list_property(env):
-    stdout = b"""core.dcos_url=http://172.17.8.101
+    stdout = b"""core.dcos_url=http://change.dcos.url
```
mgummelt: This will fail after the `sed` runs, won't it?

mgummelt:
Those shouldn't be a problem.

bin/start_tests.sh
jsancio: It gets set in the TC agent's configuration. We have a filter so that CLI builds only run on agents that have that env variable set.

bin/start_tests.sh
jsancio: I changed the TC build to make `DCOS_CONFIG` point to `/tmp/teamcity_dcos.toml`. What I think we are doing is coping that file there and running `sed` against `/tmp/teamcity_dcos.toml`. I looked at the content of those files before the tests ran and it looked correct. I couldn't look a the content of those files after the test ran because the build was stuck. Running the build again now.

bin/start_tests.sh
jsancio: For debugging. I will remove this once we get this working.

cli/tests/data/missing_params_dcos.toml
jsancio: Sorry. Side effect of running the tests (reading and saving the same file) in my local env.

cli/tests/integrations/test_config.py
jsancio: Well, it passes for py27 but not for py34. Because of the `env` parameter, this test is running against `test/data/dcos.toml` and not against `/tmp/teamcity_dcos.toml` (TC's DCOS_CONFIG).

jsancio:
Ping @mgummelt @tamarrow . PR is ready for your review!!

mgummelt:
LGTM.  Thanks for fixing

PR #265, opened 2015-07-09
[ADD] DCOS-993 Install script functionality to optionally add dcos to path

ytalashkocv:


mgummelt:
Won't this overwrite their `python` with the virtualenv's `python`?

bin/install/install-dcos-cli.sh
```diff
@@ -4,7 +4,34 @@ set -o errexit -o pipefail
...
+
+add_dcos_path_to_profile()
+{
+    UNAME=`uname`
+    case "$UNAME" in
+        Linux ) RC_NAME=".bashrc";;
+        Darwin ) RC_NAME=".bash_profile";;
+        CYGWIN* ) RC_NAME=".bashrc";;
+        MINGW* ) RC_NAME=".profile";;
+        * ) RC_NAME=".bashrc";;
```
jsancio: This worries me. I don't think this is correct for all platforms. For example Ubuntu uses `.profile` by default. If we create a `.bashrc` for them then `.profile` will not get executed. Pasted from my `.profile`:

```bash
# ~/.profile: executed by the command interpreter for login shells.
# This file is not read by bash(1), if ~/.bash_profile or ~/.bash_login
# exists.
# see /usr/share/doc/bash/examples/startup-files for examples.
# the files are located in the bash-doc package.

# the default umask is set in /etc/profile; for setting the umask
# for ssh logins, install and configure the libpam-umask package.
```

Also, this assumes that they are using bash.

jsancio:
We currently ask the user to source `$ENV_SETUP` at the end of that script: https://github.com/mesosphere/dcos-cli/pull/265/files#diff-d91144fb7701278fd2baa5eed67ff3faR101

I think we need to clean up our messaging. @ssk2 Can you help here?

What about Windows? Is this a Linux/MacOS feature only or should we implement this for Windows too?

Also, do we need to change the DCOS documentation? The DCOS UI? @ssk2 Please follow up on this.

bin/install/install-dcos-cli.sh
ytalashkocv: Yes, right. From `.profile` and `.bashrc`: 
`# ~/.profile: executed by the command interpreter for login shells.` 
and 
`# ~/.bashrc: executed by bash(1) for non-login shells.`
But `.bashrc` also executed too, in case `.profile` executing. From `.profile`:
```bash
# if running bash
if [ -n "$BASH_VERSION" ]; then
    # include .bashrc if it exists
    if [ -f "$HOME/.bashrc" ]; then
	. "$HOME/.bashrc"
    fi
fi
```

ssk2:
@mgummelt - if the path to dcos/bin is appended to their PATH then it shouldn't overwrite their system Python (since it looks in the left most location first, etc.).

bin/install/install-dcos-cli.sh
```diff
@@ -4,7 +4,34 @@ set -o errexit -o pipefail
...
+    esac
+    echo "" >> ~/"$RC_NAME"
+    echo "# add the path to the DCOS binary to your bash profile" >> ~/"$RC_NAME"
+    echo "export PATH=$1:\$PATH" >> ~/"$RC_NAME"
+}
+
+prompt_add_dcos_path_to_profile()
+{
+    while true; do
+        read -p "Do you wish to add the path to the DCOS binary to your bash profile? [yes/no] " ANSWER
```
ssk2: `"Do you wish to add the path to the DCOS binary to your bash profile? [yes/no] "` -> `Modify your bash profile to add DCOS to your PATH? [yes/no] `

ssk2:
@ytalashkocv - can we change this so that if the user has said yes to adding DCOS to their path it prints:
```source ~/.bashrc && dcos help``` (or whatever their bash profile file is). 

If they say no, print the current output (i.e. ` echo "source $ENV_SETUP && dcos help"`)

**Comment by ssk2** [comment]
@jsancio - Adding this to Windows is a good idea too. @ytalashkocv - could you add this to the Powershell script too?

**Comment by ssk2** [comment]
@jsancio, @ytalashkocv  - is it worth checking to see if a user already has DCOS in their path and removing/replacing if so?

**Comment by ssk2** [comment]
@jsancio - DCOS UI doesn't need updating. We will want to update the docs. I have created [DOCS-90](https://mesosphere.atlassian.net/browse/DOCS-90) to track this.

File: bin/install/install-dcos-cli.sh (hunk 2 above)
Comment (ytalashkocv) [inline bin/install/install-dcos-cli.sh]: Done.

**Comment by truskovskiyk** [comment]
@ssk2 PR updated

**Comment by ssk2** [comment]
@truskovskiyk Can you add a new line before and after the modify message (`Modify your bash profile to add DCOS to your PATH? [yes/no]`) please?

File: bin/install/install-dcos-cli.sh [hunk 3]
```diff
@@ -4,7 +4,44 @@ set -o errexit -o pipefail
...
+    UNAME=`uname`
+    case "$UNAME" in
+        Linux ) RC_NAME=".bashrc";;
+        Darwin ) RC_NAME=".bash_profile";;
+        CYGWIN* ) RC_NAME=".bashrc";;
+        MINGW* ) RC_NAME=".profile";;
+        * ) RC_NAME=".bashrc";;
+    esac
+    echo "" >> ~/"$RC_NAME"
+    echo "# add the path to the DCOS binary to your bash profile" >> ~/"$RC_NAME"
```
ssk2: Simplify this message to `# path to the DCOS CLI binary`.

ssk2:
@ytalashkocv - this doesn't quite work as intended since we rely on `DCOS_CONFIG` being set which is done by the env-setup script. However, once #314 lands, this will no longer be necessary - so we should only merge this after that.

bin/install/install-dcos-cli.sh
```diff
@@ -4,7 +4,44 @@ set -o errexit -o pipefail
...
+    echo "export PATH=$1:\$PATH" >> ~/"$RC_NAME"
```
ssk2: We should be adding the CLI to the end of $PATH to maintain precedence of the system Python (otherwise we'll use the virtualenv Python first when the user types in `python`)

ssk2: An example:
```
vagrant@mesos:~/dcos-cli/bin$ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/home/vagrant/dcos-cli/bin
vagrant@mesos:~/dcos-cli/bin$ which python
/usr/bin/python
vagrant@mesos:~/dcos-cli/bin$ PATH=/home/vagrant/dcos-cli/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
vagrant@mesos:~/dcos-cli/bin$ which python
/home/vagrant/dcos-cli/bin/python
```

win_bin/install/install-dcos-cli.ps1
```diff
@@ -100,6 +102,36 @@ dcos config set package.sources '[\"https://github.com/mesosphere/universe/archi
...
+switch -regex ($add_path)
+{
+  "[Yy].*" {AddToPath "$installation_path\Scripts"; break}
+  "[Nn].*" {break}
+  default {PromptAddToPath "$installation_path\Scripts"}
+}
+
 echo "Finished installing and configuring DCOS CLI."
 echo ""
 echo "Run this command to set up your environment and to get started:"
```
ssk2: Please switch the output here depending on whether or not the path has been altered, the same as the Linux script. You should rebase upon master to pick up the tweaks to this script.

ssk2:
@ytalashkocv - the script doesn't appear to work on Windows 8.1 for me:
```
PS C:\dcos-993> wget https://raw.githubusercontent.com/mesosphere/dcos-cli/dcos-993/win_bin/install/install-dcos-cli.ps1
 -OutFile install.ps1
PS C:\dcos-993> .\install.ps1 . http://dcos.com
At C:\dcos-993\install.ps1:109 char:23
+     $NewPath = $OldPath + â?T;â?T + $AddedLocation
+                          ~
You must provide a value expression following the '+' operator.
At C:\dcos-993\install.ps1:109 char:24
+     $NewPath = $OldPath + â?T;â?T + $AddedLocation
+                           ~~~
Unexpected token 'â?T' in expression or statement.
At C:\dcos-993\install.ps1:116 char:73
+     $yes = New-Object System.Management.Automation.Host.ChoiceDescription "&Yes", `
+                                                                            ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double
quotation marks ("&") to pass it as part of a string.
At C:\dcos-993\install.ps1:118 char:72
+     $no = New-Object System.Management.Automation.Host.ChoiceDescription "&No", `
+                                                                           ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double
quotation marks ("&") to pass it as part of a string.
At C:\dcos-993\install.ps1:119 char:14
+         "No, don't add DCOS to PATH."
+                 ~~~~~~~~~~~~~~~~~~~~~
The string is missing the terminator: '.
At C:\dcos-993\install.ps1:106 char:1
+ {
+ ~
Missing closing '}' in statement block.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : ExpectedValueExpression
```

ssk2:
Could we add support for zsh too? You can add the same line to the ~/.zshrc file.

bin/install/install-dcos-cli.sh
ytalashkocv: Done

bin/install/install-dcos-cli.sh
ytalashkocv: Done

win_bin/install/install-dcos-cli.ps1
ytalashkocv: Done

ytalashkocv:
@ssk2 PR Updated. Please re-review.

tamarrow:
@ssk2 please review

jsancio:
@ssk2 Can we merge this?

ssk2:
@jsancio nope, I'd like to test it.

ssk2:
@ytalashkocv - this change is better but I'd like to avoid adding the line multiple times to the users' path. When I installed it twice, my .bashrc had:
```
# path to the DCOS CLI binary
if [[ "$PATH" != *"/home/vagrant/dcos/bin"* ]];
  then export PATH=$PATH:/home/vagrant/dcos/bin;
fi

# path to the DCOS CLI binary
if [[ "$PATH" != *"/home/vagrant/dcos/bin"* ]];
  then export PATH=$PATH:/home/vagrant/dcos/bin;
fi
```
Is there a way to avoid having this twice in the users' .bashrc? Our recommended upgrade procedure is to run this script everytime the user wants to upgrade to a new version of the CLI.

bin/install/install-dcos-cli.sh
```diff
@@ -4,7 +4,54 @@ set -o errexit -o pipefail
...
+    esac
+    write_to_profile "$1" "$RC_NAME"
+    write_to_profile "$1" ".zshrc"
+}
+
+prompt_add_dcos_path_to_profile()
+{
+    while true; do
+        echo ""
+        read -p "Modify your bash profile to add DCOS to your PATH? [yes/no] " ANSWER
```
ssk2: `bash profile` -> `shell profile` since we support multiple shells.

ssk2:
@ytalashkocv - actually you can remove .zshrc support. I think this will add too much complexity.

ssk2:
@ytalashkocv - I'm still seeing errors on Windows. Is there something extra I need to do?

```PS C:\dcos> .\install.ps1 . http://localhost
At C:\dcos\install.ps1:116 char:73
+     $yes = New-Object System.Management.Automation.Host.ChoiceDescription "&Yes", `
+                                                                            ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double
quotation marks ("&") to pass it as part of a string.
At C:\dcos\install.ps1:118 char:72
+     $no = New-Object System.Management.Automation.Host.ChoiceDescription "&No", `
+                                                                           ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double
quotation marks ("&") to pass it as part of a string.
At C:\dcos\install.ps1:119 char:14
+         "No, don't add DCOS to PATH."
+                 ~~~~~~~~~~~~~~~~~~~~~
The string is missing the terminator: '.
At C:\dcos\install.ps1:105 char:1
+ {
+ ~
Missing closing '}' in statement block.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : AmpersandNotAllowed

PS C:\dcos>```

ytalashkocv:
@ssk2 About this comment https://github.com/mesosphere/dcos-cli/pull/265#issuecomment-137886910 I can mach this lines in .bashrc and replace them, but this lines too common and user can also made some changes in his/her .bashrc. Also remove .zshrc support and fix windows install script.

ssk2:
@ytalashkocv - that's a valid point. Let's leave as is for the moment and we'll update the documentation accordingly.

PR #267, opened 2015-07-10
[ADD] DCOS-297 As a user, I want to authenticate with Marathon using …

opolishchuck:
…HTTP 1.1

### DIFF
--- a/cli/dcoscli/data/config-schema/marathon.json
+++ b/cli/dcoscli/data/config-schema/marathon.json
@@ -8,6 +8,16 @@
       "title": "Marathon base URL",
       "description": "Base URL for talking to Marathon. It overwrites the value specified in core.dcos_url",
       "default": "http://localhost:8080"
+    },
+    "username": {
+      "type": "string",
+      "title": "Your Username",
+      "description": "Your Username"
+    },
+    "password": {
+      "type": "string",
+      "title": "Your Password",
+      "description": "Your Password"
     }
   },
   "additionalProperties": false
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -24,7 +24,8 @@ def create_client(config=None):
     marathon_url = _get_marathon_url(config)
 
     if 'marathon.username' in config:
-        http.auth = (config.get('marathon.username'), config.get('marathon.password'))
+        http.auth = (config.get('marathon.username'),
+                     config.get('marathon.password'))
 
     logger.info('Creating marathon client with: %r', marathon_url)
     return Client(marathon_url)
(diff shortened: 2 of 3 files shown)


cli/dcoscli/data/config-schema/marathon.json
```diff
@@ -8,6 +8,16 @@
       "title": "Marathon base URL",
       "description": "Base URL for talking to Marathon. It overwrites the value specified in core.dcos_url",
       "default": "http://localhost:8080"
+    },
+    "username": {
+      "type": "string",
+      "title": "Your Username",
```
ssk2: "Your Username" -> "Marathon Username"

cli/dcoscli/data/config-schema/marathon.json
```diff
@@ -8,6 +8,16 @@
...
+      "description": "Your Username"
+    },
+    "password": {
+      "type": "string",
+      "title": "Your Password",
```
ssk2: "Your Password" -> "Marathon password"

cli/dcoscli/data/config-schema/marathon.json
```diff
@@ -8,6 +8,16 @@
       "title": "Marathon base URL",
       "description": "Base URL for talking to Marathon. It overwrites the value specified in core.dcos_url",
       "default": "http://localhost:8080"
+    },
+    "username": {
+      "type": "string",
+      "title": "Your Username",
+      "description": "Your Username"
```
ssk2: "Username to use with Marathon for HTTP basic access authentication."

cli/dcoscli/data/config-schema/marathon.json
```diff
@@ -8,6 +8,16 @@
...
+    },
+    "password": {
+      "type": "string",
+      "title": "Your Password",
+      "description": "Your Password"
```
ssk2: "Password to use with Marathon for HTTP basic access authentication."

dcos/marathon.py
```diff
@@ -23,6 +23,10 @@ def create_client(config=None):
 
     marathon_url = _get_marathon_url(config)
 
+    if 'marathon.username' in config:
+        http.auth = (config.get('marathon.username'),
```
ssk2: This looks like it will affect anything else that uses HTTP - which most likely won't need authentication set. I.e., If I use Marathon auth, and run a Marathon command, then later try to do a `dcos package update`, http.auth will be set. Would it be better to pass in the auth object when we make HTTP calls so it doesn't stick around?

ssk2:
@opolishchuck - can we have some tests too? You may need to run a longer integration test (install a Marathon app using package install which has auth specified, change Marathon URL and try to connect?)

jsancio:
I am open to suggestions on the right solution but no secrets or passwords in dcos.toml!

jsancio:
Agreed to postpone this feature. Closing for now.

PR #268, opened 2015-07-10
add --package-version option for installing packages

tamarrow:


mgummelt:
This might be a question for @jose, but I don't like exposing this "package version" to the users.  They almost never care.  They may want to install a specific version of the upstream software.

jsancio:
@mgummelt which one is "package verions"? E.g. 0.9.0 for marathon. Do we expose the Universe package version (marathon 0 or 1) anywhere in the CLI? To me that is the only one that is consistent. For example Spark has an application version (E.g. 1.4.0) and a CLI version (E.g. 0.1.0). cc @ConnorDoyle

mgummelt:
@jsancio: This is exposing the package version, e.g. "0", or "1", which I don't like doing.  dpkg, for example, doesn't have this notion of package versions.  Just normal software versions with package-specific suffixes

jsancio:
@mgummelt Yeah. The only difference is that for us a universe package can contain multiple software. `dpkg` gets around this by using dependencies and virtual packages. We will probably implement both so maybe you are correct and we should use the software version in package.json.

tamarrow:
@mgummelt this pr shows the software version not package version. Ie for marathon it would be 0.8.0 not 0 or 1.

jsancio:
+1

mgummelt:
Oh.  Why do you call it "--pkg-version"?

tamarrow:
words are hard. `--software-version`?

mgummelt:
Looking at the code, it seems this is using the package version

tamarrow:
screen shot 2015-07-13 at 2 10 47 pm

cli/dcoscli/package/main.py
```diff
@@ -308,8 +314,16 @@ def _install(package_name, options_path, app_id, cli, app, yes):
               "repositories"
         raise DCOSException(msg)
 
-    # TODO(CD): Make package version to install configurable
-    pkg_version = pkg.latest_version()
+    version_dir = [k for k, v in iteritems(pkg.software_versions())
```
mgummelt: What's your point with that screenshot?  `software_versions()` returns a dict whose values are package versions, so based on that, it seems that `pkg_version` is a package version, not a software version

tamarrow: No, values are software versions: https://github.com/mesosphere/dcos-cli/blob/master/dcos/package.py#L1357

mgummelt: weird.  I could have sworn I just read differently.

OK, sorry for the waste of time

cli/dcoscli/package/main.py
```diff
@@ -5,7 +5,8 @@
     dcos package --info
     dcos package describe [--app --options=<file> --cli] <package_name>
     dcos package install [--cli | [--app --app-id=<app_id>]]
-                         [--options=<file> --yes] <package_name>
+                         [--pkg-version=<pkg_version>] [--options=<file> --yes]
```
mgummelt: how about `--version`?

cli/dcoscli/package/main.py
mgummelt: let's name these variables

mgummelt:
tests?

cli/dcoscli/package/main.py
```diff
@@ -308,8 +314,16 @@ def _install(package_name, options_path, app_id, cli, app, yes):
               "repositories"
         raise DCOSException(msg)
 
-    # TODO(CD): Make package version to install configurable
-    pkg_version = pkg.latest_version()
+    version_dir = [k for k, v in iteritems(pkg.software_versions())
+                   if v == pkg_version]
+    if version_dir:
+        pkg_version = ''.join(version_dir)
```
mgummelt: I don't understand this joining.  What if len(version_dir) > 1?

cli/dcoscli/package/main.py
```diff
@@ -308,8 +314,16 @@ def _install(package_name, options_path, app_id, cli, app, yes):
...
+    else:
+        if pkg_version and not version_dir:
+            emitter.publish(
```
mgummelt: We should probably exit with return code 1 at this point.  The user as explicitly asked for something that can't be done.

cli/dcoscli/package/main.py
ConnorDoyle: How about moving the logic to `Package.get_version(...)` entirely?

cli/dcoscli/package/main.py
ConnorDoyle: Can this work be avoided in case `pkg_version` is `None`?

cli/dcoscli/package/main.py
tamarrow: that's already used :(

tamarrow: using `sw-version` for now.

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py
mgummelt: Hmmm, yea, I prefer "version", and it's possible technically, but we'd have to differentiate in the help message, which is confusing.

I like Jose's suggestion to use "package version" and "package revision".  So can we call this `--package-version`?  Sorry for the back and forth

cli/dcoscli/package/main.py
```diff
@@ -5,7 +5,8 @@
     dcos package --info
     dcos package describe [--app --options=<file> --cli] <package_name>
     dcos package install [--cli | [--app --app-id=<app_id>]]
-                         [--options=<file> --yes] <package_name>
```
mgummelt: We need to standardize on whether we group all optional args into a single pair of brackets, or wrap each one independently.  I like the latter approach, which means we should split up [--options] and [--yes] here

cli/dcoscli/package/main.py
```diff
@@ -14,18 +15,20 @@
...
+                              run non-interactively
+    --all                     Apply the operation to all matching packages
+    --app                     Apply the operation only to the package's
+                              application
+    --app-id=<app-id>         The application id
+    --cli                     Apply the operation only to the package's CLI
+    --options=<file>          Path to a JSON file containing package
+                              installation options
+    --validate                Validate package content when updating sources
+    --sw-version=<sw_version> Software version of the package to install
```
mgummelt: "Package version to install"

cli/dcoscli/package/main.py
```diff
@@ -308,8 +313,14 @@ def _install(package_name, options_path, app_id, cli, app, yes):
               "repositories"
         raise DCOSException(msg)
 
-    # TODO(CD): Make package version to install configurable
-    pkg_version = pkg.latest_version()
+    if sw_version:
```
mgummelt: This function is already too long.  Can you decompose this into a `_get_package_version()` function?

cli/dcoscli/package/main.py
```diff
@@ -276,11 +279,13 @@ def _confirm(prompt, yes):
                     "'{}' is not a valid response.".format(response))
 
 
-def _install(package_name, options_path, app_id, cli, app, yes):
+def _install(package_name, sw_version, options_path, app_id, cli, app, yes):
```
mgummelt: `sw_version` conflicts with an existing variable in this function.  Let's avoid that.

dcos/package.py
```diff
@@ -1384,6 +1385,20 @@ def latest_version(self):
         pkg_versions.sort()
         return pkg_versions[-1]
 
+    def get_package_version(self, software_version):
+        """Returns the first package version for a given software version, or
```
mgummelt: This should probably return the most recent version that supports this software version

cli/tests/integrations/test_package.py
```diff
@@ -224,6 +227,14 @@ def test_install_missing_options_file():
         stderr=b"Error opening file [asdf.json]: No such file or directory\n")
 
 
+def test_install_bad_sw_version():
```
mgummelt: Can you write a test that installs an explicit (not most recent) version?

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py
tamarrow: sure

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py

dcos/package.py
tamarrow: done

cli/tests/integrations/test_package.py
tamarrow: done

tamarrow:
@mgummelt updated

cli/dcoscli/package/main.py
```diff
@@ -14,18 +15,24 @@
...
+    --app                               Apply the operation only to the
+                                        package's application
+    --app-id=<app-id>                   The application id
+    --cli                               Apply the operation only to the
+                                        package's CLI
+    --options=<file>                    Path to a JSON file containing package
+                                        installation options
+    --validate                          Validate package content when updating
+                                        sources
+    --package-version=<package_version> Package version to install
```
mgummelt: @jsancio: Do we really care about the 80-char limit in the help output?  This is getting very smushed.

mgummelt: Let's put these in alphabetic order

dcos/package.py
```diff
@@ -1384,6 +1384,31 @@ def latest_version(self):
         pkg_versions.sort()
         return pkg_versions[-1]
 
+    def get_package_revision(self, package_version=None):
```
mgummelt: Let's generalize the existing `latest_version()`, and move this functionality there

cli/tests/integrations/test_package.py
```diff
@@ -224,6 +231,28 @@ def test_install_missing_options_file():
...
 
+def test_install_specific_version():
+    stdout = (b'We recommend a minimum of one node with at least 2 ',
+              b'CPU\'s and 1GB of RAM available for the Marathon Service.\n',
+              b'Installing package [marathon] version [0.8.1]\n',
+              b'Marathon DCOS Service has been successfully installed!\n',
+              b'\tDocumentation: https://mesosphere.github.io/marathon\n',
+              b'\tIssues: https://github.com/mesosphere/marathon/issues\n\n')
+
+    _package('marathon',
```
mgummelt: You'll want to verify that 0.8.1 was actually installed.

cli/dcoscli/package/main.py
jsancio: It can be more than 80-chars just want to make sure that it prints "nice" when using `--help`. One option is to move this to a file specified here: https://github.com/mesosphere/dcos-cli/blob/master/cli/setup.py#L81

Thoughts?

dcos/package.py
```diff
@@ -1384,6 +1384,31 @@ def latest_version(self):
...
+        :returns: package revision
+        :rtype: int
+        """
+
+        if package_version:
+            software_versions = self.software_versions()
+            revision = next(
+                (pkg_ver for pkg_ver
+                    in reversed(software_versions)
+                    if software_versions[pkg_ver] == package_version), None)
```
ConnorDoyle: This reads badly, no?  Seems like the parameter should be called `software_version` instead, to match the semantics of `package_versions` and `software_versions`.  See comment above.  AFAIK this method name also introduces an alias of package version called package revision... why?

dcos/package.py
```diff
@@ -1384,6 +1384,31 @@ def latest_version(self):
...
+        :type package_version: str
+        :returns: package revision
+        :rtype: int
+        """
+
+        if package_version:
+            software_versions = self.software_versions()
+            revision = next(
+                (pkg_ver for pkg_ver
+                    in reversed(software_versions)
```
ConnorDoyle: Maybe add a note that this depends on `software_versions()` returning an `OrderedDict`.

dcos/package.py
ConnorDoyle: +1, except for naming.

Could `latest_version()` change to `latest_package_version()` and take a parameter named `software_version`?

ConnorDoyle:
Left a couple more comments.  Sorry for the drive-by review @tamarrow, some of this is mess that I left behind originally so I have opinions.

dcos/package.py
ConnorDoyle: Off the cuff, maybe it would help to globally replace all mentions of "package version" with "package number" to disambiguate it from the version of the software in the package.  The original naming is kind of unfortunate.

tamarrow: Yea, there was some discussion on naming, if you scroll to the original comments on the pr. I think we settled on `package-version` to mean `software-version` and `package-revision` for `package number` (ie what is currently `package-version`.

tamarrow: @ConnorDoyle do you feel strongly on `package-number` instead of `package-revision`. I'll go through the code and change the names to whatever we decide to not make the code harder to read.

ConnorDoyle: OK, in that case the other methods need to be renamed to avoid further confusion:

- `software_versions` => `package_versions`
- `package_versions` => `package_revisions`
- `latest_version` => `latest_revision`

ConnorDoyle: No strong opinion on new names (what you mentioned was decided sounds fine to me), it's just clear the original names are causing problems.

cli/dcoscli/package/main.py
mgummelt: I say we move these to data files, and standardize on keeping the doc part (not the option part) to 70-char width for readability.

dcos/package.py
mgummelt: +1 to `package_version` and `package_revision`, and again I don't think `package_revision` should ever be shown to the user

cli/dcoscli/package/main.py
tamarrow: done. I'll put the other docs into similar data files in a separate commit.

cli/tests/integrations/test_package.py
tamarrow: done

dcos/package.py
tamarrow: done

tamarrow:
@mgummelt @ConnorDoyle updated

cli/dcoscli/data/help/package.txt
```diff
@@ -0,0 +1,48 @@
...
+    --all                               Apply the operation to all matching packages
+    --app                               Apply the operation only to the package's
+                                        application
+    --app-id=<app-id>                   The application id
+    --cli                               Apply the operation only to the package's CLI
+    -h, --help                          Show this screen
+    --info                              Show a short description of this subcommand
+    --options=<file>                    Path to a JSON file containing package
+                                        installation options
+    --package-version=<package_version> Package version to install
```
mgummelt: Let's still keep 4 spaces between the option and the description

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
+            revision = next(
+                (pkg_ver for pkg_ver
+                    in reversed(package_versions)
+                    if package_versions[pkg_ver] == package_version), None)
+            if not revision:
+                raise DCOSException(
+                    "No package with version {} available".format(
+                        package_version))
+        else:
+            pkg_revisions = self.package_revisions()
```
mgummelt: `package_revisions` returns strings, not ints.  So the sort is going to be broken for revisions > 9, and the types in the pydoc are wrong

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
 
-    def software_versions(self):
-        """Returns a mapping from the package version to the version of the
-        software described by the package.
+    def package_versions(self):
+        """Returns a mapping from the package revision to the package version.
 
-        :returns: Map from package versions to versions of the softwre.
+        :returns: Map from package revision to package version
         :rtype: dict
```
mgummelt: rtype should be OrderedDict

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
+        :returns: package revision
+        :rtype: int
         """
 
-        pkg_versions = self.package_versions()
+        if package_version:
+            package_versions = self.package_versions()
+            # depends on package_versions() returning an OrderedDict
+            revision = next(
+                (pkg_ver for pkg_ver
```
mgummelt: s/pkg_ver/pkg_rev

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
 
-        :returns: Map from package versions to versions of the softwre.
+        :returns: Map from package revision to package version
         :rtype: dict
         """
 
-        software_package_map = collections.OrderedDict()
-        for v in self.package_versions():
+        package_version_map = collections.OrderedDict()
+        for v in self.package_revisions():
```
mgummelt: /s/v/rev

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
         vs.reverse()
         return vs
 
-    def software_versions(self):
-        """Returns a mapping from the package version to the version of the
-        software described by the package.
+    def package_versions(self):
```
mgummelt: This should probably be called `package_revision_map`

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
+        if package_version:
+            package_versions = self.package_versions()
+            # depends on package_versions() returning an OrderedDict
+            revision = next(
+                (pkg_ver for pkg_ver
+                    in reversed(package_versions)
+                    if package_versions[pkg_ver] == package_version), None)
+            if not revision:
+                raise DCOSException(
+                    "No package with version {} available".format(
```
mgummelt: This is nitpicky, but it's slightly more clear to say "No package available with version [{}]"

dcos/package.py
```diff
@@ -1354,39 +1351,55 @@ def package_versions(self):
...
+            revision = pkg_revisions[-1]
 
-        pkg_versions.sort()
-        return pkg_versions[-1]
+        return revision
 
     def __repr__(self):
 
-        v, error = self.latest_version()
+        v, error = self.latest_package_revision()
```
mgummelt: This code never got fixed.  We don't use `error` anymore.  Also... s/v/rev

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: fixed

dcos/package.py
tamarrow: done

cli/dcoscli/data/help/package.txt
tamarrow: done

tamarrow:
@mgummelt updated

dcos/package.py
```diff
@@ -1340,13 +1340,10 @@ def _data(self, path):
         full_path = os.path.join(self.path, path)
         return util.read_file(full_path)
 
-    def package_versions(self):
-        """Returns all of the available package versions, most recent first.
+    def package_revisions(self):
```
mgummelt: This still doesn't return the most recent first, since it's sorting strings, not ints

tamarrow: oh i fixed it for `latest_package_revision`

mgummelt: It needs to be fixed here.  The docs are incorrect as is.

tamarrow: moved it to here

dcos/package.py
```diff
@@ -1340,60 +1340,66 @@ def _data(self, path):
...
-        the software described by the package.
-
-        :returns: Available versions of this package
+        :returns: Available revisions of this package
         :rtype: [str]
         """
 
         vs = [f for f in os.listdir(self.path) if not f.startswith('.')]
-        vs.reverse()
+        vs.sort(key=int, reverse=True)
```
mgummelt: just have this method return ints.  That's what the versions are anyway.

tamarrow: updated

dcos/package.py
```diff
@@ -1340,60 +1340,66 @@ def _data(self, path):
...
-
-        :returns: Available versions of this package
-        :rtype: [str]
+        :returns: Available revisions of this package
+        :rtype: [int]
         """
 
         vs = [f for f in os.listdir(self.path) if not f.startswith('.')]
         vs.reverse()
         return vs
```
mgummelt: This is [str], not [int]

tamarrow: I think I'm going to keep this as a str b/c it refers to the directory name which is expected to be a str in a bunch of other places.

mgummelt: It doesn't refer to a directory name.  It refers to a revision (hence the name of the function).  We just happen to be extracting it from the directory name.

tamarrow: we store the revisions as directories and they are mainly used as lookups in that respect.

mgummelt: semantically, a revision is an integer.  if you're really set on keeping it as a string, then update the pydoc at least

tamarrow: updated

mgummelt:
The commit message needs to be more user friendly as jose specified this morning

dcos/package.py
```diff
@@ -1340,60 +1340,66 @@ def _data(self, path):
...
+            if not revision:
+                raise DCOSException(
+                    "No package available with version {}".format(
+                        package_version))
+        else:
+            pkg_revisions = self.package_revisions()
 
-        pkg_versions.sort()
-        return pkg_versions[-1]
+            if len(pkg_revisions) is 0:
```
mgummelt: `if not pkg_revisions` is the pythonic way to do this

dcos/package.py
```diff
@@ -1340,60 +1340,66 @@ def _data(self, path):
...
 
-    def latest_version(self):
-        """Returns the latest package version.
+    def latest_package_revision(self, package_version=None):
+        """Returns the most recent package revision, for a
+        given package version if specified.
 
-        :returns: The latest version of this package
+        :param package_version: a given package version
+        :type package_version: str
```
mgummelt: str | None

cli/tests/integrations/test_package.py
```diff
@@ -75,59 +76,16 @@ def _chronos_description(app_ids):
...
-    #   - Local file
-    #   - HTTPS
-    #   - Git
-    sources = [
-      "file:///Users/me/test-registry",
-      "https://my.org/registry",
-      "git://github.com/mesosphere/universe.git"
-    ]
-"""
+    stdout = pkg_resources.resource_string(
```
mgummelt: I think we should keep these decoupled.  Otherwise, we're not testing that `--help` hasn't changed unexpectedly.

cli/dcoscli/data/help/package.txt
```diff
@@ -0,0 +1,48 @@
+Install and manage DCOS package versions
```
mgummelt: why did this change?  they're not managing versions.  they're managing packages.

dcos/package.py
```diff
@@ -1340,60 +1340,66 @@ def _data(self, path):
         full_path = os.path.join(self.path, path)
```
mgummelt: There are other references to "version" that need to be changed to "revision" in this file.  e.g. "def package_json(version)"

cli/dcoscli/package/main.py
```diff
@@ -215,9 +175,9 @@ def _describe(package_name, cli, app, options_path):
         raise DCOSException("Package [{}] not found".format(package_name))
 
     # TODO(CD): Make package version to describe configurable
-    pkg_version = pkg.latest_version()
-    pkg_json = pkg.package_json(pkg_version)
-    version_map = pkg.software_versions()
+    pkg_revision = pkg.latest_package_revision()
+    pkg_json = pkg.package_json(pkg_revision)
+    version_map = pkg.package_revisions_map()
```
mgummelt: s/version_map/revision_map

cli/dcoscli/package/main.py
```diff
@@ -215,9 +175,9 @@ def _describe(package_name, cli, app, options_path):
...
     versions = [version_map[pkg_ver] for pkg_ver in version_map]
```
mgummelt: s/pkg_ver/pkg_rev

cli/dcoscli/data/help/package.txt
tamarrow: I was getting rid of using `software`. `Install and manage DCOS packages`?

cli/dcoscli/data/help/package.txt
```diff
@@ -0,0 +1,48 @@
...
+
+Usage:
+    dcos package --config-schema
+    dcos package --info
+    dcos package describe [--app --options=<file> --cli] <package_name>
```
mgummelt: can you make a JIRA to add `--package-version` to this command?

cli/dcoscli/data/help/package.txt
mgummelt: That "software" didn't have anything to do with versions, so yea

cli/dcoscli/package/main.py
```diff
@@ -321,15 +283,15 @@ def _install(package_name, options_path, app_id, cli, app, yes):
...
-    options = pkg.options(pkg_version, user_options)
+    options = pkg.options(pkg_revision, user_options)
 
-    if app and pkg.has_marathon_definition(pkg_version):
+    if app and pkg.has_marathon_definition(pkg_revision):
         # Install in Marathon
-        version_map = pkg.software_versions()
-        sw_version = version_map.get(pkg_version, '?')
+        version_map = pkg.package_revisions_map()
+        package_version = version_map.get(pkg_revision, '?')
```
mgummelt: what's this '?' for?

cli/tests/integrations/test_package.py
```diff
@@ -686,7 +671,8 @@ def _helloworld():
 
 @contextlib.contextmanager
 def _package(name,
-             stdout=b''):
+             stdout=b'',
+             args=['--yes']):
```
mgummelt: update the docstring

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

cli/tests/integrations/test_package.py
tamarrow: is there a good way of excluding sections from the syntax test? If not, it will fail the > 80 character limit

mgummelt: let's just put it in a tests/ text file

tamarrow: done

cli/tests/integrations/test_package.py
tamarrow: done

cli/dcoscli/data/help/package.txt
tamarrow: done

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py
tamarrow: I guess to represent us not knowing the `package_verison`? Don't know when this would happen.

mgummelt:
@tamarrow: ping

tamarrow:
sorry, working on a support issue.

tamarrow:
@mgummelt updated

dcos/package.py
```diff
@@ -1342,60 +1345,66 @@ def _data(self, path):
...
+    def latest_package_revision(self, package_version=None):
+        """Returns the most recent package revision, for a
+        given package version if specified.
 
-        :returns: The latest version of this package
-        :rtype: str
+        :param package_version: a given package version
+        :type package_version: str
+        :returns: package revision
+        :rtype: str | None
```
mgummelt: I don't think this can return None

dcos/package.py
```diff
@@ -1342,60 +1345,66 @@ def _data(self, path):
...
+    def package_revisions(self):
+        """Returns all of the available package revisions, most recent first.
 
-        Note that the result does not describe versions of the package, not
-        the software described by the package.
-
-        :returns: Available versions of this package
+        :returns: Available revisions of this package
         :rtype: [str]
         """
```
mgummelt: you can do this with a list comprehension

dcos/package.py
```diff
@@ -1342,60 +1345,66 @@ def _data(self, path):
...
+        if package_version:
+            pkg_rev_map = self.package_revisions_map()
+            # depends on package_revisions() returning an OrderedDict
+            revision = next(
+                (pkg_rev for pkg_rev
+                    in reversed(pkg_rev_map)
+                    if pkg_rev_map[pkg_rev] == package_version), None)
+            if not revision:
+                raise DCOSException(
+                    "No package available with version {}".format(
```
mgummelt: Can you please include the name in this exception message?

cli/dcoscli/package/main.py
```diff
@@ -321,15 +283,15 @@ def _install(package_name, options_path, app_id, cli, app, yes):
...
-    options = pkg.options(pkg_version, user_options)
+    options = pkg.options(pkg_revision, user_options)
 
-    if app and pkg.has_marathon_definition(pkg_version):
+    if app and pkg.has_marathon_definition(pkg_revision):
         # Install in Marathon
-        version_map = pkg.software_versions()
-        sw_version = version_map.get(pkg_version, '?')
+        revision_map = pkg.package_revisions_map()
+        package_version = revision_map.get(pkg_revision, '?')
```
mgummelt: The map should always have key `pkg_revision`, since the `has_marathon_definition(pkg_revision)` is true

cli/dcoscli/package/main.py
```diff
@@ -321,15 +283,15 @@ def _install(package_name, options_path, app_id, cli, app, yes):
...
 
         message = 'Installing package [{}] version [{}]'.format(
```
mgummelt: This should say "Installing Marathon app for package ..."

cli/dcoscli/package/main.py
```diff
@@ -339,17 +301,17 @@ def _install(package_name, options_path, app_id, cli, app, yes):
...
-            pkg_version,
+            pkg_revision,
             init_client,
             options,
             app_id)
 
-    if cli and pkg.has_command_definition(pkg_version):
+    if cli and pkg.has_command_definition(pkg_revision):
         # Install subcommand
         emitter.publish('Installing CLI subcommand for package [{}]'.format(
```
mgummelt: Include the version in the message

cli/dcoscli/package/main.py
```diff
@@ -309,10 +272,9 @@ def _install(package_name, options_path, app_id, cli, app, yes):
               "repositories"
         raise DCOSException(msg)
 
-    # TODO(CD): Make package version to install configurable
-    pkg_version = pkg.latest_version()
+    pkg_revision = pkg.latest_package_revision(package_version)
 
-    pre_install_notes = pkg.package_json(pkg_version).get('preInstallNotes')
+    pre_install_notes = pkg.package_json(pkg_revision).get('preInstallNotes')
```
mgummelt: set `pkg_json` to a variable.  You use it below as well.

cli/tests/integrations/test_package.py
```diff
@@ -224,6 +182,34 @@ def test_install_missing_options_file():
...
+              b'Installing package [marathon] version [0.8.1]\n'
+              b'Marathon DCOS Service has been successfully installed!\n\n'
+              b'\tDocumentation: https://mesosphere.github.io/marathon\n'
+              b'\tIssues: https:/github.com/mesosphere/marathon/issues\n\n')
+
+    with _package('marathon',
+                  stdout=stdout,
+                  args=['--yes', '--package-version=0.8.1']):
+
+        returncode_, stdout_, stderr_ = exec_command(
```
mgummelt: don't use underscores.  there's no name conflict

dcos/marathon.py
```diff
@@ -110,9 +110,9 @@ def _create_url(self, path):
 
         return urllib.parse.urljoin(self._base_url, path)
 
-    def get_version(self):
-        """Get marathon version
-        :returns: marathon version
+    def get_package_version(self):
```
mgummelt: This isn't the package version.  It was correct before.

dcos/package.py
```diff
@@ -1048,19 +1048,20 @@ def check_version(self, min_version, max_version):
...
                     self._source.url,
-                    version,
+                    package_version,
                     min_version,
                     max_version))
 
-    def get_version(self):
-        """Returns the version of this registry.
+    def get_package_version(self):
+        """Returns the package version of this registry.
```
mgummelt: This is the version of the registry, not a package.  So it was correct before.  Can you go through again and make sure all of these version renamings are correct?

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py
tamarrow: removed the `?`

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py

cli/tests/integrations/test_package.py
tamarrow: stdout is defined twice in this test

dcos/marathon.py
tamarrow: done

dcos/package.py
tamarrow: yep

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: done

tamarrow:
@mgummelt updated

dcos/package.py
```diff
@@ -1377,60 +1380,63 @@ def _data(self, path):
...
 
-        pkg_versions = self.package_versions()
-
-        if len(pkg_versions) is 0:
-            raise DCOSException(
-                'No versions found for package [{}]'.format(self.name()))
+        if package_version:
+            pkg_rev_map = self.package_revisions_map()
+            # depends on package_revisions() returning an OrderedDict
+            revision = next(
```
mgummelt:
```if pkg_rev in pkg_rev_map:
   return next(...)
else:
   return None```

File: dcos/package.py [hunk 48]
```diff
@@ -1377,60 +1380,63 @@ def _data(self, path):
...
+                    in reversed(pkg_rev_map)
+                    if pkg_rev_map[pkg_rev] == package_version), None)
+            if not revision:
+                return None
+        else:
+            pkg_revisions = self.package_revisions()
 
-        pkg_versions.sort()
-        return pkg_versions[-1]
+            if not pkg_revisions:
```
mgummelt: shouldn't ever happen

cli/dcoscli/package/main.py
```diff
@@ -215,22 +175,26 @@ def _describe(package_name, cli, app, options_path):
         raise DCOSException("Package [{}] not found".format(package_name))
 
     # TODO(CD): Make package version to describe configurable
-    pkg_version = pkg.latest_version()
-    pkg_json = pkg.package_json(pkg_version)
-    version_map = pkg.software_versions()
-    versions = [version_map[pkg_ver] for pkg_ver in version_map]
+    pkg_revision = pkg.latest_package_revision()
+    if pkg_revision is None:
```
mgummelt: shouldn't ever happen

cli/dcoscli/package/main.py
```diff
@@ -215,22 +175,26 @@ def _describe(package_name, cli, app, options_path):
...
+        raise DCOSException(
+            'No versions found for package [{}]'.format(package_name))
+
+    pkg_json = pkg.package_json(pkg_revision)
+    revision_map = pkg.package_revisions_map()
+    pkg_versions = [revision_map[pkg_rev] for pkg_rev in revision_map]
```
mgummelt: `pkg_versions = pkg.package_revisions_map().values()`

cli/tests/integrations/test_package.py
mgummelt: Ah, well in other tests we just overwrite these variables.  Let's stay consistent.

dcos/marathon.py
```diff
@@ -117,8 +117,8 @@ def _create_url(self, path):
         return urllib.parse.urljoin(self._base_url, path)
 
     def get_version(self):
-        """Get marathon version
-        :returns: marathon version
+        """Get marathon package version
```
mgummelt: Looks like you didn't revert this.

dcos/package.py
tamarrow: done

dcos/package.py
tamarrow: removed

dcos/marathon.py
tamarrow: removed

cli/dcoscli/package/main.py

cli/dcoscli/package/main.py
tamarrow: removed

cli/tests/integrations/test_package.py
tamarrow: ok

cli/dcoscli/package/main.py
tamarrow: `.values()` returns `ValuesView` in python 3.4, so I'm going to leave it as is.

tamarrow:
@mgummelt updated

cli/dcoscli/package/main.py
```diff
@@ -215,22 +175,22 @@ def _describe(package_name, cli, app, options_path):
...
 
     # TODO(CD): Make package version to describe configurable
-    pkg_version = pkg.latest_version()
-    pkg_json = pkg.package_json(pkg_version)
-    version_map = pkg.software_versions()
-    versions = [version_map[pkg_ver] for pkg_ver in version_map]
+    pkg_revision = pkg.latest_package_revision()
+    pkg_json = pkg.package_json(pkg_revision)
+    revision_map = pkg.package_revisions_map()
+    pkg_versions = [revision_map[pkg_rev] for pkg_rev in revision_map]
```
mgummelt: `pkg_versions = list(revision_map.values())`

mgummelt:
LGTM when tests pass

jsancio:
Wohoo! :fireworks:

PR #270, opened 2015-07-13
[DCOS-1133] dcos service log

mgummelt:
Follows the pattern of `dcos task log` and `dcos node log`

For marathon, uses ssh+journalctl

For non-marathon services, first we find the app running the service, then we find the task for that app, and print the logs for that task.

### DIFF
--- a/cli/dcoscli/node/main.py
+++ b/cli/dcoscli/node/main.py
@@ -18,7 +18,7 @@ Options:
     --master                Access the leading master
     --slave=<slave-id>      Access the slave with the provided ID
     --option SSHOPT=VAL     SSH option (see `man ssh_config`)
-    --config-file=<path>    Path to ssh config file
+    --config-file=<path>    Path to SSH config file
     --user=<user>           SSH user [default: core]
     --version               Show version
 """
@@ -186,12 +186,7 @@ def _ssh(master, slave, option, config_file, user):
 
     """
 
-    ssh_options = ' '.join('-o {}'.format(opt) for opt in option)
-
-    if config_file:
-        ssh_config = '-F {}'.format(config_file)
-    else:
-        ssh_config = ''
+    ssh_options = util.get_ssh_options(config_file, option)
 
     if master:
         host = mesos.MesosDNSClient().hosts('leader.mesos.')[0]['ip']
@@ -205,9 +200,8 @@ def _ssh(master, slave, option, config_file, user):
         else:
             raise DCOSException('No slave found with ID [{}]'.format(slave))
 
-    cmd = "ssh -t {0} {1} {2}@{3}".format(
+    cmd = "ssh -t {0} {1}@{2}".format(
         ssh_options,
-        ssh_config,
         user,
         host)
 
--- a/cli/dcoscli/service/main.py
+++ b/cli/dcoscli/service/main.py
@@ -3,30 +3,46 @@
 Usage:
     dcos service --info
     dcos service [--inactive --json]
+    dcos service log [--follow --lines=N --ssh-config-file=<path>]
+                     <service> [<file>]
     dcos service shutdown <service-id>
 
 Options:
-    -h, --help    Show this screen
+    -h, --help                  Show this screen
 
-    --info        Show a short description of this subcommand
+    --info                      Show a short description of this subcommand
 
-    --json        Print json-formatted services
+    --ssh-config-file=<path>    Path to SSH config file.  Used to access
+                                marathon logs.
 
-    --inactive    Show inactive services in addition to active ones.
-                  Inactive services are those that have been disconnected from
-                  master, but haven't yet reached their failover timeout.
+    --follow                    Output data as the file grows
 
-    --version     Show version
+    --inactive                  Show inactive services in addition to active
+                                ones. Inactive services are those that have
+                                been disconnected from master, but haven't yet
+                                reached their failover timeout.
+
+    --json                      Print json-formatted services
+
+    --lines=N                   Output the last N lines [default: 10]
+
+    --version                   Show version
 
 Positional Arguments:
-    <service-id>  The ID for the DCOS Service
+    <file>                      Output this file. [default: stdout]
+
+    <service>                   The DCOS Service name.
+
+    <service-id>                The DCOS Service ID
 """
+
+import subprocess
 
 import dcoscli
 import docopt
-from dcos import cmds, emitting, mesos, util
+from dcos import cmds, emitting, marathon, mesos, package, util
 from dcos.errors import DCOSException
-from dcoscli import tables
+from dcoscli import log, tables
 
 logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
@@ -57,6 +73,13 @@ def _cmds():
     """
 
     return [
+
+        cmds.Command(
+            hierarchy=['service', 'log'],
+            arg_keys=['--follow', '--lines', '--ssh-config-file', '<service>',
+                      '<file>'],
+            function=_log),
+
         cmds.Command(
             hierarchy=['service', 'shutdown'],
             arg_keys=['<service-id>'],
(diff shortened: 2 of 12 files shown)


cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+        journalctl_args += '-f '
+    if lines:
+        journalctl_args += '-n {} '.format(lines)
+
+    leader_ip = marathon.create_client().get_leader().split(':')[0]
+    master_public_ip = mesos.DCOSClient().metadata()['PUBLIC_IPV4']
+    user = 'core'
+
+    cmd = ("ssh -A -t {3} {0}@{1} " +
+           "ssh -A -t {0}@{2} " +
```
jsancio: Let's come to an agreement on the -A option. Also, you are running `journalctl` so do you want the `-T` (disable pty) or `-t` (enable pty)?

mgummelt: good point.  fixed.

cli/dcoscli/node/main.py
```diff
@@ -186,21 +185,12 @@ def _ssh(master, slave, option, config_file, user):
...
 
     master_public_ip = mesos.DCOSClient().metadata()['PUBLIC_IPV4']
-    ssh_options = ' '.join('-o {}'.format(opt) for opt in option)
 
-    if config_file:
-        ssh_config = '-F {}'.format(config_file)
-    else:
-        ssh_config = ''
+    error = ('`dcos node ssh` depends on `ssh-agent` so your private key '
+             'can be used safely to hop between nodes in your cluster')
```
jsancio: Let's make sure that we change this wording with the new implementation.

cli/dcoscli/node/main.py
```diff
@@ -23,7 +23,6 @@
     --version               Show version
 """
 
-import os
```
jsancio: Writing the comment here but it applies to the `--config-file` option above. For my own edification why do we have a special flag and not make it part of `--option`.

Also, Is it too late to change it to `--ssh-option` instead of `--option`?

jsancio: I see. We only support `-o` or ssh_config options with `--option`. Should we support all of the `ssh` command line options? I like how `mosh` solved this problem: https://mosh.mit.edu/#usage. Search for `-ssh`.

cli/dcoscli/node/main.py
```diff
@@ -214,9 +204,8 @@ def _ssh(master, slave, option, config_file, user):
         else:
             raise DCOSException('No slave found with ID [{}]'.format(slave))
 
-    cmd = "ssh -A -t {0} {1} {2}@{3} ssh -A -t {2}@{4}".format(
+    cmd = "ssh -A -t {0} {1}@{2} ssh -A -t {1}@{3}".format(
```
jsancio: Skipping for now. I think this will change after our latest discussion.

cli/dcoscli/service/main.py
```diff
@@ -3,30 +3,46 @@
...
 Options:
-    -h, --help    Show this screen
+    -h, --help              Show this screen
 
-    --info        Show a short description of this subcommand
+    --info                  Show a short description of this subcommand
 
-    --json        Print json-formatted services
+    --config-file=<path>    Path to SSH config file.  Used to access marathon
+                            logs.
```
jsancio: I like to keep it consistent with the existing `dcos ... ssh` commands but let's consider renaming this to `--ssh-config-file`. Or if we allow them to change any part of the ssh command: `--ssh` or `--ssh-option`. See `mosh`: https://mosh.mit.edu/#usage

cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+def _log(follow, lines, config_file, service, file_):
+    """Prints the contents of the logs for a given service.  The service
+    task is located by first identifying the marathon app running a
+    framework named `service`.
+
+    :param follow: same as unix tail's -f
+    :type follow: bool
+    :param lines: number of lines to print
+    :type lines: int
+    :param config_file: SSH config file.  Used for marathon.
```
jsancio: What do you mean by `Used for marathon`?

cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+    :rtype: int
+    """
+
+    lines = util.parse_int(lines)
+
+    if service == 'marathon':
+        if file_:
+            raise DCOSException('The <file> argument is invalid for marathon.'
+                                ' The systemd journal is always used for the'
+                                ' marathon log.')
```
jsancio: Okay. I am going to work on a proposal for more detail documentation of command when the user types: `dcos <commmand> --help`

cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+
+        return _log_marathon(follow, lines, config_file)
+    else:
+        if config_file:
+            raise DCOSException(
+                'The `--config-file` argument is invalid for non-marathon '
+                'services. SSH is not used.')
+        return _log_service(follow, lines, service, file_)
```
jsancio: :+1: Very nice validation of inputs and error reporting!

cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+    :rtype: dict
+    """
+
+    marathon_client = marathon.create_client()
+    app = _get_service_app(marathon_client, service_name)
+    tasks = marathon_client.get_app(app['id'])['tasks']
+    if len(tasks) != 1:
+        raise DCOSException(
+            ('We expected marathon app [{}] to be running 1 task, but we ' +
+             'instead found {} tasks').format(app['id'], len(tasks)))
```
jsancio: :+1:

cli/dcoscli/service/main.py
```diff
@@ -123,3 +146,170 @@ def _shutdown(service_id):
...
+                service_name,
+                ', '.join('[{}]'.format(app['id']) for app in apps)))
+    elif len(apps) == 0:
+        raise DCOSException(
+            'No marathon apps found for service [{}]'.format(service_name))
+    else:
+        return apps[0]
+
+
+def _log_marathon(follow, lines, config_file):
```
jsancio: Skipping this method for now as I assume that it will change.

cli/tests/integrations/common.py
```diff
@@ -380,6 +417,25 @@ def app(path, app_id, deploy=False):
...
+    exit.
+
+    :param package_name: package name
+    :type package_name: str
+    :param deploy: If True, block on the deploy
+    :type deploy: bool
+    :rtype: None
+    """
+
+    package_install(package_name, deploy, args)
```
jsancio: Is there a reason why this is outside the `try`?

cli/tests/integrations/common.py
```diff
@@ -380,6 +417,25 @@ def app(path, app_id, deploy=False):
...
+    try:
+        yield
```
jsancio: For my own edification can a yield throw/raise? Under what condition will it throw?

cli/tests/integrations/common.py
```diff
@@ -394,3 +450,36 @@ def mock_args(args):
             yield sys.stdout, sys.stderr
         finally:
             sys.stdout, sys.stderr = stdout, stderr
+
+
+def ssh_agent_output(cmd):
```
jsancio: Need to understand this method better. I'll talk to you offline.

dcos/package.py
```diff
@@ -1437,3 +1437,18 @@ def as_dict(self):
...
+    :param framework_name: framework name
+    :type framework_name: str
+    :param client: marathon client
+    :type client: marathon.Client
+    :rtype: [dict]
+    """
+
+    return [app for app in client.get_apps()
+            if app.get('labels', {}).get(
+                PACKAGE_FRAMEWORK_NAME_KEY) == framework_name]
```
jsancio: I don't remember which version of Marathon has this but they now support `v2/apps` to include a filter for the AppDefinition returned. We should identify this feature exist and use it. It is specially important under large deployment. I think @tamarrow is using this and may have more details.

dcos/util.py
```diff
@@ -598,3 +595,34 @@ def get_proxy_dict_from_env():
         if value and (name == 'http_proxy' or name == 'https_proxy'):
             proxies[name] = value
     return proxies
+
+
+def get_ssh_options(config_file, options, error):
```
jsancio: Skipping this method for now.

cli/tests/integrations/test_service.py
```diff
@@ -16,29 +20,52 @@ def zk_znode(request):
     return request
 
 
+def setup_module(module):
+    package_install('chronos', True)
+
+
+def teardown_module(module):
```
jsancio: Should we standardize on using these methods instead of `pytest.fixture(scope=module)`? I am okay with either way. There are advantages and disadvantages to either one. I just want us to pick one. cc @tamarrow

tamarrow: fine with me

dcos/package.py
tamarrow: Looks like that was added back in 2014: https://github.com/mesosphere/marathon/commit/8fd01a83b1958e6ea95c842bc84b580bd1285e35#diff-268ca9c8f5855c658cfb04e862b14d2bR372

It's tagged with `v0.7.0`, so we should be fine using it since we tell users marathon is outdated if its not `0.8.1` or above

jsancio: Doesn't look right. That commit is from 09/03/2014.

cli/dcoscli/node/main.py
mgummelt: I'm pretty sure all SSH options have an ssh_config counterpart

mgummelt: I'll change to --ssh-option

mgummelt: I like that mosh approach.  I'll do that.

mgummelt: I like --option better because the "ssh" is implicit.  This is an ssh command.

cli/dcoscli/node/main.py
mgummelt: This is all gone now

cli/dcoscli/service/main.py
mgummelt: I'll change to --ssh-config-file

cli/dcoscli/service/main.py
mgummelt: marathon is the only service requiring SSH to access the logs

cli/tests/integrations/common.py
mgummelt: It's a context manager.  We only `try` below so we can ensure the package is uninstalled if the manager's block throws an exception

cli/tests/integrations/common.py
mgummelt: This is how contextmanagers work when defined with the above decorator.  They yield once to the code block.

cli/tests/integrations/common.py
mgummelt: changed now

cli/tests/integrations/test_service.py
mgummelt: I think they solve different problems.  For the use case of "I want to do X when this module starts and Y when it stops", I think we should use setup/teardown

dcos/package.py
mgummelt: If this exists, can someone point me to the docs?

dcos/util.py
mgummelt: changed

cli/dcoscli/node/main.py
jsancio: > The problem with the mosh approach is that it forces the user to override the entire SSH command if they want to change something. This means that if they want to change the user, for example, they'll have to resolve the marathon leader on their own so they can set the host. This defeats the point of having this nice wrapper command.

Sounds good.

jsancio: > I like --option better because the "ssh" is implicit. This is an ssh command.

I thought the command was `dcos service log` in which case it is not obvious that `--option` refers to ssh options, no?

cli/dcoscli/service/main.py
jsancio: Can we add that to the comment? Something along the lines of `SSH config file; used when service is set to "marathon".`

cli/dcoscli/node/main.py
mgummelt: This file is for `dcos node`, not `dcos service`.  `dcos service` doesn't have any `--option` param.  Only `--ssh-config-file`

dcos/package.py
jsancio: @mgummelt @tamarrow This is the Marathon PR: https://github.com/mesosphere/marathon/pull/1416. Marathon 0.8.2 should have this feature.

cli/tests/integrations/test_service.py
jsancio: Sounds good!

dcos/mesos.py
```diff
@@ -12,18 +12,19 @@
...
-def get_master():
+def get_master(dcos_client=None):
     """Create a Master object using the url stored in the
     'core.mesos_master_url' property if it exists.  Otherwise, we use
     the `core.dcos_url` property
 
-    :param config: user config
-    :type config: Toml
+    :param dcos_client: DCOSClient
+    :type config: DCOSClient | None
```
jsancio: Rename `config` to `dcos_client`

jsancio:
LGTM. Just a few minor comments.

Regarding the label query against Marathon: let's file an issues and fix it in another PR.

PR #271, opened 2015-07-14
dcos-2313 don't crash if schema not correctly formatted

tamarrow:


jsancio:
Thanks! I think we need some tests that exercise the failure case. Probably easiest to write unittest for  the function `_extract_default_values`.

tamarrow:
@jsancio added tests

jsancio:
Let's use `pytest.fixture` to inject test cases. Checkout https://github.com/mesosphere/dcos-cli/blob/master/tests/test_package.py#L12-L35 and https://github.com/mesosphere/dcos-cli/blob/master/tests/test_package.py#L76-L79

tamarrow:
Didn't we decide to no longer use `pytest.fixture`?

jsancio:
> Didn't we decide to no longer use pytest.fixture?

I was referring to using it to teardown and setup modules. I think they are still useful for injecting test cases.

tamarrow:
@jsancio updated

jsancio:
Very nice! LGTM. Will :shipit: when CI passes.

PR #272, opened 2015-07-14
fix return type from [Error] to [str]

tamarrow:
Since the errors get put into util.list_to_err which expects [str].

### DIFF
--- a/dcos/package.py
+++ b/dcos/package.py
@@ -1009,7 +1009,7 @@ class Registry():
         """Validates a package registry.
 
         :returns: Validation errors
-        :rtype: [Error]
+        :rtype: [str]
         """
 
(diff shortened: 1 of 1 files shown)


tamarrow:
@ssk2 bump

ssk2:
@tamarrow, LGTM - make sure the build passes. Does this address a ticket?

tamarrow:
no, I noticed this when I was fixing a different bug. Thanks!

jsancio:
@tamarrow Should we add a test to catch this in the future?

tamarrow:
I think it might be overkill at this point. I'm surprised we didn't catch it because the `pydoc` was also wrong.

PR #273, opened 2015-07-14
add CORE_DCOS_URL

tamarrow:
So we can parameterize `core.dcos_url`.

### DIFF
--- a/bin/start_tests.sh
+++ b/bin/start_tests.sh
@@ -12,8 +12,8 @@ deactivate
 cd cli
 
 # overwrite core.dcos_url in the dcos.toml
-cp tests/data/dcos.toml $DCOS_CONFIG
-sed -i "s/change.dcos.url/$VBOX_IP/g" $DCOS_CONFIG
+cp tests/data/dcos.toml ${DCOS_CONFIG}
+sed -i "s,http://change.dcos.url,${CORE_DCOS_URL},g" ${DCOS_CONFIG}
 
 make clean env
 source env/bin/activate


bin/start_tests.sh
```diff
@@ -12,8 +12,8 @@ deactivate
 cd cli
 
 # overwrite core.dcos_url in the dcos.toml
-cp tests/data/dcos.toml $DCOS_CONFIG
-sed -i "s/change.dcos.url/$VBOX_IP/g" $DCOS_CONFIG
+cp tests/data/dcos.toml ${DCOS_CONFIG}
+sed -i "s/change.dcos.url/${CORE_DCOS_URL}/g" ${DCOS_CONFIG}
```
jsancio: Can we change this to `s/http:\/\/change.dcos.url/${CORE_DCOS_URL}/g` or something like that? I would like to be able to keep it as `dcos_url = http://...` in the `dcos.toml`.

tamarrow: sure

tamarrow:
changed `sed` separator so that don't have to escape either of the urls

tamarrow:
@mgummelt whats the status on getting the ssh tests to optionally run?

mgummelt:
what do you mean "optionally"?  I'm going to make it configurable, but they'll always run

mgummelt:
what problem are you having?

tamarrow:
We're going to have a build that takes in a dcos_url, but we don't want to have aws credentials, so those tests will always fail.

mgummelt:
I don't understand.  What does AWS have to do with our builds?

tamarrow:
for a random `dcos_url` we won't have credentials to ssh to the nodes

mgummelt:
like I said, I am planning on making the credentials configurable.  what do you mean "we don't want to have aws credentials"?

tamarrow:
We don't want teamcity to have multiple aws credentials. for this purpose we just want to pass over the ssh tests. cc @BenWhitehead

mgummelt:
TC doesn't run against AWS

tamarrow:
where is the dcos-image hosted? In any case, most dcos clusters we're testing against run on aws.

mgummelt:
TC runs a local vagrant.

And I don't think that's true, but if you're running against an AWS cluster, you'll be able to configure the tests to use whatever key you want.

tamarrow:
@mgummelt do you know when the nodes test will be configurable? If its not in the current sprint we can close this pr for now.

mgummelt:
not in current sprint

PR #275, opened 2015-07-15
[FIX] DCOS-2360 Error 'NoneType' object has no attribute 'url'

ytalashkocv:


ytalashkocv:
@mgummelt Please review

dcos/http.py
```diff
@@ -29,7 +29,7 @@ def _default_to_exception(response):
        not isinstance(response, requests.exceptions.RequestException):
         return response
 
-    url = response.request.url
+    url = response.request.url if response.request else 'no URL'
```
mgummelt: The bug here was that I was assuming all RequestException objects have a non-None `request` property, which is not true.

So the solution here is to use the `response.request.url` property only for those exceptions that support it, so you should remove this line, and just use `response.request.url` directly in the below references to `url`

ytalashkocv: Done

ytalashkocv:
@mgummelt PR updated. Please review

mgummelt:
LGTM.  Thanks!

PR #280, opened 2015-07-17
[DCOS-2402] Prints the ssh command used in `dcos node ssh` and `dcos service log

mgummelt:
marathon` to help the user debug SSH issues

### DIFF
--- a/cli/dcoscli/node/main.py
+++ b/cli/dcoscli/node/main.py
@@ -28,7 +28,7 @@ import subprocess
 import dcoscli
 import docopt
 from dcos import cmds, emitting, errors, mesos, util
-from dcos.errors import DCOSException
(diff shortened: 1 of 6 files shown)


mgummelt:
Note I don't know how to test this because for some weird reason I can't get the stderr output from the subprocess.  `stderr.read()` always returns an empty string.  It has something to do with attaching `stdin` to the pty.

mgummelt:
@jsancio: tests passing

jsancio:
> Note I don't know how to test this because for some weird reason I can't get the stderr output from the subprocess. stderr.read() always returns an empty string. It has something to do with attaching stdin to the pty.

Can you point me to test that would need fixing? Maybe an extra pair of eyes can figure out what is wrong.

jsancio:
LGTM.

PR #281, opened 2015-07-20
[DCOS-2137] Added a command to list files in a task's sandbox: `dcos task ls`

mgummelt:
I've introduced a `DCOSHTTPException` class, our first `DCOSException` subclass, which simplifies our HTTP exception handling.

I've also refactored `test_task.py` so that most apps are deployed in parallel during `setup_module`.  This makes the tests much faster, and I think we should rewrite all our integration tests to do this.


cli/tests/unit/data/app.txt
```diff
@@ -1,2 +1,2 @@
- ID         MEM   CPUS  TASKS  HEALTH  DEPLOYMENT  CONTAINER  CMD
```
mgummelt: I added an extra space of padding to the tables

mgummelt:
@jsancio: tests passing.  plz review

mgummelt:
@jsancio: ping

jsancio:
Sorry @mgummelt. This is next on my list!

cli/dcoscli/tables.py
```diff
@@ -331,5 +334,71 @@ def slave_table(slaves):
         ('ID', lambda s: s['id'])
     ])
 
-    tb = util.table(fields, slaves, sortby="HOSTNAME")
+    tb = table(fields, slaves, sortby="HOSTNAME")
+    return tb
+
+
+def _format_unix_timestamp(ts):
```
jsancio: pydoc.

dcos/http.py
```diff
@@ -91,10 +64,12 @@ def request(method,
     try:
         with requests.Session() as session:
             response = session.send(request.prepare(), timeout=timeout)
-    except Exception as ex:
-        logger.exception('Error making HTTP request: %r', request.url)
-
-        raise to_exception(ex)
+    except requests.exceptions.ConnectionError as e:
```
jsancio: Let's add a `logger.exception(...)` statement.

dcos/http.py
```diff
@@ -91,10 +64,12 @@ def request(method,
...
+        raise DCOSException('URL [{0}] is unreachable: {1}'.format(
+            e.request.url, e))
+    except requests.exceptions.Timeout as e:
```
jsancio: Let's add a `logger.exception(...)` statement.

dcos/marathon.py
```diff
@@ -22,10 +22,9 @@ def create_client(config=None):
         config = util.get_config()
 
     marathon_url = _get_marathon_url(config)
-    timeout = config.get('core.timeout', http.DEFAULT_TIMEOUT)
 
     logger.info('Creating marathon client with: %r', marathon_url)
-    return Client(marathon_url, timeout)
+    return Client(marathon_url)
```
jsancio: Why are we removing the timeout?

dcos/marathon.py
```diff
@@ -82,7 +78,27 @@ def _to_exception(response):
         msg = '\n'.join(error['error'] for error in errs)
         return DCOSException(_default_marathon_error(msg))
 
-    return DCOSException('Error: {}'.format(response_json['message']))
+    return DCOSException('Error: {}'.format(message))
```
jsancio: :+1:

dcos/marathon.py
```diff
@@ -363,11 +355,39 @@ def scale_app(self, app_id, instances, force=None):
...
-                            timeout=self._timeout)
+        response = _http_req(http.put,
+                             url,
+                             params=params,
+                             json={'instances': int(instances)})
+
+        deployment = response.json()['deploymentId']
+        return (deployment, None)
+
+    def scale_group(self, group_id, scale_factor, force=None):
```
jsancio: I think @ssk2 reverted this change. Do we need to rebase this PR?

jsancio:
Partial review. Looks like this PR overwrites some recent changes to master. Please take a look at the diff. Thanks!

dcos/http.py
mgummelt: Why do this if we're effectively re-raising the exception?  We're not losing any info.

dcos/marathon.py
mgummelt: I commented on this in the already-merged #283 PR, but I don't understand why we complicated the Marathon interface with this.  We're only ever using the default (core.timeout) value, so there's no need to make it configurable at the Marathon level.

dcos/marathon.py
mgummelt: This looks like a merge error.  I'll take this out.

mgummelt:
@jsancio: plz re-review

dcos/http.py
jsancio: In the recent PR for logging caught exception, we don't `logger.exception` if we explicitly catch DCOSException. In these two exception handlers we are raising DCOSException.

dcos/marathon.py
jsancio: I think we were trying to avoid reading and parsing the config file many times. cc @tamarrow.

Also, ideally I would like us to make the modules in the python package `dcos` able to work outside of the CLI. Importing `dcos.config` in module in the package `dcos` will make this difficult.

mgummelt: I'd like to just make the config object global, but if yall are against that, I still think it's fine to read it here, since a (likely cached) file read is marginal compared to an HTTP request.

cli/tests/integrations/test_task.py
```diff
@@ -68,50 +88,37 @@ def test_task():
...
     assert stderr == b''
-    assert len(json.loads(stdout.decode('utf-8'))) == 1
-
-    _uninstall_sleep()
+    assert len(json.loads(stdout.decode('utf-8'))) == NUM_TASKS
 
 
 def test_task_none():
-    assert_command(['dcos', 'task', '--json'],
+    assert_command(['dcos', 'task', 'bogus', '--json'],
```
jsancio: Hmm. It would be nice if we could still test `dcos task --json` with on a DCOS cluster without any tasks. This, in theory, is a different code path, no?

mgummelt: It's not.  `dcos task` just runs the same code path with a filter of `""`

cli/tests/integrations/test_task.py
```diff
@@ -262,28 +259,59 @@ def test_log_master_unavailable():
...
+def test_ls():
+    assert_command(['dcos', 'task', 'ls', 'test-app1'],
+                   stdout=b'stderr  stdout\n')
+
+
+def test_ls_multiple_tasks():
+    stderr = b'Multiple tasks matching ID [test-app]: [test-app], ' + \
+             b'[test-app2]'
+    returncode, stdout, stderr = exec_command(
+        ['dcos', 'task', 'ls', 'test-app'])
```
jsancio: Hmm. Why does this pass? Should this be `b'Multiple tasks matching ID [test-app]: [test-app1], [test-app2]'`. Notice `test-app1`.

mgummelt: Oh wow, I'm just disregarding that variable.  I'll fix it.

cli/tox.ini
```diff
@@ -22,16 +23,16 @@ commands =
...
 
 [testenv:py27-unit]
 commands =
-  py.test -x -vv {env:CI_FLAGS:} tests/unit{posargs}
+  py.test -vv {env:CI_FLAGS:} tests/unit{posargs}
 
 [testenv:py34-unit]
 commands =
-  py.test -x -vv {env:CI_FLAGS:} tests/unit{posargs}
+  py.test -vv {env:CI_FLAGS:} tests/unit{posargs}
```
jsancio: I think @tamarrow wanted this to fail early in TC. I can see the value to both approaches. Can we reach a compromise? This may complicated our build even more but maybe we can have TC fail early and developers can configure this however they wish. Thoughts?

dcos/errors.py
```diff
@@ -5,6 +5,22 @@ class DCOSException(Exception):
     pass
 
 
+class DCOSHTTPException(DCOSException):
```
jsancio: :+1:

cli/tox.ini
mgummelt: We agreed to revert that change.

mgummelt:
@jsancio: updated.  Let me know your opinion on the timeout issue

cli/tox.ini
jsancio: Okay.

dcos/marathon.py
jsancio: I would like that the functionality in the `dcos` package works outside of the context of the CLI. I would like developers to be able to use `dcos.marathon` in their script without needed the CLI. I know that it doesn't work now but I want to make sure we don't add more CLI dependencies to it.

For example, instantiating and using dcos.marathon.Client requires `dcos.http` which requires reading the environment variable DCOS_CONFIG and the file it points to.

jsancio:
> @jsancio: updated. Let me know your opinion on the timeout issue

Replied [here](https://github.com/mesosphere/dcos-cli/pull/281#discussion_r36208407)

dcos/marathon.py
mgummelt: Third-party command developers write commands using the `dcos` library, and not the `dcos-cli` library, which means `dcos` has to know about CLI functionality like e.g. DCOS_CONFIG.  If it didn't, those authors wouldn't be able to write commands that use that functionality.

jsancio: Not third-party commands but ad-hoc scripts that just want to talk to Marathon.

mgummelt:
@jsancio: added timeout back into marathon interface.  plz re-review

jsancio:
LGTM. Thanks! I will merge once tests pass.

PR #282, opened 2015-07-20
accept no_proxy env variable

tamarrow:
The requests library already processes the http{s}_proxy and no_proxy
env variables, and using the `proxies` argument was overwriting, and
thus invalidating the no_proxy env variable.

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -82,8 +82,7 @@ def request(method,
 
     try:
         with requests.Session() as session:
-            response = session.send(request.prepare(), timeout=timeout,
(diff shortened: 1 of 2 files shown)


jsancio:
LGTM. Thanks!

tamarrow:
I'll wait for confirmation from @surdy before merging. Thanks for helping with this @jsancio!

mgummelt:
we're removing proxy support?

tamarrow:
no, we're making it work better! Right now it works as long as you don't use the `no_proxy` env variable to exclude certain domains from the proxy.

PR #284, opened 2015-07-22
Always log error message when catching exception

jsancio:


dcos/jsonitem.py
```diff
@@ -141,6 +144,8 @@ def _parse_object(value):
             raise DCOSException(
                 'Unable to parse {!r} as a JSON object'.format(value))
     except ValueError as error:
+        logger.exception('Error parsing value as a JSON object')
```
tamarrow: can we include `error` in the exception?

tamarrow: nvm

tamarrow:
lgtm after tests pass

PR #289, opened 2015-07-24
Added tests for ssh connect message.

mgummelt:
I figured out why that last PR wasn't testable (stderr flush issue).  totally wasn't worth the ~5 hours of debugging time, but oh well.

### DIFF
--- a/dcos/emitting.py
+++ b/dcos/emitting.py
@@ -75,6 +75,7 @@ def print_handler(event):
 
     elif isinstance(event, errors.Error):
         print(event.error(), file=sys.stderr)
+        sys.stderr.flush()
 
     elif (isinstance(event, collections.Mapping) or
           isinstance(event, collections.Sequence) or isinstance(event, bool) or
--- a/cli/dcoscli/analytics.py
+++ b/cli/dcoscli/analytics.py
@@ -133,6 +133,7 @@ def _wait_and_capture(subproc):
         line = subproc.stderr.readline().decode('utf-8')
         err += line
         sys.stderr.write(line)
+        sys.stderr.flush()
 
     exit_code = subproc.poll()
 
--- a/cli/dcoscli/node/main.py
+++ b/cli/dcoscli/node/main.py
@@ -205,4 +205,5 @@ def _ssh(master, slave, option, config_file, user):
         host)
 
     emitter.publish(DefaultError("Running `{}`".format(cmd)))
+
     return subprocess.call(cmd, shell=True)
--- a/cli/tests/integrations/test_node.py
+++ b/cli/tests/integrations/test_node.py
@@ -154,6 +154,10 @@ def _node_ssh(args):
     stdout, stderr = _node_ssh_output(args)
 
     assert stdout
+    assert b"Running `" in stderr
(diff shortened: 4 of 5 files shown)


cli/dcoscli/service/main.py
```diff
@@ -305,4 +306,8 @@ def _log_marathon(follow, lines, ssh_config_file):
 
     emitter.publish(DefaultError("Running `{}`".format(cmd)))
 
+    # In the tests, we SIGKILL this process group, and the output
+    # doesn't get flushed unless we explicitly do so here.
+    sys.stderr.flush()
```
jsancio: Looks like python >3.3 added a flush argument for the `print` function. I say we move this flush into the emitter and just call flush every time we print. What do you think? This pattern breaks the abstraction and assumes that `emitter.publish(Error(...))` prints to `sys.stderr`

jsancio:
Good job tracking this down! Just one minor comment.

cli/dcoscli/service/main.py
mgummelt: I'd rather not because it can have adverse performance implications, and we're just doing this to get the tests to work, so I don't really care about a leaky abstraction.

jsancio: Well, I care about leaky abstractions. I think we want to `flush` for every `emitter.publish`. If we want to make a decision based on performance let's measure it so that it is an informed decision.

mgummelt: We don't need to measure IO buffering to know it has a significant impact.  It can be the difference between disk write and RAM write.  It doesn't matter much right now because our output is so small, but it could in the future, so I'd rather not flush every time.

Leaky abstractions are a bad idea because they make it dangerous to modify the implementation.  That isn't the case here.  It's always safe to flush stderr.

If you really want, I'll surface flush=False through the emitter layer, but I don't think that's a good idea either.

jsancio: I like exposing an optional flush parameter for `emitting.publish`. That is what `print` does in newer pythons. Why don't you think it is a good idea?

mgummelt: I don't like complicating an interface just to support a test.  And we'll have to pass this through 3-4 layers of abstraction in `emitting`.  Let me know what you want to do.

jsancio: I suggest that we always `flush` in `emitting.publish` if that becomes a performance problem then we can expose flush in the API.

mgummelt:
@jsancio: plz re-review

dcos/emitting.py
```diff
@@ -74,7 +74,7 @@ def print_handler(event):
         _page(event, pager_command)
 
     elif isinstance(event, errors.Error):
-        print(event.error(), file=sys.stderr)
+        print(event.error(), file=sys.stderr, flush=True)
```
jsancio: Let's see what the tests say but I don't think this works with python 2.7: https://docs.python.org/2.7/library/functions.html#print

mgummelt: ugh.  good point.

mgummelt:
@jsancio: fixed

jsancio:
LGTM.

Issue #291, opened 2015-07-25
"dcos node ssh --master" resolves host to internal IP


gabrielhartmann:
Running 
```sh
dcos node ssh --master --option IdentityFile=/Users/gabriel/.ssh/gabriel.pem --user=core
```
against a test/continuous cluster fails with

```sh
ssh: connect to host 10.0.4.191 port 22: Operation timed out
```

It appears the MesosDNSClient is resolving the internal IP of master.  This is confirmed when looking at the Mesos UI which reports

```sh
Cluster:gabriel-test
Server:10.0.4.191:5050
```

This command should resolve the public IP.

gabrielhartmann:
ssh to DCOS nodes is disabled by design

PR #292, opened 2015-07-25
Fix SSH to DCOS nodes

gabrielhartmann:
Fixing Issue #291 and #293.  The old method of retrieving the Mesos master's ip
was returning the internal private IP.  We now retrieve the public IP
for the master node.  Further, ssh to slaves didn't work because we
weren't being redirected through the master as described in the code
comments.  This PR also fixes that problem.  See issue #293.

The old ssh to master test which used to pass has been removed.  This
change broke the test, but the test didn't actually prove that anything
worked.  I've filed issue #294 to fix the ssh tests.

ssh now works to both master and slave nodes in DCOS clusters.

### DIFF
--- a/cli/dcoscli/node/main.py
+++ b/cli/dcoscli/node/main.py
@@ -23,6 +23,7 @@ Options:
     --version               Show version
 """
 
+import re
 import subprocess
 
 import dcoscli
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
I'm not sure what that test is supposed to validate.  The current method of retrieving the Master node's IP address returns a private IP address, and doesn't result in successful logins on DCOS clusters.  The test should probably be aligned with that goal.  The code in this change successfully logs into the Master node on DCOS clusters.

gabrielhartmann:
ssh to DCOS nodes is disabled by design

PR #296, opened 2015-07-28
Resolve HTTP proxy configuration on the first call

jsancio:
Through manual testing we nocited that the cli wasn't resolving proxy
setting during the first HTTP call in a session. This was because we
were using `session.send` method which doesn't do proxy resolution.
The higher-level methods: `session.request`, `session.get`, etc. do
proxy setting resolution.

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -77,22 +77,21 @@ def request(method,
         kwargs['headers'] = {'Accept': 'application/json'}
 
     timeout = timeout or _timeout()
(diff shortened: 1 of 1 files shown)


tamarrow:
ah this is why the proxy worked before we adding the `no_proxy` support because we were originally using session.request.

tamarrow:
good find! :ship:

jsancio:
Thanks. I will merge it when the tests pass.

PR #299, opened 2015-07-31
ability to configure ssl verification

ryane:
having the ability to turn off ssl verification can be useful when mesos and/or marathon UIs are behind SSL using self-signed certificates

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -12,9 +12,11 @@ def _timeout():
     config = util.get_config()
     return config.get('core.timeout', DEFAULT_TIMEOUT)
 
+
 def _ssl_verify():
     config = util.get_config()
     return config.get('core.ssl_verify', True)
+
 
 def _default_is_success(status_code):
     """Returns true if the success status is between [200, 300).
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -1,6 +1,7 @@
 import requests
 from dcos import util
 from dcos.errors import DCOSException
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
 
 logger = util.get_logger(__name__)
 
@@ -11,6 +12,9 @@ def _timeout():
     config = util.get_config()
     return config.get('core.timeout', DEFAULT_TIMEOUT)
 
+def _ssl_verify():
+    config = util.get_config()
+    return config.get('core.ssl_verify', True)
 
 def _default_is_success(status_code):
     """Returns true if the success status is between [200, 300).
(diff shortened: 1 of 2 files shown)


jsancio:
Thanks for the PR @ryane. It looks like the PR is failing some of the syntax checks. We are looking into exposing the result from those tests to the public. But here is the result for now.

```
E302: expected 2 blank lines, found 1
def _ssl_verify():
^
Separate top-level function and class definitions with two blank lines.

    Method definitions inside a class are separated by a single blank line.

    Extra blank lines may be used (sparingly) to separate groups of related
    functions.  Blank lines may be omitted between a bunch of related
    one-liners (e.g. a set of dummy implementations).

    Use blank lines in functions, sparingly, to indicate logical sections.

    Okay: def a():\n    pass\n\n\ndef b():\n    pass
    Okay: def a():\n    pass\n\n\n# Foo\n# Bar\n\ndef b():\n    pass

    E301: class Foo:\n    b = 0\n    def bar():\n        pass
    E302: def a():\n    pass\n\ndef b(n):\n    pass
    E303: def a():\n    pass\n\n\n\ndef b(n):\n    pass
    E303: def a():\n\n\n\n    pass
    E304: @decorator\n\ndef a():\n    pass
```

```
E302: expected 2 blank lines, found 1
def _default_is_success(status_code):
^
Separate top-level function and class definitions with two blank lines.

    Method definitions inside a class are separated by a single blank line.

    Extra blank lines may be used (sparingly) to separate groups of related
    functions.  Blank lines may be omitted between a bunch of related
    one-liners (e.g. a set of dummy implementations).

    Use blank lines in functions, sparingly, to indicate logical sections.

    Okay: def a():\n    pass\n\n\ndef b():\n    pass
    Okay: def a():\n    pass\n\n\n# Foo\n# Bar\n\ndef b():\n    pass

    E301: class Foo:\n    b = 0\n    def bar():\n        pass
    E302: def a():\n    pass\n\ndef b(n):\n    pass
    E303: def a():\n    pass\n\n\n\ndef b(n):\n    pass
    E303: def a():\n\n\n\n    pass
    E304: @decorator\n\ndef a():\n    pass
```

ryane:
sorry about that. it looks like it is passing now.

dcos/http.py
```diff
@@ -77,6 +83,7 @@ def request(method,
         kwargs['headers'] = {'Accept': 'application/json'}
 
     timeout = timeout or _timeout()
+    ssl_verify = _ssl_verify()
```
jsancio: Hey @ryane, we are trying to make the `dcos` package/module usable outside of the DCOS CLI. For example we would like people to use the package/module write Python scripts against DCOS. Requiring `util.get_config()` to succeed (environment variable DCOS_CONFIG points to a valid `dcos.toml`) is a unnecessary requirements. Can we refactor this to instead pass the value in?

I know that we currently require that with `_timeout()` but I consider that a bug and we will fix that ASAP.

tamarrow:
@ryane we are working on a fix for this issue. Thanks.

PR #302, opened 2015-08-03
dcos-2440 add helpful error message if no http(s) in dcos_url

tamarrow:


dcos/util.py
```diff
@@ -325,16 +325,20 @@ def _format_validation_error(error):
...
-    if match:
+    req_match = re.search("(.+) is a required property", error_message)
+    url_match = re.search("^(https{0,1}://).+", str(error.instance))
+
+    if req_match:
         message = 'Error: missing required property {}.'.format(
-            match.group(1))
+            req_match.group(1))
+    elif path == "core.dcos_url" and not url_match:
+        message = "Error: Please check your dcos_url. Missing http(s)"
```
mgummelt: I don't like having a special case here.  Can we not add this regex to the jsonschema?

tamarrow: The json-schema regex for `dcos_url` is correct, but the error message is unclear. Sadly you can't add custom errors in the json-schema itself.

dcos/util.py
```diff
@@ -325,16 +325,20 @@ def _format_validation_error(error):
...
-    match = re.search("(.+) is a required property", error_message)
-    if match:
+    req_match = re.search("(.+) is a required property", error_message)
+    url_match = re.search("^(https{0,1}://).+", str(error.instance))
+
+    if req_match:
         message = 'Error: missing required property {}.'.format(
-            match.group(1))
+            req_match.group(1))
+    elif path == "core.dcos_url" and not url_match:
```
ssk2: Is there any more general way to check for URL paths?  (I don't know this code well enough - but are these properties typed?)

ssk2:
Can you add a test for this new check pleaes?

dcos/util.py
ssk2: We might want to change this error message. This test case yields an inaccurate error message:
```
➜  cli git:(http) ✗ dcos config set core.dcos_url https:/blah
Error: Please check your dcos_url. Missing http(s)
```

tamarrow: done

dcos/util.py
tamarrow: we use a regex to validate the url in the json-schema.

tamarrow:
updated @ssk2

dcos/util.py
ssk2: Just for this particular property?

ssk2: (I was wondering if we could catch this error in any URL property)

ssk2:
Thanks @tamarrow. Aside from my comment about generality, you're good to merge. Cheers!

dcos/util.py
mgummelt: please at least do this in a more general way.  create a data structure with config vars and their error messages, and check them here.

tamarrow: Ill do a more general refactor in a separate pr

mgummelt: I would really prefer we don't merge code with unnecessary one-offs like this, and in general that doesn't meet our design standards.

But @ssk2 is the reviewer, so I'll let him decide.

I also don't understand why this error checking is in a formatting function

tamarrow:
@ssk2 changed to be more general

jsancio:
Wohoo!! :chart_with_upwards_trend:

tamarrow:
@jsancio bump

tamarrow:
@ssk2 friendly bump

dcos/jsonitem.py
```diff
@@ -226,3 +230,37 @@ def _parse_array(value):
...
+        domain_pattern +
+        r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}))'  # or ip
+        r'(?P<port>(?::\d+))?'  # port
+        r'(?P<path>(?:/?|[/?]\S+))$',  # resource path
+        value, re.IGNORECASE)
+
+    if value_regex is None:
+        scheme_match = re.match(scheme_pattern, value, re.IGNORECASE)
+        if scheme_match is None:
+            msg = 'Please check your dcos_url. Missing http(s)://'
```
jsancio: Why the hardcoded `dcos_url`? The format `uri` could be applied to any field.

tamarrow: updated

tests/test_jsonitem.py
```diff
@@ -75,6 +85,7 @@ def jsonitem_tuple(request):
         ('boolean', 'null', None),
         ('array', '[1,2,3]', [1, 2, 3]),
         ('array', 'null', None),
+        ('url', 'http://test.com', 'http://test.com'),
```
jsancio: This is not getting tested. The `params` above is set `range(12)`. I like this pattern: https://github.com/mesosphere/dcos-cli/blob/master/cli/tests/unit/test_marathon_validation.py#L22-L31

tests/test_jsonitem.py
```diff
@@ -60,6 +60,16 @@ def bad_array(request):
...
 
 
+@pytest.fixture(params=range(4))
+def bad_url(request):
+    return [
+        'www.test.com',
+        'http:/hi.com',
+        'https//www.hi.com',
+        'http://bad.port:here'
+    ][request.param]
```
jsancio: I like this pattern: https://github.com/mesosphere/dcos-cli/blob/master/cli/tests/unit/test_marathon_validation.py#L22-L31

Less error prone.

jsancio:
LGTM. Minor comments.

tests/test_jsonitem.py
tamarrow: updated

tests/test_jsonitem.py
tamarrow: nice catch, updated

tamarrow:
thanks! will merge when tests pass

PR #306, opened 2015-08-05
dcos-2523 respect DCOS_LOG_LEVEL and DCOS_DEBUG env variables

tamarrow:


cli/dcoscli/main.py
```diff
@@ -89,43 +91,57 @@ def _main():
...
     :rtype: bool
     """
 
-    if log_level is None:
+    if unset:
         os.environ.pop(constants.DCOS_LOG_LEVEL_ENV, None)
         return True
 
-    log_level = log_level.lower()
+    log_level = ((log_level and log_level.lower()) or
```
mgummelt: if you're trying to do a ternary operation, use `a if b else c`

cli/dcoscli/main.py
```diff
@@ -89,43 +91,57 @@ def _main():
...
+                 os.environ.get(constants.DCOS_LOG_LEVEL_ENV, None))
```
mgummelt: The `None` argument is redundant.  `get` returns None by default if it can't find the key

cli/dcoscli/main.py
```diff
@@ -89,43 +91,57 @@ def _main():
...
     if log_level in constants.VALID_LOG_LEVEL_VALUES:
         os.environ[constants.DCOS_LOG_LEVEL_ENV] = log_level
         return True
 
+    if log_level is None:
```
mgummelt: place this before the previous conditional

cli/dcoscli/main.py
```diff
@@ -89,43 +91,57 @@ def _main():
...
+        return True
+
     msg = 'Log level set to an unknown value {!r}. Valid values are {!r}'
```
mgummelt: This should raise an Exception

cli/dcoscli/main.py
```diff
@@ -89,43 +91,57 @@ def _main():
     subproc = Popen([executable,  command] + args['<args>'],
                     stderr=PIPE)
 
+    if log_level:
```
mgummelt: why unset?

mgummelt:
tests?

cli/dcoscli/main.py
tamarrow: done

cli/dcoscli/main.py
tamarrow: done

cli/dcoscli/main.py
tamarrow: done

cli/dcoscli/main.py
tamarrow: if they specify --log-level we unset the env variable so it doesn't persist logging. We only persist if they explicitly set the env variable

cli/dcoscli/main.py
tamarrow: done

cli/dcoscli/main.py
mgummelt: what do you mean "persist logging"?

tamarrow: persist ie keep logging with that level for all commands until they unset the env variable

mgummelt: env vars are only set for the current process.  they don't persist in the parent process.

tamarrow: yes, but if they set it in that process shouldn't we keep logging on for that process?

tamarrow:
added tests

tamarrow:
@mgummelt simplified pr

cli/dcoscli/main.py
```diff
@@ -101,10 +101,10 @@ def _config_log_level_environ(log_level):
     """
 
     if log_level is None:
```
mgummelt: If this function does nothing on None, you should probably just have the client code check for None, and then not call

cli/dcoscli/main.py
```diff
@@ -124,8 +124,8 @@ def _config_debug_environ(is_debug):
 
     if is_debug:
         os.environ[constants.DCOS_DEBUG_ENV] = 'true'
-    else:
-        os.environ.pop(constants.DCOS_DEBUG_ENV, None)
+
+    return None
```
mgummelt: we usually don't explicitly return None

cli/dcoscli/main.py
```diff
@@ -124,8 +124,8 @@ def _config_debug_environ(is_debug):
 
     if is_debug:
```
mgummelt: this function is so simple, it should probably just be removed, but up to you

mgummelt:
minor comments then LGTM

cli/dcoscli/main.py
tamarrow: done

cli/dcoscli/main.py
tamarrow: removed

cli/dcoscli/main.py
tamarrow: true, removed

tamarrow:
thanks @mgummelt. made changes

PR #311, opened 2015-08-06
Fixed slave id access

mgummelt:
fixes: https://rollbar.com/thomas-mesosphere/dcos/items/1543/occurrences/6961996376/


ssk2:
Mind giving me a brief explanation of what went wrong here?

mgummelt:
I wrote `slave.id` when it should have been `slave['id']`.  This code only gets run when a node query matches multiple nodes, so it doesn't get run much

ssk2:
Ah, okay. Thanks!

PR #313, opened 2015-08-07
add ability to configure Google OAuth 2.0 endpoint as a authentication service

ohal:
add ability to configure Google OAuth 2.0 endpoint as a authentication service
1. move secrets to $DCOS_CONFIG file
2. move to Google OAuth 2.0 endpoint

### DIFF
--- a/cli/dcoscli/data/config-schema/core.json
+++ b/cli/dcoscli/data/config-schema/core.json
@@ -42,6 +42,16 @@
         "minimum": 1,
         "title": "Request timeout in seconds",
         "description": "Request timeout in seconds"
+    },
+    "client_id": {
+      "type": "string",
+      "title": "Your OAuth client_id secret",
+      "description": "Your OAuth client_id secret"
+    },
+    "client_secret": {
+      "type": "string",
+      "title": "Your OAuth client_secret secret",
+      "description": "Your OAuth client_secret secret"
     }
   },
   "additionalProperties": false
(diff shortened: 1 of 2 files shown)


jsancio:
Thank you for the PR @ohal. Our internal integration tests for this PR are failing. We are looking into exposing the result to the public. Until then this is the error:

```
E711: comparison to None should be 'if cond is None:'
        if value == None:
                 ^
Comparison to singletons should use "is" or "is not".

    Comparisons to singletons like None should always be done
    with "is" or "is not", never the equality operators.

    Okay: if arg is not None:
    E711: if arg != None:
    E712: if arg == True:

    Also, beware of writing if x when you really mean if x is not None --
    e.g. when testing whether a variable or argument that defaults to None was
    set to some other value.  The other value might have a type (such as a
    container) that could be false in a boolean context!
```

jsancio:
@ohal This test is failing:

```
('tests/integrations/test_auth.py', 23, 'test_anonymous_login')
def test_anonymous_login():
        with patch('sys.stdin.readline', return_value='\n'), \
                patch('uuid.uuid1', return_value='anonymous@email'):
    
>           assert _mock_dcos_run([util.which('dcos'),
                                   'help'], False) == 0
E           assert 1 == 0
E            +  where 1 = _mock_dcos_run(['/dcos-cli/cli/.tox/py27-integration/bin/dcos', 'help'], False)

tests/integrations/test_auth.py:28: AssertionError
------- Stderr: -------
Missing required config parameter: "core.client_id".  Please run `dcos config set core.client_id <value>`.
```

jsancio:
No activity for a few days. I am closing this PR for now.

PR #314, opened 2015-08-07
Removed dependency on DCOS_CONFIG env var

mgummelt:


dcos/util.py
```diff
@@ -92,6 +92,22 @@ def ensure_dir(directory):
                 'Cannot create directory [{}]: {}'.format(directory, e))
 
 
+def ensure_file(path):
```
ssk2: This isn't descriptive enough alone, perhaps `ensure_file_exists` would be better!

README.rst
```diff
@@ -112,10 +112,11 @@ Running
 Tox will run unit and integration tests in both Python environments using a
 temporarily created virtualenv.
 
-You should ensure :code:`DCOS_CONFIG` is set and that the config file points
-to the Marathon instance you want to use for integration tests. If you're
-happy to use the default test configuration which assumes there is a Marathon
-instance running on localhost, set :code:`DCOS_CONFIG` as follows::
+You can set :code:`DCOS_CONFIG` to a config file that points to the
+Marathon instance you want to use for integration tests. If you're
```
ssk2: While we're updating these docs, perhaps you can change this to say "DCOS cluster" instead of "Marathon instance"?

README.rst
```diff
@@ -112,10 +112,11 @@ Running
...
+happy to use the default test configuration which assumes there is a
+Marathon instance running on localhost, set :code:`DCOS_CONFIG` as
```
ssk2: Does anyone use the default test configuration out of the box? It looks like the dcos_url contains "change.me".

cli/bin/env-setup
```diff
@@ -6,12 +6,8 @@ else
...
 fi
 
+# real, absolute path to BIN_DIR
 FULL_BIN_PATH=$(python -c "import os; print(os.path.realpath('$BIN_DIR'))")
 
+# ensure BIN_DIR is prepended to PATH
 expr "$PATH" : "${FULL_BIN_PATH}.*" > /dev/null || export PATH=$FULL_BIN_PATH:$PATH
-export DCOS_CONFIG=~/.dcos/dcos.toml
-
-if [ ! -f "$DCOS_CONFIG" ]; then
```
ssk2: Just checking - but if the user doesn't have a configuration file, will we still create a new one for them when they run the installation script?

ssk2:
@mgummelt  - a few comments but otherwise looks good.

cli/bin/env-setup
ssk2: Looks like this file is included by install.sh. Please move this creation block into `install-dcos-cli.sh` (looks like the Windows script already does this).

ssk2: Once you've moved the creation block, please remove https://github.com/mesosphere/dcos-cli/blob/master/bin/install/install-dcos-cli.sh#L49-L50.

dcos/util.py
mgummelt: It was consistent with the existing `ensure_dir` function.  I changed both to `*_exists`.

README.rst
mgummelt: good eye.  fixed

cli/bin/env-setup
mgummelt: As you see in this PR, we create the file if `dcos` tries to access it and sees it doesn't exist.  So the install script does not create the file directly.

So we shouldn't need this creation block anywhere.

And `env-setup` is still necessary to update the $PATH, so I don't think we can remove the `source` line you referenced.

ssk2: Ah sorry, I forgot to mention that #324 will add `dcos` to the user's PATH.

ssk2:
Alright, I'm happy! Please merge when the build passes :)

PR #315, opened 2015-08-10
[DCOS-1192] Catch exceptions when parsing config file

mgummelt:


dcos/config.py
```diff
@@ -14,7 +15,11 @@ def mutable_load_from_path(path):
     """
 
     with util.open_file(path) as config_file:
-        return MutableToml(toml.loads(config_file.read()))
+        try:
+            return MutableToml(toml.loads(config_file.read()))
```
tamarrow: Were we going to add the file if it doesn't exist?

mgummelt: that's in a different PR

mgummelt:
moved @jsancio to reviewer because he is the author

dcos/config.py
```diff
@@ -14,7 +15,11 @@ def mutable_load_from_path(path):
...
+        except Exception as e:
```
jsancio: Please add `logger.exception(...)`

dcos/config.py
```diff
@@ -27,7 +32,11 @@ def load_from_path(path):
     """
 
     with util.open_file(path) as config_file:
-        return Toml(toml.loads(config_file.read()))
+        try:
+            return Toml(toml.loads(config_file.read()))
+        except Exception as e:
```
jsancio: Please add `logger.exception(...)`

jsancio:
Thanks. Looks good. Can we add tests for this PR?

dcos/config.py
mgummelt: When in general should we use `logger.exception`?  Can you add that to the review wiki?

I understand adding it if there's valuable info that isn't contained in the exception message, but that's not the case here.

mgummelt:
added test

PR #321, opened 2015-08-10
[DCOS-610] Moved "missing options file" message for `dcos package install

mgummelt:
--options=<missing.json>` to the beginning of stderr

### DIFF
--- a/cli/tests/integrations/test_package.py
+++ b/cli/tests/integrations/test_package.py
@@ -260,8 +260,6 @@ def test_install_missing_options_file():
         ['dcos', 'package', 'install', 'chronos', '--yes',
          '--options=asdf.json'],
         returncode=1,
-        stdout=b'We recommend a minimum of one node with at least 1 CPU and '
-               b'2GB of RAM available for the Chronos Service.\n',
         stderr=b"Error opening file [asdf.json]: No such file or directory\n")
 
 
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -339,6 +339,8 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
             msg = "Package [{}] not available".format(package_name)
         raise DCOSException(msg)
 
+    user_options = _user_options(options_path)
+
     pkg_json = pkg.package_json(pkg_revision)
     pre_install_notes = pkg_json.get('preInstallNotes')
     if pre_install_notes:
@@ -346,8 +348,6 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
         if not _confirm('Continue installing?', yes):
             emitter.publish('Exiting installation.')
             return 0
-
-    user_options = _user_options(options_path)
 
     options = pkg.options(pkg_revision, user_options)
 


cli/tests/integrations/test_package.py
```diff
@@ -671,6 +669,46 @@ def test_search_table():
     assert stderr == b''
 
 
+def test_search_ends_with_wildcard():
```
tamarrow: did you just move these tests? Can we do that in a separate pr please?

mgummelt: the author mistakenly added them to the end of the file, which should be where private functions go.

I often find minor issues like this when on separate tasks.  it's too much overhead to split them up into separate branches.

tamarrow: `git add -p`! I like separate commits b/c it makes the git history cleaner. and is easier to revert if necessary. We can decide whether we not to do this or not cli? @jsancio @ssk2 thoughts?

mgummelt:
We've reverted commits maybe a couple of times in the 4 months I've been here.  But i find a small separate formatting/design issue to fix in almost every PR I write.  The extra context switching cost is not worth any `git revert` benefit.  I would be very strongly against this policy.

cli/tests/integrations/test_package.py
mgummelt: We've reverted commits maybe a couple of times in the 4 months I've been here. But i find a small separate formatting/design issue to fix in almost every PR I write. The extra context switching cost is not worth any git revert benefit. I would be very strongly against this policy.

ssk2: Small typos are fine but I'm wary of moving 40-odd lines up within another PR.

mgummelt: well this is just a copy+paste, so can you please review this one @tamarrow ?

tamarrow: I'd rather you moved it, but the change looks fine.

mgummelt: @jsancio: can you arbitrate this so we don't run into this issue in future PRs?

The question is whether we should allow minor fixes that are otherwise irrelevant to the PR.  Including:

a) docstring fixes
b) pydoc fixes
c) code movement (like this one)
d) naming inconsistencies
e) typos 

@ssk2 mentioned that he's fine with this, but wary of 40 lines changes like this, but this is really just one "thing", since it's a copy+paste.  I'm of course against an irrelevant real 40-line change as well.

As discussed, the tradeoff is:
+ less friction to fixing problems
- more difficult `git revert`

Since I run into these sorts of issues very regularly, and since I can only think of two times we've run `git revert` in the four months I've been here, I think it's clearly beneficial to allow this.  Otherwise, I'm disincentivized from fixing these issues.  I'll fix a minor issue if I can do it immediately, but not if I have to 1) commit my changes to the current branch, 2) create a new branch, 3) run syntax checks, 4) push, 5) write a new PR, 6) switch back to the original branch and try to remember where I was.

tamarrow:
LGTM. merge when tests pass

PR #323, opened 2015-08-11
[DCOS-2435] Improve error message when an app is locked by another deployment.

mgummelt:


dcos/marathon.py
```diff
@@ -56,6 +56,10 @@ def _to_exception(response):
         return DCOSException(
             'Error while fetching [{0}]: HTTP {1}: {2}'.format(
                 response.request.url, response.status_code, response.reason))
+    elif response.status_code == 409:
+        return DCOSException(
+            'App is locked by one or more deployments. '
```
ssk2: `App` -> `App or group` since this can affect either (and in this case I think the Marathon API error message which refers only to an `App` is confusing).

ssk2:
One naming change but otherwise looks good, thanks for fixing this!

dcos/marathon.py
mgummelt: fixed

ssk2:
Great, merge after successful build please.

PR #325, opened 2015-08-11
Moved three tests in test_package.py

mgummelt:


tamarrow:
merge when tests pass. thanks for making it a separate pr :)

PR #327, opened 2015-08-12
DCOS-2262 Send hostname segment.io track events

jsancio:
Verified that the value gets sent by doing manual tests.

### DIFF
--- a/cli/dcoscli/analytics.py
+++ b/cli/dcoscli/analytics.py
@@ -6,6 +6,7 @@ import uuid
 import dcoscli
 import docopt
 import rollbar
+import six
 from concurrent.futures import ThreadPoolExecutor
 from dcos import http, util
 from dcoscli.constants import (ROLLBAR_SERVER_POST_KEY,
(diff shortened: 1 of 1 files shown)


cli/dcoscli/analytics.py
```diff
@@ -261,10 +262,18 @@ def _base_properties(conf=None):
         cmd = 'dcos'
         full_cmd = 'dcos'
 
+    try:
+        dcos_hostname = six.moves.urllib.parse.urlparse(
+            conf.get('core.dcos_url')).hostname
+    except:
```
mgummelt: except DCOSException

cli/dcoscli/analytics.py
```diff
@@ -261,10 +262,18 @@ def _base_properties(conf=None):
...
+        dcos_hostname = None
+
     return {
         'cmd': cmd,
         'full_cmd': full_cmd,
         'dcoscli.version': dcoscli.version,
         'python_version': str(sys.version_info),
-        'config': json.dumps(list(conf.property_items()))
+        'config': json.dumps(list(conf.property_items())),
+        'DCOS_HOSTNAME': dcos_hostname,
```
mgummelt: plz stay consistent with lower case

mgummelt:
I think this will make test_analytics fail

cli/dcoscli/analytics.py
jsancio: I didn't pick the key name. The analytic team want it this way so they can do a `group by`.

cli/dcoscli/analytics.py
jsancio: Why DCOSException? `urlparse` can fail.

mgummelt: nvm

jsancio:
@mgummelt Are we happy? I want get this in before Friday.

mgummelt:
LGTM

jsancio:
Great! Thanks.

PR #332, opened 2015-08-13
dcos-297 add HTTP 1.1 basic auth

tamarrow:
This will prompt the use once per command for username/password if the server responds that it needs HTTPBasicAuth. It also allows the username/password (or just the username) to be stored in `core.dcos_url` instead of typing credentials in (`http://username:password@dcos_url`).


cli/dcoscli/data/config-schema/core.json
```diff
@@ -9,7 +9,7 @@
     },
     "dcos_url": {
       "type": "string",
-      "pattern": "^(?:(?:https?)://)(?:(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\\.?)+(?:[a-zA-Z]{2,6}\\.?|[a-zA-Z0-9-]{2,}\\.?)?|\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3})(?::\\d+)?(?:/?|[/?]\\S+)$",
+      "pattern": "^(?:(?:https?)://)(([a-zA-Z0-9_-])+(:[a-zA-Z0-9_-]+)?@){0,1}(?:(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\\.?)+(?:[a-zA-Z]{2,6}\\.?|[a-zA-Z0-9-]{2,}\\.?)?|\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3})(?::\\d+)?(?:/?|[/?]\\S+)$",
```
mgummelt: do you mean to restrict usernames and passwords to such a restricted character set?

dcos/http.py
```diff
@@ -79,6 +83,42 @@ def request(method,
...
 
+    return response
+
+
+def request(method,
+            url,
+            is_success=_default_is_success,
+            timeout=None,
+            **kwargs):
+    """Sends an HTTP request.
```
mgummelt: If this is different than `_request`, please update the docstring to clarify

cli/dcoscli/data/config-schema/core.json
tamarrow: yes for base64 encoding

dcos/http.py
```diff
@@ -79,6 +83,42 @@ def request(method,
...
+    """
+
+    if 'headers' not in kwargs:
+        kwargs['headers'] = {'Accept': 'application/json'}
+
+    timeout = timeout or _timeout()
+
+    response = _request(method, url, is_success, timeout, **kwargs)
+
+    if response.status_code == 401:
```
mgummelt: This should probably loop until the user provides correct auth

dcos/http.py
```diff
@@ -1,11 +1,18 @@
+import getpass
+import sys
+from urlparse import urlparse
```
mgummelt: I don't think this is py2/3 compatible.  You have to use `six.moves.urllib.parse`.  see `mesos.py`

dcos/http.py
```diff
@@ -180,3 +220,29 @@ def delete(url, **kwargs):
...
+
+    :param response: requests.response
+    :type response: requests.Response
+    :returns: process return code
+    :rtype: int
+    """
+
+    parsed_url = urlparse(response.url)
+    hostname = parsed_url.hostname
+    if response.headers.get('www-authenticate') == 'Basic realm="Restricted"':
```
mgummelt: Why do we only support the "Restricted"  realm?

dcos/http.py
```diff
@@ -1,11 +1,18 @@
...
 import requests
 from dcos import util
 from dcos.errors import DCOSException, DCOSHTTPException
+from requests.auth import HTTPBasicAuth
 
 logger = util.get_logger(__name__)
 
 DEFAULT_TIMEOUT = 5
 
+AUTH_HOSTS = {}  # hostname -> AuthBase()
```
mgummelt: this should be keyed on (hostname, realm) pair, no?

cli/tests/integrations/test_config.py
```diff
@@ -410,6 +416,9 @@ def test_prepend_url_validation(env):
 
 def test_fail_url_validation(env):
     _fail_url_validation('set', 'core.dcos_url', 'http://bad_domain/', env)
+    _fail_url_validation('set', 'core.dcos_url', 'http://@domain/', env)
+    _fail_url_validation('set', 'core.dcos_url', 'http://a.b:b.a@domain/', env)
+    _fail_url_validation('set', 'core.dcos_url', 'http://user:pass@/', env)
```
mgummelt: we should test the actual user authentication exchange.  do we have a DCOS image with basic auth enabled?  If not, let's mock it.

dcos/http.py
```diff
@@ -180,3 +220,29 @@ def delete(url, **kwargs):
...
+    if response.headers.get('www-authenticate') == 'Basic realm="Restricted"':
+        username = parsed_url.username
+        if username is None:
+            sys.stdout.write('Username: ')
+            sys.stdout.flush()
+            username = sys.stdin.readline().strip().lower()
+        password = getpass.getpass()
+        AUTH_HOSTS[hostname] = HTTPBasicAuth(username, password)
+    else:
+        raise DCOSException("We don't support this authentication. {}".format(
```
mgummelt: `Server responded with an HTTP "www-authenticate" field of {}.  We only support {}."

and

`Invalid HTTP response: server returned an HTTP 401 response with no "www-authenticate" field`

dcos/http.py
```diff
@@ -180,3 +220,29 @@ def delete(url, **kwargs):
 def silence_requests_warnings():
     """Silence warnings from requests.packages.urllib3.  See DCOS-1007."""
     requests.packages.urllib3.disable_warnings()
+
+
+def _get_http_auth_credentials(response):
```
mgummelt: This is really a `set`, not a `get`

dcos/http.py
```diff
@@ -180,3 +220,29 @@ def delete(url, **kwargs):
...
+    """Get username/password for HTTP Basic Auth
+
+    :param response: requests.response
+    :type response: requests.Response
+    :returns: process return code
```
mgummelt: incorrect

cli/dcoscli/data/config-schema/core.json
mgummelt: I'm pretty sure username/password in the URL is plaintext, and `requests` does the base64 encoding.  have you tested this against a real basic auth server?

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: Maybe 3 times? Infinite times seems unnecessary.

cli/tests/integrations/test_config.py
jsancio: I think we are going to have a DCOS Image with it soon. @spahl Are we going to be able to enable basic authentication when configuring the vagrant image?

dcos/http.py
mgummelt: They can Ctrl-C whenever they want, but sure

dcos/http.py
tamarrow: changed

dcos/http.py
tamarrow: fixed

dcos/http.py
jsancio: Does requests parse the content of `www-authenticate` for us? I ask this because it is possible for a server to return multiple authentication methods.

mgummelt:
also note that we'd be sending passwords over non-https.  product is aware of this?

dcos/http.py
tamarrow: done

dcos/http.py
jsancio: That's correct. From RFC1945:

> The realm attribute (case-insensitive) is required for all authentication schemes which issue a challenge. The realm value (case-sensitive), in combination with the canonical root URL of the server being accessed, defines the protection space. These realms allow the protected resources on a server to be partitioned into a set of protection spaces, each with its own authentication scheme and/or authorization database. The realm value is a string, generally assigned by the origin server, which may have additional semantics specific to the authentication scheme.

Not sure what they mean by canonical root URL.

dcos/http.py
tamarrow: yep. it puts it in a dict: http://www.python-requests.org/en/latest/user/quickstart/#response-headers

dcos/http.py
mgummelt: the effect of the function is that it sets the value in the dictionary.

If you'd like to keep it `get`, return the credentials, and set them in the caller, then that would be fine (and probably better to keep things as functional as possible).

dcos/http.py
```diff
@@ -180,3 +220,29 @@ def delete(url, **kwargs):
...
+            sys.stdout.write('Username: ')
+            sys.stdout.flush()
+            username = sys.stdin.readline().strip().lower()
+        password = getpass.getpass()
+        AUTH_HOSTS[hostname] = HTTPBasicAuth(username, password)
+    else:
+        raise DCOSException("We don't support this authentication. {}".format(
+            response.content))
+
+    return 0
```
jsancio: I think this should return a `BasicAuth` this is going to be useful when we replace this with a long running process that securely stores and returns user credentials. Thoughts?

dcos/http.py
mgummelt: I think he means there could be multiple instances of the www-authenticate header

cli/dcoscli/data/config-schema/core.json
jsancio: Yeah. It looks like encoding the username and password in the URL only supports username and passwords with valid URL characters.

jsancio: Having said that I would be very happy if we could get rid of this patter and leverage jsonschema `url` pattern or Python's url parser.

tamarrow: yea. Sadly jsonschema doesn't actually implement the url type.

dcos/http.py
tamarrow: `It is also special in that the server could have sent the same header multiple times with different values, but requests combines them so they can be represented in the dictionary within a single mapping, as per RFC 7230`

mgummelt: Right, but this code doesn't handle that correctly, right?

tamarrow: im fixing

tamarrow:
@jsancio updated

cli/dcoscli/data/config-schema/core.json
mgummelt: did you find out if other characters are supported in the username/password?

mgummelt:
still missing tests

dcos/http.py
```diff
@@ -79,6 +87,47 @@ def request(method,
...
 
+    return response
+
+
+def request(method,
+            url,
+            is_success=_default_is_success,
+            timeout=None,
+            **kwargs):
+    """Sends an HTTP request. If the server responds with 'authentication
```
mgummelt: If the server responds with a 401

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+    return HTTPBasicAuth(username, password)
+
+
+def _get_realm(response):
+    """Return authentication realm requested by server for supported type or None
+
+    :param response: requests.response
+    :type response: requests.Response
+    :returns: realm
+    :rtype: str
```
mgummelt: str | None

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+        sys.stdout.write('Username: ')
+        sys.stdout.flush()
+        username = sys.stdin.readline().strip().lower()
+    password = getpass.getpass()
+
+    return HTTPBasicAuth(username, password)
+
+
+def _get_realm(response):
+    """Return authentication realm requested by server for supported type or None
```
mgummelt: supported type?  do you mean 'basic'?  if so, just say so.

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+    :rtype: str
+    """
+
+    if response.headers.get('www-authenticate'):
+        auths = response.headers['www-authenticate'].split(',')
+        basic_realm = next(auth_type for auth_type in auths
+                           if auth_type.rstrip().lower().startswith("basic"))
+        if basic_realm:
+            realm = basic_realm.split('=')[-1].rstrip().lower()
+            return realm
```
mgummelt: I'd prefer if we added `else` to these `if`s to remove side effects

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+
+    return None
+
+
+def _set_http_auth_credentials(response):
+    """Set authentication credentials required by server
```
mgummelt: set them where?

dcos/http.py
mgummelt: Since client code might want to make an http request without prompting for username/password, we should probably make that configurable with a kwarg

dcos/http.py
```diff
@@ -25,11 +33,12 @@ def _default_is_success(status_code):
...
-def request(method,
-            url,
-            is_success=_default_is_success,
-            timeout=None,
-            **kwargs):
+def _request(method,
+             url,
+             is_success=_default_is_success,
+             timeout=None,
+             realm=None,
```
mgummelt: `realm` is a strange parameter to send.  Can we just make this `auth`, and do the lookup in the caller?

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+    hostname = parsed_url.hostname
+    user = parsed_url.username
+
+    if response.headers.get('www-authenticate'):
+        realm = _get_realm(response)
+        if realm:
+            AUTH_CREDS[(hostname, realm)] = _get_basic_auth_credentials(user)
+        else:
+            msg = ("Server responded with an HTTP 'www-authenticate' field of "
+                   "{}, DCOS only supports 'Basic'".format(
```
mgummelt: put the templated variable in quotes as well, for consistency

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+    :param username: username user for authentication
+    :type username: str
+    :returns: HTTPBasicAuth
+    :rtype: requests.auth.HTTPBasicAuth
+    """
+
+    if username is None:
+        sys.stdout.write('Username: ')
+        sys.stdout.flush()
+        username = sys.stdin.readline().strip().lower()
```
mgummelt: If the username is set, we should print it, so they know what username they're providing a password for

dcos/http.py
```diff
@@ -180,3 +229,70 @@ def delete(url, **kwargs):
...
+
+    return None
+
+
+def _set_http_auth_credentials(response):
+    """Set authentication credentials required by server
+
+    :param response: requests.response
+    :type response: requests.Response
+    :returns: status code
```
mgummelt: what's the purpose of this status code? why not just return None?

dcos/http.py
```diff
@@ -79,6 +87,47 @@ def request(method,
...
+
+    response = _request(method, url, is_success, timeout, **kwargs)
+
+    i = 0
+    while i < 3 and response.status_code == 401:
+        _set_http_auth_credentials(response)
+        # try request again, with auth
+        response = _request(method, url, is_success, timeout,
+                            _get_realm(response), **kwargs)
+        i += 1
```
mgummelt: If i == 3, you should probably raise a DCOSException('Authentication failed')

tamarrow:
Holding off on tests to see if we can get a dcos-image with auth

dcos/http.py
tamarrow: ok

dcos/http.py
tamarrow: oka

dcos/http.py
tamarrow: ok

dcos/http.py
tamarrow: ok

dcos/http.py
tamarrow: good point

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: done

cli/dcoscli/data/config-schema/core.json
tamarrow: yes, changed the regex

tamarrow:
updated, added a few mocks

dcos/http.py
```diff
@@ -1,11 +1,19 @@
...
 from dcos.errors import DCOSException, DCOSHTTPException
+from requests.auth import HTTPBasicAuth
+
+from six.moves.urllib.parse import urlparse
 
 logger = util.get_logger(__name__)
 
 DEFAULT_TIMEOUT = 5
 
+AUTH_CREDS = {}  # (hostname, realm) -> AuthBase()
```
jsancio: Can we verify that we don't have a race condition? I remember that we perform HTTP requests over multiple threads: segment.io and dcos. If this is the case, how do we plan to address those data races? cc @mgummelt.

dcos/http.py
```diff
@@ -180,3 +230,73 @@ def delete(url, **kwargs):
...
+    :param response: requests.response
+    :type response: requests.Response
+    :rtype: None
+    """
+
+    parsed_url = urlparse(response.url)
+    hostname = parsed_url.hostname
+    user = parsed_url.username
+
+    if response.headers.get('www-authenticate'):
```
mgummelt: if 'www-authenticate' in response.headers

dcos/http.py
```diff
@@ -79,8 +86,51 @@ def request(method,
...
+    """
+
+    if 'headers' not in kwargs:
+        kwargs['headers'] = {'Accept': 'application/json'}
+
+    timeout = timeout or _timeout()
+
+    response = _request(method, url, is_success, timeout, **kwargs)
+
+    i = 0
```
mgummelt: might be nice to factor this out into a separate function

dcos/http.py
```diff
@@ -180,3 +230,73 @@ def delete(url, **kwargs):
...
+def _get_realm(response):
+    """Return authentication realm requested by server for 'Basic' type or None
+
+    :param response: requests.response
+    :type response: requests.Response
+    :returns: realm
+    :rtype: str | None
+    """
+
+    if response.headers.get('www-authenticate'):
```
mgummelt: `if 'www-authenticate' in response.headers`

dcos/http.py
jsancio: Should we also print the URL (hostname and realm)? So that the user knows where we are sending those credentials?

dcos/http.py
mgummelt: I think we're fine.  Python dicts are thread-safe for single operations.

dcos/http.py
mgummelt: Ah yea, definitely the hostname.  Not the realm.  It'll likely confuse users.

dcos/http.py
```diff
@@ -79,8 +86,51 @@ def request(method,
...
+    if 'headers' not in kwargs:
+        kwargs['headers'] = {'Accept': 'application/json'}
+
+    timeout = timeout or _timeout()
+
+    response = _request(method, url, is_success, timeout, **kwargs)
+
+    i = 0
+    while i < 3 and response.status_code == 401:
+        _set_http_auth_credentials(response)
```
jsancio: Doesn't this always ask the user's username and password? I thought we wanted to cache their credential in case we do multiple request to the same (hostname, realm). Where is the code that implements that?

For example, I know that we have commands that fist call `marathon/v2/info` and then `marathon/v2/...`. Have we tested this case? Do we need to implement BasicAuth in DCOS before we merge this PR?

dcos/http.py
```diff
@@ -79,8 +86,51 @@ def request(method,
...
+        hostname = urlparse(response.url).hostname
+        auth = AUTH_CREDS.get((hostname, _get_realm(response)))
+        # try request again, with auth
+        response = _request(method, url, is_success, timeout, auth, **kwargs)
+        i += 1
+
     if is_success(response.status_code):
         return response
+    elif i == 3:
+        return DCOSException("Authentication failed")
```
mgummelt: return? raise.

dcos/http.py
jsancio: But we don't. At a high-level the implementation should be doing the following:
```
if (hostname, realm) in AUTH_CREDS:
  # reuse the cache credentials
  ...
else:
  # ask and store credentials
  ...
  AUTH_CREDS[(hostname, realm)] = ...
```

dcos/http.py
tamarrow: fixed

dcos/http.py
tamarrow: fixed

dcos/http.py
tamarrow: fixed. If we have more than 1 commands, we cache in the global dict, so we don't ask for creds several times. Verified with some marathon adds/updates. I'll add some tests for this. 

We don't need BasicAuth to merge this PR, it will just be dead code. As was going to wait for it though, to make sure me and seb's implementations work together.

dcos/http.py
tamarrow: fixed

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: updated

dcos/http.py
tamarrow: changed a few things around so it should be safe according to this: http://effbot.org/pyfaq/what-kinds-of-global-value-mutation-are-thread-safe.htm

tamarrow:
@jsancio updated

dcos/http.py
```diff
@@ -1,11 +1,27 @@
...
 logger = util.get_logger(__name__)
+lock = threading.RLock()
 
 DEFAULT_TIMEOUT = 5
 
+# only accessed from _request_with_auth
+AUTH_CREDS = {}  # (hostname, realm) -> AuthBase()
+
+
+def _timeout():
```
jsancio: Please rebase. We removed this in master.

dcos/http.py
```diff
@@ -72,6 +89,92 @@ def request(method,
...
+    i = 0
+    while i < 3 and response.status_code == 401:
+        hostname = urlparse(response.url).hostname
+        creds = (hostname, _get_realm(response))
+
+        with lock:
+            if creds not in AUTH_CREDS:
+                auth = _get_http_auth_credentials(response)
+            else:
+                auth = AUTH_CREDS[creds]
```
jsancio: Perfect. +1.

dcos/http.py
```diff
@@ -72,6 +89,92 @@ def request(method,
...
+
+        # try request again, with auth
+        response = _request(method, url, is_success, timeout, auth, **kwargs)
+
+        # only store credentials if they're valid
+        with lock:
+            if creds not in AUTH_CREDS and response.status_code == 200:
+                AUTH_CREDS[creds] = auth
```
jsancio: Good point!

dcos/http.py
```diff
@@ -1,11 +1,27 @@
...
+
 import requests
 from dcos import util
 from dcos.errors import DCOSException, DCOSHTTPException
+from requests.auth import HTTPBasicAuth
+
+from six.moves.urllib.parse import urlparse
 
 logger = util.get_logger(__name__)
+lock = threading.RLock()
```
jsancio: Okay but it doesn't look like you need a reentrant lock, no?

jsancio:
LGTM. Just some minor comments.

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: true, changed to `threading.Lock()`

tamarrow:
thanks @jsancio, updated with latest comments

PR #341, opened 2015-08-21
Amend license information

ssk2:


mgummelt:
Make sure the commit is complete enough for end users.  "Amend license information" doesn't say anything about the effect of the change.

ssk2:
Got it! In this case nothing changed,  we just added the full license
instead of a link to it.

Make sure the commit is complete enough for end users. "Amend license
information" doesn't say anything about the effect of the change.

—
Reply to this email directly or view it on GitHub
<https://github.com/mesosphere/dcos-cli/pull/341#issuecomment-134257896>.

PR #342, opened 2015-08-21
[FIX] DCOS-2884 Preserve both stdout and stderr calling "--help" on the subcommand

YuraTolochkevich:


cli/dcoscli/help/main.py
```diff
@@ -112,9 +114,9 @@ def _help_command(command):
...
-        emitter.publish(stderr.decode("utf-8"))
-        return 1
-    emitter.publish(stdout.decode("utf-8"))
-    return 0
+    executable = subcommand.command_executables(command)
+
+    subproc = Popen([executable,  command] + ['--help'],
+                    stderr=PIPE)
+
+    return analytics.wait_and_track(subproc)
```
jsancio: We don't need to call this through the analytics code path. The `dcos` program will do this for us. We should be able to just:
```
 return subprocess.call([executable, command, '--help'])
```

That would inherit `dcos help`'s stdout, stdin, stderr which is what we want.

jsancio:
Thanks! I left you one comment.

jsancio:
@YuraTolochkevich I reviewed this PR and left you one comment. Thanks.

YuraTolochkevich:
@Thanks, fixed!

jsancio:
LGTM!

PR #344, opened 2015-08-25
[ADD] DCOS-2912 Remove DCOS_PRODUCTION environment variable and only report analytics if version is not SNAPSHOT

YuraTolochkevich:


jsancio:
LGTM!

PR #345, opened 2015-08-27
dcos-419 add support for SSL

tamarrow:
first pass


tamarrow:
fyi ran into this (since I'm running python 2.7.6): http://stackoverflow.com/questions/29099404/ssl-insecureplatform-error-when-using-requests-package. Will probably want to document it

jsancio:
@tamarrow Do you want me to review this? Tests are not passing.

tamarrow:
@jsancio thats cause ssl_verify is set to true, ill change it to false

jsancio:
Also we are missing tests for the positive and negative cases. Missing files, invalid ssl, etc.

tamarrow:
@jsancio added tests

jsancio:
@tamarrow ping

tamarrow:
@jsancio rebased

cli/dcoscli/main.py
```diff
@@ -79,6 +85,10 @@ def _main():
        not auth.check_if_user_authenticated():
         auth.force_auth()
 
+    #  TODO (tamar): support reading tuple:
+    #  (/path/to/cert, /path/to/private/key)
```
jsancio: What is this TODO about? Do we still need to implement this?

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,26 @@
+from dcos import http, util
+from dcos.errors import DCOSException
+
+import pytest
+from six.moves import urllib
+
+DCOS_URL = util.get_config()['core.dcos_url'].replace('http://', 'https://')
+DCOS_URL = urllib.parse.urljoin(DCOS_URL, 'marathon/')
```
jsancio: Not a fan of this coding style. Why not just?
```
DCOS_URL = urllib.parse.urljoin(
    util.get_config()['core.dcos_url'].replace('http://', 'https://'),
    'marathon/')
```

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_dont_verify_ssl():
+    resp = http.request("get", DCOS_URL, verify=False)
```
jsancio: Why are these integration tests calling the `http` module? Why aren't we using the CLI to test this stuff? I am interested in the UX. What do we print on success and on failure?

dcos/package.py
```diff
@@ -754,7 +754,7 @@ def update_sources(config, validate=False):
                         source.url,
                         stage_dir)
 
-                    errors.append(e.message)
+                    errors.append(str(e))
```
jsancio: Why did we change this?

dcos/util.py
```diff
@@ -237,6 +237,22 @@ def configure_process_from_environ():
     configure_debug(os.environ.get(constants.DCOS_DEBUG_ENV))
 
 
+def set_ssl_info_env_vars():
```
jsancio: Let's move this to `dcoscli.main` no other component should be doing this.

dcos/util.py
```diff
@@ -237,6 +237,22 @@ def configure_process_from_environ():
...
+    """Set SSL info from config to environment variables
+
+    :rtype: None
+    """
+
+    config = get_config()
```
jsancio: When we move this to `dcosli.main` let's also pass the config in.

cli/dcoscli/main.py
tamarrow: Yes, right now we only allow specifying `.pem`.

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,26 @@
...
+    assert resp.status_code == 200
+
+
+def test_verify_ssl():
+    with pytest.raises(DCOSException) as e:
+        http.request("get", DCOS_URL, verify=True)
```
jsancio: Same. Let's test this using the dcos cli. Also, I would like to see the full output from this user story. Ideally in the test. If not possible then a comment in the PR.

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,26 @@
...
+    assert "certificate verify failed" in str(e)
+
+
+def test_verify_ssl_with_bad_cert():
+    with pytest.raises(DCOSException) as e:
+        cert = 'tests/data/ssl/fake.pem'
+        http._request("get", DCOS_URL, verify=True, cert=cert)
```
jsancio: Same. Let's test this using the dcos cli. Also, I would like to see the full output from this user story. Ideally in the test. If not possible then a comment in the PR.

cli/tests/integrations/test_ssl.py
tamarrow: updated

dcos/package.py
tamarrow: This has always been wrong, and I noticed it when testing.

cli/tests/integrations/test_dcos.py
```diff
@@ -62,6 +62,12 @@ def test_help():
 
     DCOS_DEBUG                  If set then enable further debug messages which
                                 are sent to stdout.
+
+    DCOS_SSL_VERIFY             If set specifies whether to verify SSL certs
+                                for HTTPS.
+
+    DCOS_CERT                   This environment variable points to the
+                                location of the SSL cert .pem file
```
jsancio: Let's document this better. Please state that the user can control the value of these environment variables by using `dcos config set ...`.

dcos/util.py
tamarrow: moved

jsancio:
Looks good. Looks like we are missing one test: SSL verification is enabled and we have valid server certificate. Is it possible to test this with the Vagrant DCOS?  cc @cmaloney.

dcos/util.py
tamarrow: ok

cli/tests/integrations/test_dcos.py
tamarrow: updated

cli/dcoscli/main.py
```diff
@@ -33,6 +33,15 @@
 
     DCOS_DEBUG                  If set then enable further debug messages which
                                 are sent to stdout.
+
+    DCOS_SSL_VERIFY             If set specifies whether to verify SSL certs
+                                for HTTPS. Can also be configured by setting
+                                `core.ssl_config` in the config.
```
jsancio: How about:

```
If set specifies whether to verify SSL certs for HTTPS. To enable permanently use `dcos config set core.ssl_config true`.
```

cli/dcoscli/main.py
```diff
@@ -33,6 +33,15 @@
...
+
+    DCOS_CERT                   This environment variable points to the
+                                location of the SSL cert .pem file. Can also be
+                                configured by setting `core.cert` in the
+                                config.
```
jsancio: How about:

```
This environment variable points to the location of the SSL cert .pem file. To enable permanently
use `dcos config set core.cert <path-to-cert>`.
```

cli/dcoscli/main.py
jsancio: Are we going to implement this before merging this PR? If not then let's remove the TODO and create an Issue.

cli/dcoscli/main.py
```diff
@@ -140,3 +154,19 @@ def wrapper(*args, **kwargs):
...
+    """Set SSL info from config to environment variables
+
+    :param config: config
+    :type config: Toml
+    :rtype: None
+    """
+
+    if 'core.ssl_verify' in config:
+        os.environ[constants.DCOS_SSL_VERIFY_ENV] = str(
+            config['core.ssl_verify'])
```
jsancio: This means that `dcos.toml` takes precedence over the `DCOS_SSL_VERIFY` environment variable. Is this what we want? As a user this would be surprising to me.

cli/dcoscli/main.py
```diff
@@ -140,3 +154,19 @@ def wrapper(*args, **kwargs):
...
+
+    if 'core.cert' in config:
+        os.environ[constants.DCOS_CERT_ENV] = config['core.cert']
```
jsancio: This means that dcos.toml takes precedence over the DCOS_CERT environment variable. Is this what we want? As a user this would be surprising to me

bin/start_tests.sh
```diff
@@ -15,6 +15,9 @@ cd cli
 cp tests/data/dcos.toml $DCOS_CONFIG
 sed -i "s/change.dcos.url/$VBOX_IP/g" $DCOS_CONFIG
 
+# overwrite core.dcos_url in the ssl.toml
+sed -i "s/change.dcos.url/$VBOX_IP/g" tests/data/ssl/ssl.toml
```
jsancio: This means that running the test would change the repository. This is the reason why we `cp` the file in line 15 above.

I think you won't need to do this if you allow environment variable to take precedence over `dcos.toml`. Thoughts?

dcos/package.py
jsancio: :+1:

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,91 @@
...
+
+    returncode, stdout, stderr = exec_command(
+        ['dcos', 'marathon', 'app', 'list'], env)
+    assert returncode == 0
+    assert stderr == b''
+
+    config_unset('core.ssl_verify', None, env)
+
+
+def test_verify_ssl_without_ca_env_var(env):
```
jsancio: Technically this should be `test_verify_ssl_withtou_cert_env_var`.

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,91 @@
...
+
+    returncode, stdout, stderr = exec_command(
+        ['dcos', 'marathon', 'app', 'list'], env)
+    assert returncode == 1
+    assert "certificate verify failed" in stderr.decode('utf-8')
+
+    env.pop(constants.DCOS_SSL_VERIFY_ENV)
+
+
+def test_verify_ssl_without_ca_config(env):
```
jsancio: Technically this should be `test_verify_ssl_withtou_cert_config`.

cli/tests/integrations/test_ssl.py
```diff
@@ -0,0 +1,91 @@
+import os
```
jsancio: Tests look good. What did we decide regarding the self signed cert from the admin router? We are missing a test for valid SSL and HTTPS url.

tamarrow:
@jsancio updated!

tamarrow:
@jsancio all tests pass

dcos/http.py
```diff
@@ -162,11 +174,23 @@ def request(method,
     if 'headers' not in kwargs:
         kwargs['headers'] = {'Accept': 'application/json'}
 
-    response = _request(method, url, is_success, timeout, **kwargs)
+    if verify is None and os.environ.get(constants.DCOS_SSL_VERIFY_ENV):
```
jsancio: I think you want: `if verify is None and constants.DCOS_SSL_VERIFY_ENV in os.environ:`

dcos/http.py
```diff
@@ -162,11 +174,23 @@ def request(method,
...
+        verify = os.environ.get(constants.DCOS_SSL_VERIFY_ENV)
```
jsancio: you already checked it about so you want: `os.environ[constants.DCOS_SSL_VERIFY_ENV]`

dcos/http.py
```diff
@@ -162,11 +174,23 @@ def request(method,
...
+        if verify.lower() == "true":
+            verify = True
+        elif verify.lower() == "false":
+            verify = False
+
+    # Silence 'Unverified HTTPS request' and 'SecurityWarning' for bad certs
+    if verify is not None:
+        silence_requests_warnings()
```
jsancio: I think we do this in a few places. Should we remove them now that we do them here?

cli/tests/integrations/test_ssl.py
jsancio: @tamarrow Can we test that overriding the CA works? I don't remember if we decided to punt on this.

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: done

dcos/http.py
tamarrow: I'd rather not, in case some new error surfaces that we are unaware of.

tamarrow:
@jsancio updated

jsancio:
LGTM. Great job @tamarrow !

PR #347, opened 2015-08-28
[DCOS-2993] Added `--master-proxy` option to `dcos node ssh`.  This enables `dcos

mgummelt:
node ssh` in AWS.


cli/dcoscli/node/main.py
```diff
@@ -200,11 +189,35 @@ def _ssh(master, slave, option, config_file, user):
...
+            master_public_ip,
+            host)
+    else:
+        cmd = "ssh -t {0}{1}@{2}".format(
+            ssh_options,
+            user,
+            host)
 
     emitter.publish(DefaultError("Running `{}`".format(cmd)))
+    if (not master_proxy) and master_public_ip:
```
tamarrow: can we move this to the else block above?

tamarrow:
one small nit and then LGTM when tests pass

cli/dcoscli/node/main.py
mgummelt: I think it makes more sense after the "Running" message

PR #351, opened 2015-09-10
dcos-3103 Add post uninstall notes reporting

jsancio:


cli/tox.ini
```diff
@@ -23,11 +23,11 @@ commands =
 
 [testenv:py27-integration]
 commands =
-  py.test -vv {env:CI_FLAGS:} tests/integrations{posargs}
+  py.test -vv -x {env:CI_FLAGS:} tests/integrations{posargs}
 
 [testenv:py34-integration]
 commands =
-  py.test -vv {env:CI_FLAGS:} tests/integrations{posargs}
+  py.test -vv -x {env:CI_FLAGS:} tests/integrations{posargs}
```
jsancio: @mgummelt and @tamarrow, I know that you both talked about this but I found it very helpful while testing this change locally. What do you we think?

tamarrow: I've always been for it :)

cli/tests/data/dcos.toml
```diff
@@ -1,7 +1,7 @@
 [core]
 reporting = false
-timeout = 5
 email = "test@mail.com"
+timeout = 5
```
tamarrow: why did we move this?

cli/tox.ini
mgummelt: this is temporary while I'm testing

dcos/package.py
```diff
@@ -260,6 +263,17 @@ def is_match(app):
             logger.info(
                 'Found the following frameworks: {}'.format(framework_ids))
 
+            # Emit post uninstall notes
+            emitter.publish(
+                DefaultError(
+                    'Uninstalled package [{}] version [{}]'.format(
+                        package_json.get('name'),
+                        package_json.get('version'))))
```
tamarrow: We should probably put the `name`/`version` in an `if` or else the message wouldn't make sense

tamarrow:
lgtm, just a few nits

cli/tox.ini
mgummelt: oh sorry.  I thought this was my PR.  It's helpful with local testing, but I'd rather see all test results from the async test run.

Up to you guys now though.

cli/tests/data/dcos.toml
jsancio: Our TOML library moves them around since it doesn't use an order dictionary.

dcos/package.py
jsancio: They have to be set or else it if a bug. I'll fix the code.

cli/tox.ini
jsancio: Okay, I'll change it to `-x` since we should be running tests locally before pushing PRs.

jsancio:
@tamarrow updated PR.

tamarrow:
@jsancio :ship:

PR #352, opened 2015-09-14
bugfix for creating file if doesn't exist

tamarrow:


tamarrow:
@mgummelt can you look at this asap please. Pull request #314 broke the install script b/c we no longer create the toml file explicitly, and we couldn't create it later b/c we open files with `w` instead of `w+`

dcos/config.py
```diff
@@ -16,6 +17,7 @@ def load_from_path(path, mutable=False):
     :rtype: Toml | MutableToml
     """
 
+    util.ensure_dir_exists(os.path.dirname(path))
```
mgummelt: This is fine if you want to get this out ASAP, but it'd be better to put the directory creation back in the env setup script.  We don't want to create directories if the user sets a custom DCOS_CONFIG to /non/existent/dir/path

PR #354, opened 2015-09-17
dcos-2955 Fix bin directory resolution

jsancio:


dcos/subcommand.py
```diff
@@ -323,6 +322,24 @@ def uninstall(package_name):
...
+    :returns: Absolute path to virutalenv program
+    :rtype: str
+    """
+
+    virtualenv_path = os.path.join(bin_directory, 'virtualenv')
+    if not os.path.exists(virtualenv_path):
+        virtualenv_path = util.which('virtualenv')
+
+    if virtualenv_path is None:
+        raise DCOSException('Unable to find the virtualenv program')
```
tamarrow: Since now we don't require virtualenv. Should this be a more general message that we can't find the bin dir?

tamarrow:
LGTM, one small comment. Should we update the README as well?

dcos/subcommand.py
jsancio: This is finding virtualenv. We need virtualenv for installing the subcommands. E.g. `dcos spark`. This change just makes it so that we don't need virtualenv for install the core DCOS CLI.

jsancio:
> Should we update the README as well?

The README doesn't say anything about installing the CLI. We need to change docs.mesosphere.com.

jsancio:
I will merge this. Thanks!

PR #357, opened 2015-09-21
Docker container

timfallmk:
Added this as a submodule of https://github.com/mesosphere/dcos-cli-docker


jsancio:
@timfallmk can we implement this without using submodule? Also, please update the README in the DCOS CLI repo explaining how to use the docker container. Thanks!

timfallmk:
@jsancio updated.

mgummelt:
@jsancio: plz review

mgummelt:
@timfallmk: look over my changes.

The install is part of the build now, and we'll build an image for each version.  Since Dockerfiles aren't parametrizable, I'm doing this with a hacky sed command.

I also removed the debugging image and the LICENSE.  What would go wrong without the LICENSE?

bin/docker_push.sh
```diff
@@ -0,0 +1,21 @@
...
+set -e -x
+
+sed -i -e "s/\$docker_cli_version/${DCOS_CLI_VERSION}/g" docker/Dockerfile
+
+pushd docker
+
+docker build -t mesosphere/dcos-cli:${DCOS_CLI_VERSION} .
+docker tag -f mesosphere/dcos-cli:${DCOS_CLI_VERSION} mesosphere/dcos-cli:latest
+docker push mesosphere/dcos-cli:${DCOS_CLI_VERSION}
+docker push mesosphere/dcos-cli:latest
```
jsancio: cc @gabrielhartmann . I know that previously have been pushing to this name.

README.rst
```diff
@@ -15,6 +15,14 @@ command and for specific subcommands through :code:`dcos <subcommand> --help`.
 Additional documentation for the CLI and for the DCOS in general is available
 in the `Mesosphere docs`_.
 
+To run the DCOS CLI as a container, you can pull it from the docker hub with::
+
+    docker run -i -t mesosphere/dcos-cli <url of your cluster>
```
jsancio: I thought that we agreed that we wanted to run this as: `docker run -i -t -v ... mesosphere/dcos-cli dcos ...`.

jsancio: Also how is `dcos package install` going to work when a `command.json` is specified? It looks like the file system will get lost once the container is stopped.

mgummelt: see my comment in slack.

Doing so would preclude us from running like this (providing a url and getting a bash prompt), which I think is a much more common use case (by far, for me at least).

You at least have to configure your container with the URL somehow.

bin/docker_push.sh
mgummelt: Was he pushing?  I thought dcos-tests was just building locally under this name

README.rst
mgummelt: cc @pyronicide 

We're trying to decide between two uses of the dcos-cli docker:

1) `docker run -it -v ~/.dcos:/root/.dcos mesosphere/dcos-cli dcos package list`
2)

```
docker run -it mesosphere/dcos-cli
# dcos package list
```

1)
+ you can script and pipe commands from your host
- stupidly long commands.  we'd have to also ship some wrapper (e.g. `dcos-docker`)

2)
+  commands are still the same.  self-contained environment
- config and subcommand changes don't persist
-

bin/docker_push.sh
timfallmk: @mgummelt I think we can get away with leaving out the debugger build. It should keep auto-building in my personal docker hub namespace.

PR #358, opened 2015-09-24
Add exception catching for /shutdown.

ssk2:
The Mesos API removed /shutdown as of 0.24. This fix will try /teardown first and then /shutdown if /teardown returns a 404.


jsancio:
LGTM.

PR #359, opened 2015-09-24
New option for installing, but not running a package's app: dcos pack…

mgummelt:
…age install --stopped


cli/dcoscli/data/help/package.txt
```diff
@@ -50,6 +53,8 @@ Options:
                                            config.json and --options.  If not provided,
                                            print the raw templates.
 
+    --stopped                              Run 0 instances of the marathon app.
```
tamarrow: The wording is a little weird. We're not stopping, just not automatically starting. maybe `--install-only`?

PR #362, opened 2015-10-12
Escape newlines when rendering template

discordianfish:
If we read an options json file which includes properly encoded newlines
like this:

    { "foo": "bar\nbaz" }

And have a template like:

    { "bzz": "{{ foo }}" }

Without this change, the resulting file would be:

    { "bzz": "bar
    baz" }

which isn't valid json.

Alternatively people would need to double escape the newline like this:

    { "foo": "bar\\nbaz" }

which is quite confusing and unexpected from a users perspective.

One thing to keep in mind is that with this change, it's not possible to
use newlines outside strings. Since that should be very uncommon and
unreadable, I don't think we need/should support that.

The proper solution to all this would be to stop using template to
render json. The simplest alternative would be to simply override
elements by merging marathon.json with some user defined json. If that
doesn't work for all use cases, we would need to come up with a json
transformator (think XSLT) or DSL.

### DIFF
--- a/dcos/util.py
+++ b/dcos/util.py
@@ -487,7 +487,7 @@ def render_mustache_json(template, data):
     """
 
     try:
-        r = CustomJsonRenderer()
+        r = CustomJsonRenderer(escape=lambda s: s.replace("\n", "\\n"))
         rendered = r.render(template, data)
     except Exception as e:
         logger.exception(
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -11,7 +11,7 @@ def test_render_mustache_json():
     template = '{ "x": {{{xs}}}, "y": {{{ys}}}, "z": "{{{z}}}"}'
     xs = [1, 2, 3]
     ys = {'y1': 1, 'y2': 2}
-    z = 'abc'
+    z = 'abc\ndef'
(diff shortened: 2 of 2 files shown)


jsancio:
Yes @discordianfish. I like your suggestion of merging the marathon JSON with the user supplied JSON. Unfortunately, that doesn't work for two reasons:

- The original intend was to support more scheduler than just Marathon. For example Aurora.
- We need to figure out a way to dynamically render the `cmd` field. I mention that specifically because it tends to be the more complicated one.

I thought about allowing packages to supply a `marathon.py` with a well known function that gets executed and is expected to generate a marathon.json. This could be extended to `aurora.py`, etc.

jsancio:
@discordianfish The existing tests pass. Can you please provide a test for this feature?

discordianfish:
@jsancio I've changed the existing renderer testcase to include a multiline string but as you said, the tests fail. I have no idea why, this is exactly the behavior my change fixed. I just double checked to make sure the fix actually works. Just pass any multiline string as option to the cli and with my change it work, without it doesn't.

Why all this doesn't apply for the test - no idea. I assume it's somehow related to pystache not calling the escape function for static strings in the code or something but since I'm not a python expert, maybe you have some ideas?

PR #366, opened 2015-10-19
Stop handling status 409 specially

discordianfish:
409 can include a meaningful error message like other errors as well.
For instance a duplicated id yields 409 for which the current error is
very misleading.

In general the exception handling could use some refactoring. I don't really get why it's handling 400 special either. Some code comments would be useful here.

### DIFF
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -672,8 +672,9 @@ def test_app_locked_error():
         assert_command(
             ['dcos', 'marathon', 'app', 'stop', 'sleep-two-instances'],
             returncode=1,
-            stderr=(b'App or group is locked by one or more deployments. '
-                    b'Override with --force.\n'))
+            stderr=(b"Error: App is locked by one or more deployments. "
+                    b"Override with the option '?force=true'. View details"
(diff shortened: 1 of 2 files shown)


jsancio:
@discordianfish, looks like this breaks existing tests. Looks like catching `409` makes sense in some cases. We added that error handling because Marathon's error message are very HTTP specific. For example adding `?force` doesn't make sense in the command line.

discordianfish:
@jsancio It's definitely less confusing than getting the error about deployments if in fact a task with the same id already exists for example. I would prefer merging this as a incremental improvement.
As I said, this error handling in general could use some refactoring. If we want to 'translate' every error the api returns into something the user can do with the CLI (which makes sense), then we need some structure around that. For example use different error codes for every error case or return json as error (maybe via content negotiation to not break existing clients) which again includes some id / error code which can be mapped to specific errors.

For now, I'll change the test.

jsancio:
Hey @air, can we get this prioritized in Marathon? Let me know if you need to me to outline specific functional requirements.

air:
I actually don't follow what the ask is here for Marathon, can you spell it out for me? :sweat_smile:

tamarrow:
@jsancio bump

PR #372, opened 2015-11-09
Change the resource type for disk and cpu

jsancio:
The resource type for both disk and cpu should be floats/number not
integers.


tamarrow:
:+1: will merge once tests pass

PR #374, opened 2015-11-12
add support for app definitions with multiple apps in them

lloesche:
Marathon's /v2/apps API endpoint supports multiple apps. You can either PUT a single app `{}` or multiple apps `[{},{},{}]` in a single request. This PR adds support for the `dcos marathon` subcommand to handle multi app JSON files.

### DIFF
--- a/cli/dcoscli/marathon/main.py
+++ b/cli/dcoscli/marathon/main.py
@@ -364,22 +364,33 @@ def _add(app_resource):
     :returns: process return code
     :rtype: int
     """
-    application_resource = _get_resource(app_resource)
+    application_resources = _get_resource(app_resource)
 
     # Add application to marathon
     client = marathon.create_client()
 
     # Check that the application doesn't exist
-    app_id = client.normalize_app_id(application_resource['id'])
+    if not isinstance(application_resources, list):
+        application_resources = [application_resources]
 
-    try:
-        client.get_app(app_id)
-    except DCOSException as e:
-        logger.exception(e)
-    else:
-        raise DCOSException("Application '{}' already exists".format(app_id))
+    error_msg = ''
+    for application_resource in application_resources:
+        app_id = client.normalize_app_id(application_resource['id'])
 
-    client.add_app(application_resource)
+        try:
+            client.get_app(app_id)
+        except DCOSException as e:
+            logger.exception(e)
+        else:
+            if error_msg is not '':
+                error_msg += "\n"
+            error_msg += "Application '{}' already exists".format(app_id)
+
+    # Abort entire deployment if any app(s) already exist
+    if error_msg is not '':
+        raise DCOSException(error_msg)
+
+    client.add_app(application_resources)
 
     return 0
 
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -245,6 +245,18 @@ def remove_app(app_id):
     assert_command(['dcos', 'marathon', 'app', 'remove', app_id])
 
 
+def remove_apps(app_ids):
+    """ Remove multiple apps
+
+    :param app_ids: list of ids of apps to remove
+    :type app_ids: list
+    :rtype: None
+    """
+
+    for app_id in app_ids:
+        remove_app(app_id)
+
+
 def package_install(package, deploy=False, args=[]):
     """ Calls `dcos package install`
 
@@ -467,7 +479,10 @@ def app(path, app_id, wait=True):
     try:
         yield
     finally:
-        remove_app(app_id)
+        if isinstance(app_id, list):
+            remove_apps(app_id)
+        else:
+            remove_app(app_id)
         watch_all_deployments()
 
 
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -203,6 +203,30 @@ def test_remove_app():
     _list_apps()
 
 
+def test_add_multiapp():
+    with _zero_instance_multiapp():
+        _list_apps(['zero-instance-app1', 'zero-instance-app2'])
+
+
+def test_remove_multiapp():
+    with _zero_instance_multiapp():
+        pass
+    _list_apps()
+
+
+def test_add_existing_multiapp():
+    with _zero_instance_multiapp():
+        app_path = \
+            'tests/data/marathon/apps/zero_instance_sleep_v2_multiapp.json'
+        with open(app_path) as fd:
+            stderr = (b"Application '/zero-instance-app1' already exists\n"
+                      b"Application '/zero-instance-app2' already exists\n")
+            assert_command(['dcos', 'marathon', 'app', 'add'],
+                           returncode=1,
+                           stderr=stderr,
+                           stdin=fd)
+
+
 def test_add_bad_json_app():
     with open('tests/data/marathon/apps/bad.json') as fd:
         returncode, stdout, stderr = exec_command(
@@ -408,7 +432,7 @@ def test_app_add_invalid_request():
 
     assert returncode == 1
     assert stdout == b''
-    assert re.match(b"Error on request \[POST .*\]: HTTP 400: Bad Request:",
+    assert re.match(b"Error on request \[PUT .*\]: HTTP 400: Bad Request:",
                     stderr)
 
     stderr_end = b"""{
@@ -417,7 +441,7 @@ def test_app_add_invalid_request():
       "errors": [
         "host is not a valid network type"
       ],
-      "path": "/container/docker/network"
+      "path": "(0)/container/docker/network"
     }
   ],
   "message": "Invalid JSON"
@@ -784,8 +808,11 @@ def _list_apps(app_id=None):
     if app_id is None:
         assert len(result) == 0
     else:
-        assert len(result) == 1
-        assert result[0]['id'] == '/' + app_id
+        if not isinstance(app_id, list):
+            assert len(result) == 1
+            assert result[0]['id'] == '/' + app_id
+        else:
+            assert len(result) == len(app_id)
 
     assert returncode == 0
     assert stderr == b''
@@ -879,3 +906,10 @@ def _zero_instance_app_through_http():
             yield
         finally:
             server.shutdown()
+
+
+@contextlib.contextmanager
+def _zero_instance_multiapp():
+    with app('tests/data/marathon/apps/zero_instance_sleep_v2_multiapp.json',
+             ['zero-instance-app1', 'zero-instance-app2']):
+        yield
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -290,7 +290,10 @@ class Client(object):
         else:
             app_json = app_resource
 
-        response = _http_req(http.post, url,
+        if not isinstance(app_json, list):
+            app_json = [app_json]
+
+        response = _http_req(http.put, url,
                              json=app_json,
                              timeout=self._timeout)
 
--- a/cli/tests/data/marathon/apps/zero_instance_sleep_v2_multiapp.json
+++ b/cli/tests/data/marathon/apps/zero_instance_sleep_v2_multiapp.json
@@ -1,0 +1,24 @@
+[
+  {
+    "id": "zero-instance-app1",
+    "cmd": "sleep 2000",
+    "cpus": 0.1,
+    "mem": 16,
+    "instances": 0,
+    "labels": {
+      "PACKAGE_ID": "zero-instance-app",
+      "PACKAGE_VERSION": "1.2.3"
+    }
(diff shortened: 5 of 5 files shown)


jsancio:
@tamarrow Can you take a look at this PR?

cli/dcoscli/marathon/main.py
```diff
@@ -364,22 +364,27 @@ def _add(app_resource):
...
-        raise DCOSException("Application '{}' already exists".format(app_id))
+    for application_resource in application_resources:
+        app_id = client.normalize_app_id(application_resource['id'])
+
+        try:
+            client.get_app(app_id)
+        except DCOSException as e:
+            logger.exception(e)
+        else:
+            raise DCOSException("Application '{}' already exists".format(
```
tamarrow: This will end execution. Do we want to try all the apps and add all the ones that don't already exist?

cli/tests/integrations/common.py
```diff
@@ -245,6 +245,18 @@ def remove_app(app_id):
...
+def remove_apps(app_ids):
+    """ Remove multiple apps
+
+    :param app_ids: list of ids of apps to remove
+    :type app_ids: list
+    :rtype: None
+    """
+
+    for app_id in app_ids:
+        assert_command(['dcos', 'marathon', 'app', 'remove', app_id])
```
tamarrow: you can just call `remove_app(app_id)` here

cli/dcoscli/marathon/main.py
lloesche: If I have a multi app definition esp. as part of a package I would either want all or none of them to be deployed. Half finished deployments are nasty. Likewise if half of the app is already deployed but the other half is missing there's clearly something wrong. In which case I'd either want to remove the existing apps and do a clean deployment or update the group (as soon as we have group support).

dcos/marathon.py
```diff
@@ -290,7 +290,10 @@ def add_app(self, app_resource):
         else:
             app_json = app_resource
 
-        response = _http_req(http.post, url,
+        if not isinstance(app_json, list):
+            app_json = [app_json]
+
+        response = _http_req(http.put, url,
```
tamarrow: why did we change this to a `put`?

dcos/package.py
```diff
@@ -1373,10 +1376,18 @@ def marathon_json(self, revision, options):
...
+        if not isinstance(init_desc, list):
+            labels = init_desc.get('labels', {})
 
-        labels.update(package_labels)
-        init_desc['labels'] = labels
+            labels.update(package_labels)
+            init_desc['labels'] = labels
+        else:
+            for i in range(len(init_desc)):
+                app = init_desc[i]
```
tamarrow: you can use `enumerate` here: `for i, app in enumerate(init_desc)`

cli/tests/integrations/test_marathon.py
```diff
@@ -203,6 +203,17 @@ def test_remove_app():
     _list_apps()
 
 
+def test_add_multiapp():
+    with _zero_instance_multiapp():
+        _list_apps(['zero-instance-app1', 'zero-instance-app2'])
+
```
tamarrow: can we add tests for the negative cases also?

dcos/package.py
```diff
@@ -58,10 +58,13 @@ def install_app(pkg, revision, init_client, options, app_id):
...
-                     app_id,
-                     init_desc['id'])
-        init_desc['id'] = app_id
+        if not isinstance(init_desc, list):
+            logger.debug('Setting app ID to "%s" (was "%s")',
+                         app_id,
+                         init_desc['id'])
+            init_desc['id'] = app_id
+        else:
+            logger.debug('Not setting app ID for multi app json')
```
tamarrow: why are we not setting app_id here?

dcos/package.py
```diff
@@ -1373,10 +1376,18 @@ def marathon_json(self, revision, options):
...
-        init_desc['labels'] = labels
+            labels.update(package_labels)
+            init_desc['labels'] = labels
+        else:
+            for i in range(len(init_desc)):
+                app = init_desc[i]
+                labels = app.get('labels', {})
+                labels.update(package_labels)
+                app['labels'] = labels
+                init_desc[i] = app
```
tamarrow: can we add a test for this case?

cli/dcoscli/marathon/main.py
tamarrow: Makes sense. Also, we do have group support in the cli!

lloesche: I meant for when I use `dcos package install` and then it should deploy a group with multiple applications. Afaik this is not supported yet?! At least I couldn't find it in the code and the marathon.json is always send to the /v2/apps endpoint.

dcos/marathon.py
lloesche: I talked with @aquamatthias and the /v2/apps endpoint supports only single apps via the POST method but single as well as multiple apps via the PUT method. When sending a POST you get the complete app definition returned while when using PUT you get the deployment ID returned. I checked the code base and it seems we're not currently doing anything with the return value. When talking with the Marathon team they said that the deployment ID might be useful for implementing a blocking function that waits for the deployment to finish after sending a app definition. This would allow user stories where they can use dcos-cli in their scripts and perform steps like 1) start database server, 2) create database schema, which is currently harder to implement in simple bash.

dcos/package.py
lloesche: The app_id has to be unique in Marathon. When I have a marathon.json with multiple apps I can't set the app_id for all of them to the same value. I could set it for the first app found in the list, but that feels like odd behaviour. We could extend it so that app_id could also be a list and then override them in the order they were supplied, but again that seems like a very artificially constructed use case.

cli/tests/integrations/test_marathon.py
lloesche: Sorry, I suck at tests, can you explain what you mean by negative case? I went through the existing list of tests and thought that the others like e.g. bad json or adding multiple existing apps are already covered by the existing tests since they are using the same method and technically all apps (single and multi) are now multiple apps. Just that a single app ends up being a list with just one element in it. If that makes sense.

tamarrow: Some tests I was thinking of: a multiple apps test when 1 app already exists but the other doesn't, when both/all of the apps already exist, when none of the apps exist, but one has an error.

dcos/marathon.py
tamarrow: Makes sense. The app definition returned should contain the deployment idea, right? Could we parse it out so we can continue to return "Created deployment x"?

dcos/package.py
tamarrow: The other option is to append -1, -2, etc to the app-id name, but that is also arbitrary. I think its fine to leave as is.

dcos/marathon.py
lloesche: Yes, though I see that as a separate PR. Atm nothing is done with the return value. It's just passed through.

cli/tests/integrations/test_marathon.py
lloesche: Awesome, thanks! I implemented the one where both apps already exist. That is testing the new code that generates the multi line error message. The other cases are covered by the existing tests as they don't differ between single and multi app deployments. E.g. if one app exists but one has an error, from dcos-cli's perspective that's the same as a single app having an error as it's just sending the json object to Marathon without any regard for how many apps are in there. Like, there is no cli code that distinguishes at that point.

lloesche:
@tamarrow all changes should be implemented. I squashed the commits. I'm rerunning the tests because one of the old ones (test_log_two_tasks_follow) failed for what looked like a timing error. It succeeded for Python 2.7 but failed for Python 3. There seem to be several tests that are a bit flaky. This one didn't receive data after a 1s sleep. I also saw others that perform a task scaling from 0 to 10 instances followed by a `dcos marathon deployment list --json` which sometimes failed because the deployment was already finished by the time the command was executed.

cli/dcoscli/marathon/main.py
```diff
@@ -364,22 +364,33 @@ def _add(app_resource):
...
+    error_msg = ''
+    for application_resource in application_resources:
+        app_id = client.normalize_app_id(application_resource['id'])
+
+        try:
+            client.get_app(app_id)
+        except DCOSException as e:
+            logger.exception(e)
+        else:
+            if error_msg is not '':
```
jgehrcke: At this point, `error_msg` always is `''`. So this condition is never met. Am I missing something?

dcos/package.py
jsancio: Let's think about this some more. It is possible that we can just remove `--app-id` from the `dcos package install` command. I am not sure if anyone uses this feature. Let me think about it.

To me, it sounds good to add support for multiple app to the `dcos marathon app` commands. I am worry about using this in the `dcos package install` command. I need to understand what is driving this change. Let's meet on Monday so I can better understand the user story.

lloesche: There's a number of situations where this would be useful. Mainly it's me not wanting a single overloaded Docker container with a supervisor process inside. Like k8s is currently using.
Think openvpn. You need the openvpn service itself running on the public slaves as well as some administrative service running internally. You could make the user run something like `dcos package install openvpn-gateway; dcos package install openvpn-adminserver` but wouldn't it be nicer if they could just do `dcos package install openvpn`?
Or in my case, I have an edge router consisting of two services. One sets up VRRP on the public slave nodes, the other sets up a haproxy config generator (https://github.com/mesosphere/marathon-lb). I don't want the user to have to run `dcos package install haproxy-vrrp; dcos package install marathon-lb` because the one without the other has no function. Instead installing marathon-lb should start two applications, keepalived which is provided with the dcos-image directly and a marathon-lb docker container. I'd also like to keep the haproxy container unprivileged. Whereas keepalived requires host root privileges to manipulate the network interface IP configuration. If both run in the same privileged container that's a security risk since haproxy is externally accessible.

Removing --app-id should be a separate PR imho.

sttts:
This PR is related to DCOS-3535 in Jira about group support for dcos packages.

For Kubernetes we would really like to split our huge "Uber Container" which now hosts a handful of processes and its own S6 init daemon. If we could split that up into a handful of containers would be great, either via Marathon groups or multiple apps.

sttts:
For the specific use case: the Kubernetes Uber container has these processes:

- S6
  - nginx proxy
  - etcd
  - k8s-apiserver
  - k8s-controller-manager
  - k8s-mesos-scheduler

kolloch:
If I understand this correctly, this patch detects automatically whether the `marathon.json` contains an array or an object. If it contains an array, it will use a different request to push all contained apps to Marathon.

@lloesche: I really appreciate what you are trying to achieve but lists of apps is not the right approach. 

Let's please use groups with PUT. Why? With groups we make sure that all components (as the ones mentioned by @stts) are "grouped" so you can see in Marathon that they are related. If we update the package and we add/remove components, Marathon will calculate a diff and remove unneeded apps.

As for the detection whether a group or a single app must be deployed, I would suggest different file names instead of auto-detection based on the file content. E.g. if a `marathon_group.json` file is present, use that and assume the content is a group. If a `marathon.json` is present, assume that an app is contained.

That said, let's move forward with the idea because I really do like the idea that packages can install more than one component at once.

stts:
NAVER - http://www.naver.com/
--------------------------------------------

keliong@naver.com 님께 보내신 메일 <Re: [dcos-cli] add support for app definitions with multiple apps in them (#374)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

lloesche:
@kolloch not quite, if it detects an objects it encapsulates it in a list. That's all. It always PUTs to /v2/apps.

Like discussed in our meeting on Tuesday I'll remove the `dcos package` code and just leave the `dcos marathon add` parts.

For `dcos package` to deploy several apps groups support will be implemented in a separate PR.

cli/dcoscli/marathon/main.py
```diff
@@ -364,22 +364,33 @@ def _add(app_resource):
...
+                error_msg += "\n"
+            error_msg += "Application '{}' already exists".format(app_id)
+
+    # Abort entire deployment if any app(s) already exist
+    if error_msg is not '':
+        raise DCOSException(error_msg)
```
jsancio: Let's write a tests for this condition. We are trying to add multiple applications and only one of them exists. The whole deployment should stop.

jsancio:
LGTM @lloesche . One minor comment. :shipit: when addressed.

lloesche:
After discussion with @jsancio about the best way to implement multi app support we decided to instead use groups exclusively.

PR #375, opened 2015-11-13
dcos-3512 send CLUSTER_ID to segment.io

tamarrow:


cli/dcoscli/analytics.py
```diff
@@ -261,6 +261,12 @@ def _base_properties(conf=None):
         logger.exception('Unable to find the hostname of the cluster.')
         dcos_hostname = None
 
+    try:
+        cluster_id = mesos.DCOSClient().metadata().get('CLUSTER_ID')
```
jsancio: @tamarrow Can you confirm that we only make this call when `core.reporting` is `True`. In other words we shouldn't make this call when reporting is disabled.

tamarrow: yep, I added a test.

tamarrow:
@jsancio bump

jsancio:
LGTM! :ship: ing it!

PR #377, opened 2015-11-16
Add question about which missing library it is?

mhenrixon:
There was no information about which library that is missing from the home-brew installation. Is this still a problem and if so could we add the details of which library is missing so that we can get home-brew updated eventually?


ssk2:
Good question @mhenrixon, let me investigate.

ssk2:
@mhenrixon, I haven't been able to replicate this issue, so I'm removing that line from the readme. Thanks for raising the issue!

PR #390, opened 2015-12-15
dcos-4107 check that assets don't have same filename in `package bundle`

tamarrow:


cli/tests/data/package/create-directory/marathon.json.mustache
```diff
@@ -5,8 +5,8 @@
   "cpus": {{cassandra.framework.cpus}},
   "mem": {{cassandra.framework.mem}},
   "uris": [
-    "{{package.assets.uris.cassandra-mesos.tar.gz}}",
-    "{{package.assets.uris.jre-linux.tar.gz}}"
+    "{{resource.assets.uris.cassandra-mesos}}",
+    "{{resource.assets.uris.jre-linux}}"
```
jsancio: I thought was agreed that we would turn this to:

```
{{resource.assets.uris.cassandra-mesos-tar-gz}}
{{resource.assets.uris.jre-linux-tar-gz}}
```

I.e. replace `.` with `-`.

cli/tests/integrations/test_package_bundle.py
```diff
@@ -60,6 +60,18 @@ def test_bundle_fail_missing_package_json():
...
+def test_bundle_fail_duplicate_asset_names():
+    common.assert_command(
+        ['dcos', 'package', 'bundle', '--output-directory=/tmp',
+         'tests/data/package/invalid-asset-names-directory'],
+        returncode=1,
+        stderr=(b'Error bundling package. Multiple assets have the '
+                b'same filename [\'cassandra-mesos.gz\', '
+                b'\'cassandra-mesos.tar.gz\']\n'))
+
+    assert not glob.glob(_PACKAGE_NAME_GLOB)
```
jsancio: We need to change the test so that the file are `cassandra-mesos.tar.gz` and `cassandra-mesos-tar-gz`.

cli/dcoscli/package/main.py
```diff
@@ -682,7 +682,14 @@ def _bundle_uris(uris_directory, zip_file):
...
     """
 
-    for filename in sorted(os.listdir(uris_directory)):
+    uris = sorted(os.listdir(uris_directory))
+    # these uris will be found through a mustache template from the filename
+    # (without the extension) so make sure each name is unique.
+    uri_names = set(name.split('.')[0] for name in uris)
+    if len(uris) != len(uri_names):
+        raise DCOSException('Error bundling package. Multiple assets have the '
+                            'same filename {}'.format(uris))
```
jsancio: Let's change the wording a little: "Error bundling package. Multiple assets map to the same property name (periods [.] are replaced with dash [-]): ..."

What do you think?

jsancio:
Looks good!

cli/tests/data/package/create-directory/marathon.json.mustache
tamarrow: Yes, we should probably specify this will happen in the universe README as well

tamarrow: We're going to need to change the mustache template to s/./- after resource.assets.uris when we implement `package import`

tamarrow:
@jsancio updated

cli/dcoscli/package/main.py
```diff
@@ -682,10 +682,20 @@ def _bundle_uris(uris_directory, zip_file):
     :rtype: None
     """
 
-    for filename in sorted(os.listdir(uris_directory)):
+    uris = sorted(os.listdir(uris_directory))
+    # these uris will be found through a mustache template from the property
+    # name so make sure each name is unique.
+    uri_names = set("-".join(name.split('.')) for name in uris)
```
jsancio: Hmm. This is confusing to read. Using the replace operation is easier to read, no? `name.replace('.', '-')`

jsancio:
The two rename in this commit don't look right, no?

cli/dcoscli/package/main.py
```diff
@@ -682,10 +682,20 @@ def _bundle_uris(uris_directory, zip_file):
...
+        raise DCOSException(
+            'Error bundling package. Multiple assets map to the same property '
+            'name (periods [.] are replaced with dashes [-]): {}'.format(uris))
+    for filename in uris:
         fullpath = os.path.join(uris_directory, filename)
 
-        zip_file.write(fullpath, arcname='assets/uris/{}'.format(filename))
+        # s/./- so can be rendered by mustache template.
+        zip_file.write(fullpath, arcname='assets/uris/{}'.format(
+            "-".join(filename.split("."))))
```
jsancio: Same as above. Using `replace` is easier to read.

cli/tests/data/package/create-directory/marathon.json.mustache
jsancio: @tamarrow This is a feature of the bundle command so we should document this behaviour here. The Universe specification doesn't say anything about making zip packages.

Also, I am not sure I follow your comment around `package import`. The mustache templates should already be using `-` instead of `.` or else `version-2.x` branch in the `universe` repo wouldn't work. During the import operation we don't touch the name of the properties in `resource.json`. We only `rebase` the URLs in their values.

cli/dcoscli/package/main.py
jsancio: Wait. Ignore my comment above. We shouldn't be renaming the file. We should only be check that renaming the property doesn't collide. The reason why we perform the check above is so that we guarantee that two files don't have the same property name.

We do this because mustache resolve properties with `.` differently. The reason why we can't change the filename is because the code executing in the Mesos task expects the files to have the same name.

jsancio:
Looks like github doesn't distinguish between new binary file and replaced binary files. I don't think this commit should change this file:  `cli/tests/data/package/cassandra.zip`.

tamarrow:
@jsancio updated. I also changed the `marathon.json.mustache` files in the test data with the new spec, which caused me to update `cassandra.zip`

cli/dcoscli/package/main.py
```diff
@@ -682,7 +682,15 @@ def _bundle_uris(uris_directory, zip_file):
...
 
-    for filename in sorted(os.listdir(uris_directory)):
+    uris = sorted(os.listdir(uris_directory))
+    # these uris will be found through a mustache template from the property
+    # name so make sure each name is unique.
+    uri_names = set(name.replace('.', '-') for name in uris)
+    if len(uris) != len(uri_names):
+        raise DCOSException(
+            'Error bundling package. Multiple assets map to the same property '
+            'name (periods [.] are replaced with dashes [-]): {}'.format(uris))
```
jsancio: @tamarrow `...format(uris)` list all of the packages not just the ones that collide.

jsancio:
Thanks. Looks good. I will merge once the tests pass.

PR #401, opened 2016-01-07
parameterize ccm cluster channel

tamarrow:


BenWhitehead:
Updated rules and build reporting from teamcity. this should be good to merge.

PR #403, opened 2016-01-08
install using cosmos server

tamarrow:
This works with the helloworld package only right now.

### DIFF
--- a/dcos/package.py
+++ b/dcos/package.py
@@ -38,35 +38,6 @@ PACKAGE_REGISTRY_VERSION_KEY = 'DCOS_PACKAGE_REGISTRY_VERSION'
 PACKAGE_FRAMEWORK_NAME_KEY = 'DCOS_PACKAGE_FRAMEWORK_NAME'
 
 
-def install_app(pkg, revision, init_client, options, app_id):
-    """Installs a package's application
-
-    :param pkg: the package to install
-    :type pkg: Package
-    :param revision: the package revision to install
-    :type revision: str
-    :param init_client: the program to use to run the package
-    :type init_client: object
-    :param options: package parameters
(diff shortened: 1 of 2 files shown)


dcos/package.py
```diff
@@ -1636,3 +1607,69 @@ def get_apps_for_framework(framework_name, client):
...
+        :param revision: the package revision to install
+        :type revision: str
+        :param options: package parameters
+        :type options: dict
+        :param app_id: app ID for installation of this package
+        :type app_id: str
+        :rtype: None
+        """
+        url = urllib.parse.urljoin(self.cosmos_url, 'v1/package/install')
+        data = '{"name": "helloworld"}'
```
tamarrow: I ended up hardcoding this for now. I think there's some issue with parsing out the data, kept getting `400`s using json or other formats

dcos/package.py
```diff
@@ -1636,3 +1607,69 @@ def get_apps_for_framework(framework_name, client):
...
+
+        # Send the descriptor to init
+        init_client.add_app(init_desc)
+
+class Cosmos(PackageManager):
+    """Implementation of Package Manager using Cosmos"""
+    def __init__(self, cosmos_url):
+        self.cosmos_url = cosmos_url
+
+    def install_app(self, pkg, revision, options, app_id):
```
tamarrow: options isn't used right now, but we'll need it to pass options the user supplies from the command line

dcos/package.py
jsancio: Why hardcode to "helloworld"?

tamarrow: that's the only package that works right now

cruhland:
LGTM

PR #404, opened 2016-01-11
Remove EXHIBITOR_URL environment variable

jsancio:
We are not using the environment variable during the tests. Let's remove
all reference to it.


tamarrow:
:+1: we can also get rid of `VBOX_IP`

PR #411, opened 2016-01-22
cli for cosmos full install + describe

tamarrow:


cli/dcoscli/package/main.py
```diff
@@ -211,21 +211,13 @@ def _describe(package_name,
             'If --package-versions is provided, no other option can be '
             'provided')
 
-    pkg = package.resolve_package(package_name)
-    if pkg is None:
-        raise DCOSException("Package [{}] not found".format(package_name))
+    package_manager = package._get_package_manager()
```
jsancio: We are calling a private function.

cli/dcoscli/package/main.py
```diff
@@ -356,35 +332,31 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
             emitter.publish('Exiting installation.')
             return 0
 
-    options = pkg.options(pkg_revision, user_options)
+    # render options before start installation
+    options = pkg.options(user_options)
 
-    revision_map = pkg.package_revisions_map()
-    package_version = revision_map.get(pkg_revision)
+    if app and pkg.has_mustache_definition():
```
jsancio: I think we should change what Cosmos's describe returns so that it doesn't include the mustache. This means that we should just tell Cosmos to install it. If it doesn't exists then Cosmos should just return a well known error and we handle it on our side.

dcos/package.py
```diff
@@ -1553,6 +1520,350 @@ def __repr__(self):
...
+        """
+
+        return self._marathon_template
+
+    def has_mustache_definition(self):
+        """Dummy method since all packages in cosmos must have mustache
+           definition.
+        """
+
+        return True
```
jsancio: Ic. Ignore my previous comment. I get it now.

dcos/package.py
```diff
@@ -1625,14 +1936,14 @@ def _get_package_manager():
 class PackageManager():
     """Package Manager using local file system"""
 
-    def install_app(self, pkg, revision, options, app_id):
+    def install_app(self, pkg, options, app_id):
         """Installs a package's application
 
         :param pkg: the package to install
-        :type pkg: Package
+        :type pkg: PackageVersion
```
jsancio: we removed the revision parameter below.

dcos/subcommand.py
```diff
@@ -230,16 +228,19 @@ def _write_package_source(pkg):
     source_path = os.path.join(pkg_dir, 'source')
 
     with util.open_file(source_path, 'w') as source_file:
-        source_file.write(pkg.registry.source.url)
+        registry = pkg.registry()
+        if registry == "cosmos":
+            url = pkg.cosmos_url()
```
jsancio: I need to understand this better but this doesn't look right. We are suppose to write the repository url, no? Looks like we are writing cosmos' url which may include multiple repositories.

I think we need this to uniquely identify a package. A package is uniquely identified by the tuple `(repo/source url, package name, package revision)`.

PR #419, opened 2016-01-26
endpoint for package search using cosmos

tamarrow:


dcos/package.py
```diff
@@ -2045,6 +2059,23 @@ def uninstall_app(self, app_name, remove_all, app_id):
...
+        :rtype: [packages]
+        """
+        url = urllib.parse.urljoin(self.cosmos_url, 'v1/package/search')
+        response = http.post(url, json={"query": query})
+
+        if response.status_code != 200:
+            raise DCOSException(response.json().get("message"))
+
+        packages = response.json()
+        return [{"packages": packages, "source": "cosmos"}]
```
jsancio: Okay @tamarrow. Let's review this tomorrow. I suspect that `"source": "cosmos"` is not what we want.

tamarrow: no, it's really just a placeholder until we add support for sources. ill add an issue for it

PR #423, opened 2016-01-26
dcos-4134 integrate with acs AuthN flow

tamarrow:


dcos/http.py
```diff
@@ -311,55 +314,87 @@ def _get_basic_auth_credentials(username, hostname):
...
+    def __call__(self, r):
+        dcos_url = util.get_config_vals(
+            ['core.dcos_url'], util.get_config())[0]
+        url = urllib.parse.urljoin(dcos_url, 'acs/api/v1/auth/login')
+        creds = {"uid": self.uid, "password": self.password}
+
+        # using private method here, so we don't retry on this request
+        # error here will be bubbled up to _request_with_auth
+        response = _request('post', url, json=creds)
+        if response.status_code == 200:
```
jsancio: What happens if the `status_code` is not `200`?

jsancio: I think I get it. If the header is not set then the following call will get fail with a `400`.

tamarrow:
@jsancio updated with changes we talked about today

dcos/http.py
```diff
@@ -311,55 +314,102 @@ def _get_basic_auth_credentials(username, hostname):
...
+    url = urllib.parse.urljoin(dcos_url, 'acs/api/v1/auth/login')
+    creds = {"uid": uid, "password": password}
+
+    # using private method here, so we don't retry on this request
+    # error here will be bubbled up to _request_with_auth
+    response = _request('post', url, json=creds)
+
+    token = None
+    if response.status_code == 200:
+        token = response.json()['token']
```
jsancio: Okay. Or `token = response.json().get('token') if response.status_code == 200 else None` if we want functional code.

jsancio:
Looks great! :shipit:

PR #427, opened 2016-01-28
deprecate `package update`

tamarrow:


cli/dcoscli/package/main.py
```diff
@@ -160,9 +160,10 @@ def _update(validate):
     :rtype: int
     """
 
-    config = util.get_config()
+    emitter.publish("This command is being deprecated")
```
jsancio: Hmm. I think we should only print this when the CLI is configure to talk to `Cosmos`. Eventually, we are going to add code that auto discover if Cosmos is installed and configure it automatically.

jsancio:
LGTM. Just one comment to address. Let me know what you think. Thanks!

cli/dcoscli/package/main.py
tamarrow: sgtm

tamarrow:
@jsancio updated

cli/tests/integrations/test_package.py
```diff
@@ -62,10 +66,13 @@ def test_update_with_validation():
...
-    assert b'source' in stdout
-    assert b'Validating package definitions...' in stdout
-    assert b'OK' in stdout
     assert stderr == b''
+    if os.environ.get(constants.COSMOS_URL_ENV) is None:
+        assert b'source' in stdout
+        assert b'Validating package definitions...' in stdout
+        assert b'OK' in stdout
+    else:
+        assert stdout == b'This command is being deprecated\n'
```
jsancio: Sorry should have mentioned this before. "This command is deprecated\n". From their point of view it already happened. :)

jsancio:
One minor wording comment. Looks good to me!

jsancio:
:ship: it

PR #429, opened 2016-01-29
add cosmos headers to requests

tamarrow:


dcos/cosmospackage.py
```diff
@@ -289,9 +296,50 @@ def package_versions(self):
...
+
+    :param response: response from cosmos
+    :type response: Response
+    :returns: whether response has an error or not
+    :rtype: bool
+    """
+    content_type = response.headers.get('Content-Type')
+    if content_type is None:
+        return True
+    return _get_header("error") in content_type
```
jsancio: If we change this method to something like `_check_for_cosmos_error` and make it throw a `DCOSException` the we can change the `if ... : raise ...` pattern to just `_check_for_cosmos_error`. Even better we can use Python's decorators: http://simeonfranklin.com/blog/2012/jul/1/python-decorators-in-12-steps/

We have an example at: https://github.com/mesosphere/dcos-cli/blob/master/dcos/util.py#L533-L553

BenWhitehead:
@tamarrow LGTM, though we may need to update the value for the `vnd` namespace, but that will be decided latter and should be an easy fix.

tamarrow:
@jsancio updated to use a decorator! much nicer now

dcos/cosmospackage.py
```diff
@@ -152,6 +140,56 @@ def update_sources(self, validate=False):
...
+        :rtype: Response
+        """
+
+        url = urllib.parse.urljoin(self.cosmos_url,
+                                   'v1/package/{}'.format(request))
+        try:
+            response = http.post(url, json=params,
+                                 headers=_get_cosmos_header(request))
+        except DCOSHTTPException as e:
+            response = e.response
```
jsancio: Can you write a comment as to why we are doing this? This is not clear to me.

jsancio:
LGTM.

dcos/cosmospackage.py
tamarrow: done

PR #452, opened 2016-02-09
update message from marathon error from cosmos change

tamarrow:


jsancio:
@tamarrow Let's add a test for this so we don't regress by accident.

tamarrow:
@jsancio added test. This only passes for `cosmos` since we're about to break legacy in `cosmos` branch anyways.

jsancio:
LGTM.

PR #457, opened 2016-02-10
Make NamedTemporaryFile play nice on Windows

surdy:
On Windows `NamedTemporaryFile` does not lend itself to be opened the second time when it is already open. Source : https://docs.python.org/2/library/tempfile.html

> 
Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). 

Also closing the file would delete it by default. So to open the file second time on windows(and also not break on Unix like systems) the workaround is to set `delete=False` while creating the file, closing it after use and then manually deleting it after done.  

Without this fix you would see errors like : 
`
2016-02-09 16:09:02,290 c:\users\surdy\downloads\dcos\lib\site-package\dcos\subcommand.py:_execute_install:411 - Install script's stdout: b''
2016-02-09 16:09:02,290 c:\users\surdy\downloads\dcos\lib\site-packages\dcos\subcommand.py:_execute_install:412 - Install script's stderr: b"Could not open requirements file: [Errno 13] Permission denied: 'C:\\\\Users\\\\surdy\\\\AppData\\\\Local\\\\Temp\\\\tmpe9tm00ft'\r\nYou are using pip version 7.1.2, however version 8.0.2 is available.\r\nYou should consider upgrading via the 'python -m pip install --upgrade pip' command.\r\n
`

### DIFF
--- a/dcos/subcommand.py
+++ b/dcos/subcommand.py
@@ -4,7 +4,6 @@ import json
 import os
 import shutil
 import subprocess
-import tempfile
 
 from dcos import constants, util
 from dcos.errors import DCOSException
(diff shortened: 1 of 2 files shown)


jsancio:
@surdy and I talked offline. He is working on a contextmanager for this.

surdy:
Thanks @jsancio for the pointers on contextmanager. 

@tamarrow I have now resurrected the util.temptext function to replace NamedTemporaryFile for windows portability.

tamarrow:
@surdy thanks for contributing! Can you fix the syntax issue and then we can :ship:!

tamarrow:
@surdy looks like you have to rebase

PR #465, opened 2016-02-12
mock post request in acl test

tamarrow:


cli/tests/integrations/test_http_auth.py
```diff
@@ -70,11 +70,13 @@ def test_get_http_auth_credentials_basic(auth_mock):
...
 
 @patch('dcos.http._get_auth_credentials')
-def test_get_http_auth_credentials_acl(auth_mock):
+@patch('dcos.http._request')
+def test_get_http_auth_credentials_acl(req_mock, auth_mock):
     m = Mock()
     m.url = 'http://domain.com'
     m.headers = {'www-authenticate': 'acsjwt"'}
     auth_mock.return_value = ("username", "password")
+    req_mock.status_code = 404
```
jsancio: is 404 really the return code?

jsancio: Also @tamarrow are we able to tell the difference from user not logged in (do the login flow) and user don't have permission (fail operation)?

tamarrow: this is just a mock, we don't actually hit the endpoint. But yes, we can tell the difference between the two

PR #467, opened 2016-02-13
don't send CLUSTER_ID to segment for config subcommand and just 'dcos'

tamarrow:


jsancio:
We already do this: https://github.com/mesosphere/dcos-cli/blob/master/cli/dcoscli/main.py#L42-L44

Can we make sure that we are adding this check in the right place?

tamarrow:
that just make sure the user is registered, but doesn't stop us reporting. that's why i added tests!

cli/dcoscli/analytics.py
```diff
@@ -261,11 +262,12 @@ def _base_properties(conf=None):
...
-        cluster_id = mesos.DCOSClient().metadata().get('CLUSTER_ID')
-    except:
-        logger.exception('Unable to get the cluster_id of the cluster.')
-        cluster_id = None
+    cluster_id = None
+    if command and command != "config":
+        try:
+            cluster_id = mesos.DCOSClient().metadata().get('CLUSTER_ID')
+        except:
+            logger.exception('Unable to get the cluster_id of the cluster.')
```
jsancio: Let's log the exception too.

jsancio:
:+1: :+1: LGTM. One minor comment!

PR #468, opened 2016-02-17
Enable --options to interpret ~ in the path

surdy:
Currently `dcos package install --options=~/path/to/options.json` or `dcos package install --options=~surdy/path/to/options.json` fails silently as it cannot process `~`. 

I have used `os.path.expanduser`to get around this problem

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -207,6 +207,8 @@ def _describe(package_name,
     # render the template
     if options_path:
         render = True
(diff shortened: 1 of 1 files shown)


jsancio:
@surdy Let's add a test. Checkout out `cli/tests/integrations/test_package.py`

surdy:
@jsancio Cool, will do so and update. Thanks :)

jsancio:
LGTM! :shipit:

PR #477, opened 2016-02-18
error if user runs legacy cli with cosmos enabled dcos

tamarrow:
#477

### DIFF
--- a/dcos/cosmospackage.py
+++ b/dcos/cosmospackage.py
@@ -1,0 +1,53 @@
+from dcos import emitting, http, util
+
+from six.moves import urllib
+
+logger = util.get_logger(__name__)
+
+emitter = emitting.FlatEmitter()
+
+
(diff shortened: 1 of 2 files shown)


dcos/cosmospackage.py
```diff
@@ -0,0 +1,53 @@
+from dcos import emitting, http, util
```
jsancio: Okay. We are going to get a merge conflict when we merge the cosmos branch.

jsancio:
LGTM. Did we test this manually?

dcos/cosmospackage.py
tamarrow: shouldn't we not merge this commit with the cosmos branch since it has the opposite functionality re need (ie not letting users use the package subcommands if they have cosmos)?

tamarrow:
yep, merging!

PR #478, opened 2016-02-18
check for cosmos in each `package` subcommand call

tamarrow:
We need to have _main always return all available commands to show the
user


jsancio:
:+1:

PR #487, opened 2016-02-23
error based on user command auth not auth to get CLUSTER_ID

tamarrow:
We try to get the CLUSTER_ID before we run the user specified command.
If the user if authenticated, but not authorized (to get the
CLUSTER_ID), let them continue and error based on access to the command
they ran.

### DIFF
--- a/cli/dcoscli/main.py
+++ b/cli/dcoscli/main.py
@@ -9,7 +9,7 @@ import docopt
 import pkg_resources
 from dcos import (auth, constants, emitting, errors, http, mesos, subcommand,
                   util)
-from dcos.errors import DCOSException
+from dcos.errors import DCOSAuthenticationException, DCOSException
 from dcoscli import analytics
 
 logger = util.get_logger(__name__)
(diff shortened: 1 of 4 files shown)


cli/dcoscli/main.py
```diff
@@ -60,8 +60,9 @@ def _main():
     if dcoscli.version != 'SNAPSHOT' and command and command != "config":
         try:
             cluster_id = mesos.DCOSClient().metadata().get('CLUSTER_ID')
-        except DCOSException:
-            raise
+        except DCOSException as e:
+            if "Authentication failed" in str(e):
```
jsancio: Should this be `... not in str(e)`?

Also, should we instead throw an specific `DCOSAuthException` and catch that instead? This exception can inherit `DCOSHTTPException`.

tamarrow: no, we want to throw the 401 (Authenticated failed) error, but not the 403 error, because that could be valid in the user specified command. I'll add `DCOSAuthException` so that the code is more readable

jsancio: I see, then it sounds like you want to throw and catch `DCOSAuthorizationException`

tamarrow:
@jsancio updated

jsancio:
LGTM!

Issue #492, opened 2016-02-24
Able to add unsupported file:///path/to/universe repos


nickbp:
If they're no longer supported, this should return an error. At the moment it succeeds, only producing errors later:

dcos package repo add local file:///path/to/universe/ (success)
dcos package search kafka (NOW it fails)

tamarrow:
@cruhland mind explaining the rationale for this?

cruhland:
There are actually two distinct issues here:

1. `file://` URLs no longer work with the Cosmos-enabled CLI (and if we don't want to support them, we should have the CLI immediately return an error on `repo add`)
1. In general, Cosmos will not be able to respond with an error immediately if a repo is added that is invalid in some way, because it asynchronously downloads the repo contents. For now, we have minimal error detection for `repo add` and instead report errors when the repositories are used.

cc @tamarrow @nickbp

nickbp:
Maybe add some server-side synchronous verification on the url before starting the async add operation?

But ultimately it feels like 'repo add' should just be synchronous. I imagine most users would prioritize knowing that the 'repo add' really succeeded (and getting direct feedback if it didn't) over being quickly returned to a commandline

tamarrow:
We have an issue for this against cosmos, so I'm closing this one.

PR #497, opened 2016-02-25
error if adding repo with unsupported protocol

tamarrow:
fixes part of #492


jsancio:
Let's not implement this here. Cosmos should do this for us instead. Please file an issues against it if we agree.

tamarrow:
Sure! Filed an issue against cosmos.

PR #500, opened 2016-02-25
allow boolean values to be specified as upper or lower case

### LINKED ISSUE #495: core.reporting value displays as uppercase in "dcos config show", but must be set as lowercase in "dcos config set"
nick@augustao:~/code/dcos-cli$ dcos config show
core.reporting=True
nick@augustao:~/code/dcos-cli$ dcos config set core.reporting False
Unable to parse 'False' as a boolean: No JSON object could be decoded
nick@augustao:~/code/dcos-cli$ dcos config set core.reporting false
[core.reporting]: changed from 'True' to 'False'

tamarrow:
When setting a boolean value in the config the value had to be all lowercase because we we're reading the value in as JSON. However, we display the config values as python values, where boolean values are 'True/False'. This is confusing to the user, so instead allow the user to specify boolean values as upper or lower case. Fixes #495


tamarrow:
Once this is merged, I will cherry-pick the commit to the `legacy` branch

jsancio:
@tamarrow I don't think we need to cherry pick this to legacy doesn't look like a critical fix.

jsancio:
LGTM.

PR #505, opened 2016-02-26
Feature/framework id support

abridgett:
This adds the ability to find/list/log tasks filtered by framework ID.  Some frameworks (Spark for example) don't provide very unique task IDs and this is easier into the bargain.

It also fixes one issue with mesos.py where when "completed" was set, _only_ completed tasks were returned (the documentation claimed both were).

One buglet I can't see how to resolve is that the argument parsing forces the user to provide a task name for the "ls" and "logs" commands (I've tried a few approaches to no avail).

### DIFF
--- a/cli/dcoscli/data/help/task.txt
+++ b/cli/dcoscli/data/help/task.txt
@@ -2,13 +2,15 @@ Manage DCOS tasks
 
 Usage:
     dcos task --info
-    dcos task [--completed --json <task>]
-    dcos task log [--completed --follow --lines=N] <task> [<file>]
-    dcos task ls [--long] <task> [<path>]
+    dcos task [--completed] [--json] [--framework=<framework_id>] [<task>]
+    dcos task log [--completed] [--follow] [--lines=N] [--framework=<framework_id>] <task> [<file>]
+    dcos task ls [--long] [--framework=<framework_id>] <task> [<path>]
 
 Options:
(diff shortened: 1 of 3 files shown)


jsancio:
@tamarrow can you take a look?

abridgett:
Actually, there's a problem somewhere :-(  I'll do some more testing and find out what's going on.

abridgett:
that's fixed one problem, still looking at the other (the directory that both ls and logs look at on the slave does not exist - the last ID is incorrect)

abridgett:
Sorry about that - this seems to be better! I'll wait a bit and test more thoroughly next time.

tamarrow:
@abridgett can you add some tests for this please?

abridgett:
I'd love to but I must admit I'm finding it rather difficult to figure out how (mocked or unmocked), probably want to grab the frameworkID of marathon and look for those tasks (it seems that most test apps are run via marathon)

tamarrow:
We should add tests here: https://github.com/mesosphere/dcos-cli/blob/master/cli/tests/integrations/test_task.py. These should give you some good test examples. Let me know if I can help. thanks!

tamarrow:
@abridgett anything I can help with?

tamarrow:
@abridgett I am closing this pr. If you would like to keep working on it, feel free to re-open.

PR #509, opened 2016-02-29
update `dcos package repo remove` to only take repo-name as paramater

### LINKED ISSUE #496: "dcos package repo remove" could require less typing
At the moment, 'package repo remove' is requiring either --repo-name or --repo-url. But the CLI has access to the list of existing repos and should be able to autodetect what the user means

For example, if I had a repo list like this:

main: https://example.com/version-2.x.zip
test: https://example.com/test.zip

Then the CLI should be able to check for presence in that list to automatically decide what either of these mean:

"dcos package repo remove test"
"dcos package repo remove https://example.com/version-2.x.zip"

If this autodetection creates other problems of its own, maybe just make "--repo-name" optional and default to assuming that the user is entering a repo name if they don't explicitly specify --repo-url?

tamarrow:
fixes #496


jsancio:
LGTM!

PR #515, opened 2016-03-02
link users on how to downgrade if using legacy dcos with new cli

tamarrow:


tamarrow:
issue #515

cli/dcoscli/package/main.py
```diff
@@ -763,7 +763,8 @@ def _get_package_manager():
     if cosmos_manager.enabled():
         return cosmos_manager
     else:
-        msg = ("This version of the dcos-cli is unsupported for your DCOS "
-               "cluster. Please use a dcos-cli version < 0.4.0 or upgrade your"
-               " cluster to DCOS version >= 1.6.1")
+        msg = ("This version of the DCOS CLI is not supported for your "
+               "cluster. Please downgrade the CLI to an older version: "
+               "https://docs.mesosphere.com/administration/introcli/updatecli/"
```
jsancio: That link doesn't tell the user how to downgrade. We need instruction to use the legacy install script.

tamarrow: scroll down

jsancio: I see it. @joel-hamill can we create a separate page for installing the legacy CLI? If I was confused, I am sure our users will be too.

joel-hamill: @jsancio @tamarrow Is this good enough, or does it actually need it's own page: https://docs.mesosphere.com/administration/cli/updatecli/#scrollNav-2.

tamarrow: works for me, updated

PR #516, opened 2016-03-03
allows use same cassandra framework in test

tamarrow:
Since we now have two version in the repo.

### DIFF
--- a/cli/tests/integrations/test_service.py
+++ b/cli/tests/integrations/test_service.py
@@ -54,7 +54,7 @@ def test_service_table():
 
 
 def test_service_inactive():
-    package_install('cassandra', True)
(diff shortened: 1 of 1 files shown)


cli/tests/integrations/test_service.py
```diff
@@ -93,7 +93,7 @@ def test_service_inactive():
 
 
 def test_service_completed():
-    package_install('cassandra', True)
+    package_install('cassandra', True, ['--package-version=0.2.0-1'])
```
jsancio: Okay for now. Are we solving it this way because the latest Cassandra had too many changes? Let's file an issue to use the latest version.

jsancio:
LGTM.

Issue #449, opened 2016-02-09
`dcos config show` breaks when no config exists


cmaloney:
1. Install the DCOS CLI
1. Notice you have some cached things and `rm -rf ~/.dcos`
1. do a `dcos config show` to try and figure out how to set the cluster URL
```
$dcos config show

Cannot create file [/home/cody/.dcos/dcos.toml]: [Errno 2] No such file or directory: '/home/cody/.dcos/dcos.toml'
```

Doing a `mkdir ~/.dcos` fixes the error / allows me to run `dcos config set core.dcos_url http://<dcos_url>`

jsancio:
@cmaloney Thanks! This was intentional. The CLI will create `dcos.toml` and the `cache` directory but it doesn't create $DCOS_CONFIG (or `~/.dcos` the default). I could be convinced otherwise :).

I think we decided not to do this in case the user sets $DCOS_CONFIG to something like `/usr/something/.dcos` by accident. Else we would have to do `mkdir -p $DCOS_CONFIG` for this to work in all cases.

cc @tamarrow

cmaloney:
I'd vote for creating it in the default directory by default rather than failing to create a file inside the directory. It's how most command line utilities I use work (Although most programs drop their default config into `~/.config/<name>` nowadays).

At the very least a `dcos config show` / purely read operation shouldn't fail because the config doesn't exist with a can't create error. It should either say something like "config file doesn't exist. To make one do X" or preferrably to me print out that the current config is empty, and if I do a `dcos config set` create the file and put my basic options into.

tamarrow:
I'm in favor of having `dcos config show` give a more helpful error and tell the user the create a dir. What do you think @jsancio ?

PR #527, opened 2016-03-11
create config toml at `DCOS_CONFIG` if doesn't exist

tamarrow:


tamarrow:
#449

dcos/util.py
```diff
@@ -130,6 +130,7 @@ def ensure_file_exists(path):
 
     if not os.path.exists(path):
         try:
+            ensure_dir_exists(os.path.dirname(path))
```
jsancio: We should only create directory for DCOS_CONFIG when the value is the default. We should not create the directories when the user has overwritten their DCOS_CONFIG.

tamarrow:
@jsancio updated

jsancio:
:feet:

PR #530, opened 2016-03-17
Rename `node --master` -> `node --leader`

branden:
See https://mesosphere.atlassian.net/browse/DCOS-5441.

### DIFF
--- a/cli/dcoscli/node/main.py
+++ b/cli/dcoscli/node/main.py
@@ -29,6 +29,11 @@ def _main():
         _doc(),
         version="dcos-node version {}".format(dcoscli.version))
 
+    if args.get('--master'):
+        raise DCOSException(
+            '--master has been deprecated. Please use --leader.'
+        )
+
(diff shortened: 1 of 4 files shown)


tamarrow:
@branden can we add a deprecation notice for `--master`?

cli/dcoscli/node/main.py
```diff
@@ -24,9 +25,16 @@ def main():
 @decorate_docopt_usage
 def _main():
     util.configure_process_from_environ()
+    argv = sys.argv[1:]
```
branden: I'd much rather let `docopt.docopt()` decide what `argv` is, but I'm not sure how else to add this deprecation notice without adding `--master` back to the usage string.

tamarrow: yes, sadly you have to also add `--master` back into the usage string.

tamarrow:
@branden the tests failed from a syntax error: dcoscli/node/main.py:3:1: F401 'sys' imported but unused

PR #532, opened 2016-03-17
add ssl verification for acs auth login

tamarrow:
We use the private `_request` method for acs auth login which doesn't have the ssl check in the main `request` method, so we need to add that check for that request.


tamarrow:
The same fix for legacy versions: https://github.com/mesosphere/dcos-cli/pull/533

PR #533, opened 2016-03-17
add ssl verification for acs auth login

tamarrow:


dcos/http.py
```diff
@@ -182,16 +205,7 @@ def request(method,
...
-        verify = os.environ[constants.DCOS_SSL_VERIFY_ENV]
-        if verify.lower() == "true":
-            verify = True
-        elif verify.lower() == "false":
-            verify = False
-
-    # Silence 'Unverified HTTPS request' and 'SecurityWarning' for bad certs
-    if verify is not None:
-        silence_requests_warnings()
+    verify = _verify_ssl(verify)
```
jsancio: We never use the variable `verify`.

dcos/http.py
```diff
@@ -414,7 +428,7 @@ def _get_dcos_acs_auth(username, password, hostname):
 
         # using private method here, so we don't retry on this request
         # error here will be bubbled up to _request_with_auth
-        response = _request('post', url, json=creds)
+        response = _request('post', url, json=creds, verify=_verify_ssl())
```
jsancio: There is side effect in `_verify_ssl`. This style of coding is hard to debug and understand in the future. How can we fix this?

dcos/http.py
tamarrow: we use it on line 211

dcos/http.py
tamarrow: I'll move the call to `silence_requests_warning` outside of `_verify_ssl` so there aren't side effects.

dcos/http.py
jsancio: Sounds good. I still would like us to get rid of the side effect in `_verify_ssl` and maybe move it to `_request` what do you think?

tamarrow: I don't want to move `_verify_ssl` to `_request` because in the case of auth, we might call `_request` multiple times and we'll be doing unnecessary calls. But ill remove the side effects by moving the call to  `silence_requests_warning` outside of `_verify_ssl`

jsancio: :+1:

jsancio:
I take it this doesn't pass because we are unable to test the CLI against old clusters. How are we making sure that the CLI still works?

tamarrow:
The package tests wont pass in legacy (the current fail is marathon bug, fix is here: https://github.com/mesosphere/dcos-cli/pull/534). I'm running this branch against an old cluster on teamcity using the build that allows we run the tests on a specific ccm channel.

jsancio:
LGTM after we address the side effect comment above. :shipit:

PR #534, opened 2016-03-18
update test file due to marathon being more strict with app ids

tamarrow:
This is causing our tests to fail. Filed as issue with marathon https://github.com/mesosphere/marathon/issues/3529 for the regression


jsancio:
LGTM.

PR #542, opened 2016-03-24
update vanilla cli to be one executable

tamarrow:


jsancio:
First pass comments.

dcos/subcommand.py
```diff
@@ -435,3 +465,52 @@ def package_json(self):
...
+        with self.pool:
+            self.future = self.pool.submit(self.target)
+
+    def exitcode(self):
+        if self.exception() is not None:
+            return 1
+        else:
+            return self.future.result()
+
+    def exception(self):
```
jsancio: What is the return type for this method?

dcos/subcommand.py
```diff
@@ -435,3 +465,52 @@ def package_json(self):
...
+        self.command = command
+        self.args = args
+        self.subproc = Popen([self.executable,  self.command] + self.args,
+                             stderr=PIPE)
+        self.err = None
+
+    def exitcode(self):
+        return self.subproc.poll()
+
+    def exception(self):
```
jsancio: What is the return type for this method?

cli/dcoscli/config/main.py
```diff
@@ -155,3 +155,7 @@ def _validate():
 
     emitter.publish("Congratulations, your configuration is valid!")
     return 0
+
+
+if __name__ == "__main__":
+    main()
```
jsancio: Why do we need this? I don't see this change for the other "main"?

jsancio:
@tamarrow I am reviewing this.

cli/dcoscli/main.py
```diff
@@ -115,32 +126,11 @@ def signal_handler(signal, frame):
...
-        return result
-    return wrapper
-
-
 def set_ssl_info_env_vars(config):
     """Set SSL info from config to environment variable if enviornment
        variable doesn't exist
 
-    :param config: config
+    :paceback.print_exc()ram config: config
```
jsancio: This doesn't look right.

cli/dcoscli/main.py
```diff
@@ -67,17 +64,31 @@ def _main():
             msg = 'Unable to get the cluster_id of the cluster.'
             logger.exception(msg)
 
+    # send args call to segment.io
+    reporting_executor = ThreadPoolExecutor(max_workers=2)
```
jsancio: Use `with ThreadPoolExecutor` so that it gets shutdown and more importantly we join/wait on all of the tasks finishing.

cli/dcoscli/main.py
```diff
@@ -67,17 +64,31 @@ def _main():
...
+    report = config.get('core.reporting', True)
+    if report and dcoscli.version != 'SNAPSHOT':
```
jsancio: I see that we do this in a couple of places. I think we should move this into the `analytics` module.

cli/dcoscli/main.py
```diff
@@ -67,17 +64,31 @@ def _main():
...
+        sc = subcommand.SubcommandMain(command)
     else:
-        return analytics.wait_and_capture(subproc)[0]
+        executable = subcommand.command_executables(command)
+        sc = subcommand.SubcommandProcess(executable, command, args['<args>'])
+
+    exitcode, err = sc.run_and_capture()
+
+    if err:
+        if dcoscli.version != 'SNAPSHOT' and report:
```
jsancio: I see that we do this in a couple of places. I think we should move this check into the `analytics` module.

cli/dcoscli/main.py
tamarrow: fixed

cli/dcoscli/main.py
```diff
@@ -67,17 +64,31 @@ def _main():
...
     # so that if you have auth enabled we don't ask for user/pass multiple
     # times (with the text being out of order) before we can cache the auth
     # token
-    subproc = Popen([executable,  command] + args['<args>'],
-                    stderr=PIPE)
-
-    if dcoscli.version != 'SNAPSHOT':
-        return analytics.wait_and_track(subproc, cluster_id)
+    if command in subcommand.default_subcommands():
+        sys.argv[1:] = [command] + args['<args>']
```
jsancio: I see. You are doing this since we need to remove `[options]` from the args that we pass to `docopt`. Instead of changing `sys.argv` why don't we just pass the array of string explicitly to `docopt`? They have an `argv` parameter that you can override.

Also, let's move this code to `subcommand.SubcommandMain`.

cli/dcoscli/main.py
```diff
@@ -67,17 +64,31 @@ def _main():
...
+        sc = subcommand.SubcommandMain(command)
```
jsancio: Let's do the same thing we do for `subcommand.SubcommandProcess`. We can turn this constructor to `subcommand.SubcommandMain(command, args['<args>'])`

cli/setup.py
```diff
@@ -92,14 +92,7 @@
...
         'console_scripts': [
-            'dcos=dcoscli.main:main',
-            'dcos-help=dcoscli.help.main:main',
-            'dcos-config=dcoscli.config.main:main',
-            'dcos-marathon=dcoscli.marathon.main:main',
-            'dcos-package=dcoscli.package.main:main',
-            'dcos-service=dcoscli.service.main:main',
-            'dcos-task=dcoscli.task.main:main',
-            'dcos-node=dcoscli.node.main:main'
+            'dcos=dcoscli.main:main'
```
jsancio: Exciting!!! :balloon:

cli/tests/integrations/test_node.py
```diff
@@ -180,13 +180,7 @@ def _node_ssh(args):
     stdout, stderr, returncode = _node_ssh_output(args)
     assert returncode is None
 
-    assert stdout
     assert b"Running `" in stderr
-    num_lines = len(stderr.decode().split('\n'))
-    expected_num_lines = 2 if '--master-proxy' in args else 3
-    assert (num_lines == expected_num_lines or
-            (num_lines == (expected_num_lines + 1) and
-             b'Warning: Permanently added' in stderr))
```
jsancio: Can you explain this to me?

dcos/config.py
```diff
@@ -203,9 +203,15 @@ def get_config_schema(command):
...
                 'data/config-schema/core.json').decode('utf-8'))
-
-    executable = subcommand.command_executables(command)
-    return subcommand.config_schema(executable)
+    elif command in subcommand.default_subcommands():
+        resource = "data/config-schema/{}.json".format(command)
+        return json.loads(
+            pkg_resources.resource_string(
+                'dcoscli', resource).decode('utf-8'))
+        return subcommand.default_config_schema(command)
```
jsancio: There are two return statements. I like the second one. We don't need `resource` if we do the second one.

dcos/subcommand.py
```diff
@@ -118,6 +124,30 @@ def distributions():
...
 
+def command_info(command):
+    """top level documentation of default DCOS-CLI command
+    :param command: name of command
+    :param type: str
+    :returns: command summary
+    :rtype: str
+    """
+
+    return command_documentation(command).split('\n')[0]
```
jsancio: We added `subcommand.default_config_schema`. We should do the same here and add `default_command_info` and `default_command_documentation`. I think the implementation should use `SubcommandMain` what do you think? This will keep things consistent.

dcos/subcommand.py
```diff
@@ -118,6 +124,30 @@ def distributions():
...
+    """documentation of default DCOS-CLI command
+    :param command: name of command
+    :param type: str
+    :returns: command summary
+    :rtype: str
+    """
+
+    command_module = "dcoscli.{}.main".format(command)
+    m = import_module(command_module)
+    return getattr(m, "_doc")().rstrip('\n')
```
jsancio: `_doc` means private and should not be access it outside the module.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
         package_json_path = os.path.join(self._dir(), 'package.json')
         with util.open_file(package_json_path) as package_json_file:
             return util.load_json(package_json_file)
+
+
+class SubcommandMain():
+    """Representes a subcommand running in the main thread"""
```
jsancio: I think you want to move this to the `__init__` method.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
...
+
+    def __init__(self, command):
+        """
+        :param commad: function to run in thread
+        :type command: func
```
jsancio: This looks like it is a `str`. It is only in `run_and_capture` that we look for the func.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
...
+
+    def run_and_capture(self):
+        """
+        Run a command and capture exceptions. This is a blocking call
+        :returns: tuple of exitcode, error (or None)
+        :rtype: int, str | None
+        """
+
+        command_module = "dcoscli.{}.main".format(self.command)
+        m = import_module(command_module)
```
jsancio: I am not super excited about this pattern. We already have a static list of default subcommands. Why not just turn that into a map/dictionary? We can use the map here to find the function. If the map doesn't have the key then we know that we need to exec a process.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
...
+            exit_code = m.main()
+        except Exception:
+            err = traceback.format_exc()
+            traceback.print_exc()
+            exit_code = 1
+        return exit_code, err
+
+
+class SubcommandProcess():
+    """Representes a subcommand running by a forked process"""
```
jsancio: I think you want to move this documentation to `__init__`.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
...
+class SubcommandMain():
+    """Representes a subcommand running in the main thread"""
+
+    def __init__(self, command):
+        """
+        :param commad: function to run in thread
+        :type command: func
+        """
+
+        self.command = command
```
jsancio: I would make this private.

dcos/subcommand.py
```diff
@@ -435,3 +465,77 @@ def package_json(self):
...
+        :type executable: executable
+        :param command: command to run by executable
+        :type command: str
+        :param args: arguments for command
+        :type args: [str]
+        """
+
+        self.executable = executable
+        self.command = command
+        self.args = args
```
jsancio: I would make all of these fields private.

dcos/subcommand.py
tamarrow: I'm not sure about adding these methods to `SubcommandMain'. That class just runs a main method in a thread, and doesn't know about the internals of the running method. It also takes command args as the constructor which wouldn't make sense outside that context.

jsancio: Ah Right. I see the problem. Maybe the default subcommand is a map of string to module so we can do something like `default_modules[command].help_doc()` or `default_modules[command].main(options)`.

tamarrow:
@jsancio updated

cli/dcoscli/analytics.py
```diff
@@ -236,13 +201,10 @@ def _command():
...
 
-    # avoid circular import
-    import dcoscli.main
-
-    args = docopt.docopt(dcoscli.main._doc(),
+    args = docopt.docopt(subcommand.default_doc("dcos"),
                          help=False,
                          options_first=True)
-    return args['<command>']
+    return args['<command>'] or ""
```
jsancio: How about `args.get('<command>', "")`?

cli/dcoscli/main.py
```diff
@@ -67,26 +63,28 @@ def _main():
...
-    :rtype: str
-    """
-    return pkg_resources.resource_string(
-        'dcoscli',
-        'data/help/dcos.txt').decode('utf-8')
+        if err:
+            analytics.track_err(
+                reporting_executor, exitcode, err, config, cluster_id)
+
+    return exitcode
```
jsancio: Move this to inside the `with`. Python is fancy but technically `exitcode` is only define inside the `with`.

dcos/subcommand.py
```diff
@@ -61,6 +66,72 @@ def get_package_commands(package_name):
...
+    from dcoscli.task import main as task_main
+
+    return {'config': config_main,
+            'help': help_main,
+            'marathon': marathon_main,
+            'node': node_main,
+            'package': package_main,
+            'service': service_main,
+            'task': task_main
+            }
```
jsancio: This is duplication that is going to be difficult to track down when we add more subcommands.

```
_default_modules = {
  'config': from dcoscli.config import main,
  'help': from dcoscli.help import main,
  ...
}

def default_subcommands():
   return _default_modules.keys()
```

You can remove `def default_modules()`. It is not used outside the module. Also, we need to move this code to the `dcoscli` package. The `dcoscli` package depends on `dcos`, not the other way around.

dcos/subcommand.py
```diff
@@ -61,6 +66,72 @@ def get_package_commands(package_name):
...
+    :param command: default DCOS-CLI command
+    :type command: str
+    :returns: config schema of command
+    :rtype: dict
+    """
+
+    resource = "data/config-schema/{}.json".format(command)
+    return json.loads(
+        pkg_resources.resource_string(
+            'dcoscli', resource).decode('utf-8'))
```
jsancio: Need to move this code to the `dcoscli` package. We have made this mistake before. How can we add tests to make sure that we don't do this?

dcos/subcommand.py
```diff
@@ -61,6 +66,72 @@ def get_package_commands(package_name):
...
+    :param command: default DCOS-CLI command
+    :type command: str
+    :returns: config schema of command
+    :rtype: dict
+    """
+
+    resource = "data/help/{}.txt".format(command)
+    return pkg_resources.resource_string(
+        'dcoscli',
+        resource).decode('utf-8')
```
jsancio: Need to move this code the `dcoscli` package.

dcos/subcommand.py
```diff
@@ -435,3 +521,77 @@ def package_json(self):
...
+        self._args = args
+
+    def run_and_capture(self):
+        """
+        Run a command and capture exceptions. This is a blocking call
+        :returns: tuple of exitcode, error (or None)
+        :rtype: int, str | None
+        """
+
+        m = default_modules().get(self._command)
```
jsancio: We don't check if `m is None`. In that case then just use `map[self._command]`.

jsancio:
LGTM @tamarrow. This PR is good to merge after you address my comments!

jsancio:
:balloon: :beers: :gem: :fireworks:

PR #547, opened 2016-03-30
add support for creating a dcos-cli binary on linux

tamarrow:


binary/Dockerfile.linux-binary
```diff
@@ -0,0 +1,27 @@
...
+  jq \
+  make \
+  openssh-client \
+  git \
+  sudo \
+&& sudo apt-get update --fix-missing \
+&& sudo apt-get install -y python-dev build-essential \
+&& sudo apt-get install -y python-pip python-virtualenv \
+&& pip install pip --upgrade \
+&& pip install pyinstaller oauth2client
```
jsancio: We can make this part of the `requirement.txt` and we get it for free when do you do `make env`, no?

jsancio: Note, that I think we have two `requirement.txt` files. We probably want to make this part of the one in `dcos-cli/cli`.

Makefile
```diff
@@ -14,3 +14,7 @@ doc: env
 
 packages: env
 	bin/packages.sh
+
+binary: clean env packages
+	rm -rf dist build
+	pyinstaller binary/binary.spec
```
jsancio: I think we should put this in `cli/Makefile`. Technically we are turn the `dcoscli` package into a native binary.

binary/Dockerfile.linux-binary
```diff
@@ -0,0 +1,27 @@
+# This image runs the dcos-cli test suite.
```
jsancio: This should go in `cli/binary/Dockerfile.linux-binary` see previous comment.

binary/binary.spec
```diff
@@ -0,0 +1,34 @@
+# -*- mode: python -*-
```
jsancio: Same here. We should move this to the `cli` folder (`dcoscli`).

cli/bin/test-binary.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
+
+BASEDIR=`dirname $0`/..
+
+cd $BASEDIR
+
+PATH=$BASEDIR/../dist:$PATH
+source env/bin/activate
+cp tests/data/dcos.toml $DCOS_CONFIG
+py.test tests/integrations
```
jsancio: I think you can avoid the activate and deactivate by doing `env/bin/py.test tests/integrations`.

cli/Makefile
```diff
@@ -9,6 +9,13 @@ env:
 test: env
 	bin/test.sh
 
+binary: clean env packages
+	rm -rf dist build
```
jsancio: Let's add this to the clean step/target. We can add it to the `bin/clean.sh` script.

cli/Makefile
```diff
@@ -9,6 +9,12 @@ env:
 test: env
 	bin/test.sh
 
+binary: clean env packages
+	pyinstaller binary/binary.spec
+
+test-binary: env
```
jsancio: Shouldn't we depend on the `binary` target instead of `env`?

tamarrow: yep, updated.

cli/requirements.txt
```diff
@@ -1,4 +1,6 @@
 sphinx>=1.3.1, <2.0
 tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
+pytest>=2.9.1
+oauth2client>=1.5.2
```
jsancio: Why do we need this here?

jsancio:
LGTM. One minor comment.

cli/requirements.txt
tamarrow: Was needed when we had the original oauth, but we don't need it now! Removed it from install.py also.

PR #554, opened 2016-04-05
don't print core.dcos_acs_token to stdout/stderr

tamarrow:


dcos/config.py
```diff
@@ -43,7 +43,16 @@ def set_val(name, value):
...
+    if name == "core.dcos_acs_token":
+        msg = "[{}]: ".format(name)
+        if not value_exists:
+            msg += "set"
+        elif old_value == new_value:
+            msg += "already set to that value"
+        else:
+            msg += "changed"
+        emitter.publish(msg)
+    elif not value_exists:
```
jsancio: I noticed that we were print this value even if I don't run the command `dcos config ...`. Can you verity that we don't print this to `stdout` as it will break scripts.

tamarrow: yes, we print it when we change the acs_token from a request to auth: https://github.com/mesosphere/dcos-cli/blob/master/dcos/http.py#L440

tamarrow: @jsancio updated to output these messages to stderr

tamarrow:
@jsancio bump

cli/tests/integrations/common.py
```diff
@@ -492,7 +492,6 @@ def config_set(key, value, env=None):
         env=env)
 
     assert returncode == 0
-    assert stderr == b''
```
jsancio: Should this be `assert stdout == b''`?

jsancio:
LGTM!

Issue #525, opened 2016-03-10
Provide more precise error when querying cosmos against a nonexistent cluster 


nickbp:
Host that isn't running DCOS:

$ dcos config set core.dcos_url http://example.com
[core.dcos_url]: changed from 'http://pool-cae8-elasticl-1op07zlyuov6s-277026749.us-west-2.elb.amazonaws.com/' to 'http://example.com'
$ dcos package repo list
This version of the dcos-cli is unsupported for your DCOS cluster. Please use a dcos-cli version < 0.4.0 or upgrade your cluster to DCOS version >= 1.6.1
$ dcos package repo add foo http://foo.com
This version of the dcos-cli is unsupported for your DCOS cluster. Please use a dcos-cli version < 0.4.0 or upgrade your cluster to DCOS version >= 1.6.1

Host that doesn't resolve:

$ dcos config set core.dcos_url http://onstuhoesntuhoesunthoeusntoeusntoe
[core.dcos_url]: set to 'http://onstuhoesntuhoesunthoeusntoeusntoe'
$ dcos package repo list
This version of the dcos-cli is unsupported for your DCOS cluster. Please use a dcos-cli version < 0.4.0 or upgrade your cluster to DCOS version >= 1.6.1
$ dcos package repo add foo http://foo.com
This version of the dcos-cli is unsupported for your DCOS cluster. Please use a dcos-cli version < 0.4.0 or upgrade your cluster to DCOS version >= 1.6.1

Could either:
- Have more precise error detection in these cases, providing a different message when they occur.
- Failing that, add a clause to existing error message saying e.g. "or set core.dcos_url to something else"

tamarrow:
At that point, we'd have to verify that the url is actually for a DCOS cluster.

PR #555, opened 2016-04-06
be more accurate when checking if cosmos enabled

tamarrow:
#525

### DIFF
--- a/dcos/cosmospackage.py
+++ b/dcos/cosmospackage.py
@@ -37,10 +37,14 @@ class Cosmos():
         # authorized for the command specified, not this endpoint
         except DCOSAuthorizationException:
             return True
-        # all other errors are treated as endpoint not available
+        # allow exception through so we can show user actual http exception
+        # except 404, because then the url is fine, just not cosmos enabled
+        except DCOSHTTPException as e:
+            logger.exception(e)
+            return e.status() != 404
         except Exception as e:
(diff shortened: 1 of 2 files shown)


dcos/cosmospackage.py
```diff
@@ -37,10 +37,14 @@ def enabled(self):
         # authorized for the command specified, not this endpoint
         except DCOSAuthorizationException:
             return True
-        # all other errors are treated as endpoint not available
+        # allow exception through so we can show user actual http exception
+        # except 404, because then the url is fine, just not cosmos enabled
         except Exception as e:
             logger.exception(e)
-            return False
+            if "HTTP 404: Not Found" in str(e):
```
jsancio: Okay. For my own edification, I take it that requests doesn't return a typed exception with the status code when it is 404. @tamarrow can we confirm?

tamarrow: nope: https://github.com/mesosphere/dcos-cli/blob/master/dcos/http.py#L222
screen shot 2016-04-05 at 5 58 46 pm

jsancio: We can fix that code, no? We have the response. We can add a field to DCOSHTTPException for the status.

tamarrow: yep! updated

jsancio:
LGTM.

PR #557, opened 2016-04-06
Segment bugfix

tamarrow:
If no command specified args has command key, but it equals None

### DIFF
--- a/cli/dcoscli/analytics.py
+++ b/cli/dcoscli/analytics.py
@@ -205,7 +205,7 @@ def _command():
     args = docopt.docopt(default_doc("dcos"),
                          help=False,
                          options_first=True)
-    return args.get('<command>', "")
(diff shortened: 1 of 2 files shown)


cli/dcoscli/analytics.py
```diff
@@ -205,7 +205,7 @@ def _command():
     args = docopt.docopt(default_doc("dcos"),
                          help=False,
                          options_first=True)
-    return args.get('<command>', "")
```
jsancio: Oh, this is because docopt adds an entry for it and sets it to `None`.

jsancio:
@tamarrow Please change the name of the PR to be more descriptive.

jsancio:
LGTM!

jsancio:
Can we add a tests for this?

tamarrow:
updated with test

PR #570, opened 2016-04-12
remove tracking (rollbar + segment.io)

tamarrow:
I'm leaving the `core.reporting` config value since we are going to be adding some tracking back in.


jsancio:
LGTM!

PR #571, opened 2016-04-12
clean up toml test files

tamarrow:
Instead of using different toml files in tests, just alter the config value for that test. This makes the tests more manageable especially since we now need to add auth credentials in the toml file.


jsancio:
LGTM

PR #575, opened 2016-04-13
Added functionality to 'dcos marathon task stop <task_id> --wipe' in …

meichstedt:
…order to wipe stateful services

Background:
I’ve started writing a deep dive on setting up [stateful services in marathon](https://github.com/dcos/dcos-docs/pull/66). One thing is wiping a task from state including reservations and persistent volumes. That’s possible in marathon via a POST to `/v2/tasks/delete` (yes, POST to /delete, don’t ask) as described here https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-tasks-delete – for the sake of having the deep dive tutorial merely on dcos cli without any direct marathon api interaction, I’ve added the tasks delete command. It doesn’t cover all possibilities of the marathon api, but it would be good to be able to wipe tasks via the cli, I guess.

### DIFF
--- a/cli/dcoscli/marathon/main.py
+++ b/cli/dcoscli/marathon/main.py
@@ -72,6 +72,11 @@ def _cmds():
             function=_task_list),
 
         cmds.Command(
+            hierarchy=['marathon', 'task', 'stop'],
+            arg_keys=['<task-id>', '--wipe'],
+            function=_task_stop),
+
+        cmds.Command(
             hierarchy=['marathon', 'task', 'show'],
             arg_keys=['<task-id>'],
             function=_task_show),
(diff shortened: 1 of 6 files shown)


cli/dcoscli/marathon/main.py
```diff
@@ -781,6 +786,27 @@ def _task_list(app_id, json_):
...
+    :rtype: int
+    """
+
+    client = marathon.create_client()
+    task = client.stop_task(task_id, wipe)
+
+    if task is None:
+        raise DCOSException("Task '{}' does not exist".format(task_id))
+
+    emitter.publish(task)
```
tamarrow: can you add a test for this? What does it return when successful?

dcos/marathon.py
```diff
@@ -669,6 +669,37 @@ def get_task(self, task_id):
...
+        :param wipe: whether remove reservations and persistent volumes.
+        :type wipe: bool
+        :returns: a tasks
+        :rtype: dict
+        """
+
+        if not wipe:
+            params = None
+        else:
+            params = {'wipe': 'true'}
```
tamarrow: should we have the user confirm (with Y/n) before we wipe their data?

cli/dcoscli/marathon/main.py
meichstedt: Sure, I've added happy path tests for `stop` and `stop --wipe` as well as a test for each case that verifies failure when sending an unknown task_id.

The return value on the happy path is a list of the stopped tasks.

dcos/marathon.py
meichstedt: I don't think so. the default behavior of `delete` / `stop` simply kills the task. Adding `--wipe` already is the confirmation that the task should be wiped off the state. (We also don't ask to confirm a `dcos marathon app remove`, in which case the task will also be wiped.)

tamarrow: :+1:

tamarrow:
LGTM. Thanks for adding this :)

Issue #567, opened 2016-04-11
Mechanism for getting the version of DC/OS


jsancio:
Please answer the following questions before submitting your issue. Thanks!

### What version of the DCOS CLI are you using (`dcos --version`)?
SNAPSHOT

### What version of DCOS are you using?
1.6.1

### What operating system and version are you using?
Fedora 22

### What did you do?
If possible please provide a recipe for reproducing.
I am trying to get the version of DC/OS

### What did you expect to see?
`dcos --version` should display the CLI version and the DC/OS version. I don't know if this is the right solution but the user should be able to get the version of DC/OS. This helps to debug issues.

### What did you see instead?
`dcos --version` only shows the version of the CLI

tamarrow:
We could output {DCOS_VERSION: 1-7.dev, DCOS_CLI_VERSION: 0.4.2}.
We can get the dcos_version from `dcos-metadata/dcos-version.json`

jsancio:
Sounds good to me.

PR #577, opened 2016-04-15
add DCOS version to `dcos --version`

tamarrow:
#567 
screen shot 2016-04-15 at 2 42 59 pm

cc @jsancio

### DIFF
--- a/cli/dcoscli/main.py
+++ b/cli/dcoscli/main.py
@@ -8,6 +8,8 @@ from dcos import constants, emitting, errors, http, subcommand, util
 from dcos.errors import DCOSException
 from dcoscli.subcommand import SubcommandMain, default_doc
 
+from six.moves import urllib
+
 logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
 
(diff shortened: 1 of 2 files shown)


cli/dcoscli/main.py
```diff
@@ -20,12 +23,36 @@ def main():
...
+    :param config: config
+    :type config: Toml
+    :returns: versions of DCOS and DCOS-CLI
+    :rtype: dict
+    """
+
+    dcos_url = config.get("core.dcos_url")
+    dcos_version = "N/A"
+    if dcos_url:
+        url = urllib.parse.urljoin(dcos_url, 'dcos-metadata/dcos-version.json')
```
jsancio: Why not use `dcos.http`?

cli/dcoscli/main.py
```diff
@@ -20,12 +23,36 @@ def main():
         return 1
 
 
+def _get_versions(config):
```
jsancio: Why not pass `dcos_url`? It is the only thing that we use from the config file.

jsancio:
@tamarrow Can you include the sample output of `dcos --version` in the description for this PR?

cli/dcoscli/main.py
```diff
@@ -20,12 +23,36 @@ def main():
...
+    :returns: versions of DCOS and DCOS-CLI
+    :rtype: dict
+    """
+
+    dcos_url = config.get("core.dcos_url")
+    dcos_version = "N/A"
+    if dcos_url:
+        url = urllib.parse.urljoin(dcos_url, 'dcos-metadata/dcos-version.json')
+        try:
+            res = http.get(url)
```
tamarrow: @jsancio using `dcos.http` library. ^above is just to create the url

jsancio:
@tamarrow thanks for the output. Why don't we turn this until more human friendly output like:

```
dcos --version
dcos-cli/0.4.3 dcos/1.7.0
```

I am worry that `dcos/1.7.0` may not be enough information when running against `testing/continuous` builds. Is there a way to get and display the git commit sha in addition to the version? cc @cmaloney

tamarrow:
I wanted to make it scriptable, but since its mostly used for debugging, a string is fine. I can also add the dcos-image commit, which is accessable by that same endpoint.
So something like: `dcos-cli/0.4.3 dcos/1.7.0/dcos-image/hash-here`
cc @jsancio

cmaloney:
I'd really rather not have another format for versions of the dcos build... Could we just output "dcos/version/<version>"
"dcos/commit/<dcos-image-commit>"
"dcos/bootstrap-id/<bootstrap-id>"
all as three different things?

tamarrow:
@jsancio updated with `dcos config show` style formatting: 
screen shot 2016-04-18 at 4 51 23 pm

jsancio:
LGTM. :shipit:

cli/dcoscli/main.py
```diff
@@ -20,12 +22,44 @@ def main():
...
+        except Exception:
+            pass
+
+    return (
+        "dcoscli.version={}\n".format(dcoscli.version) +
+        "dcos.version={}\n".format(dcos_info.get("version", "N/A")) +
+        "dcos.commit={}\n".format(dcos_info.get(
+            "dcos-image-commit", "N/A")) +
+        "dcos.bootstrap-id={}".format(dcos_info.get("bootstrap-id", "N/A"))
+    )
```
jsancio: Okay for now. Let's file an issue to use the same code we wrote for `dcos config show`.

PR #586, opened 2016-04-25
Install binary subcommand

tamarrow:


dcos/subcommand.py
```diff
@@ -219,8 +244,32 @@ def _write_package_json(pkg):
...
+    """
+    cli = resources.get("cli")
+    if cli and cli.get("binaries"):
+        arch = platform.architecture()[0]
+        if arch != "64bit":
+            logger.debug("Unsupported archecture {}\n", arch)
+            return None
+        system = platform.system().lower()
+        if system == "darwin":
+            system = "osx"
```
jsancio: We could also just call it `darwin` in the package spec. Maybe that is the correct platform name.

dcos/subcommand.py
```diff
@@ -231,19 +280,32 @@ def _install_env(pkg, options):
...
+    if binary_cli:
+        _install_with_binary(
             pkg.name(),
             env_dir,
-            install_operation['pip'])
+            binary_cli)
     else:
-        raise DCOSException("Installation methods '{}' not supported".format(
-            install_operation.keys()))
+        install_operation = pkg.command_json(options)
```
jsancio: I think we changed the spec so that `command.json` is not a mustache template and instead is just a regular JSON object. We need to do this so that we can reliably install subcommands after the UI has installed an package application. This means that we should change the signature and implementation for `pkg.command_json` so that it doesn't take an option.

dcos/subcommand.py
```diff
@@ -320,6 +382,46 @@ def _find_virtualenv(bin_directory):
...
+    :type env_directory: str
+    :param binary_cli: binary cli to install
+    :type binary_cli: str
+    :rtype: None
+    """
+
+    # Do not replace util.temptext NamedTemporaryFile
+    # otherwise bad things will happen on Windows
+    with util.temptext() as text_file:
+        fd, requirement_path = text_file
```
jsancio: It doesn't look like we are using `text_file`, `fd` or `requirement_path`. Coping and pasting code is evil. :)

dcos/subcommand.py
```diff
@@ -320,6 +382,46 @@ def _find_virtualenv(bin_directory):
...
+
+        try:
```
jsancio: Ignoring this part as I think you are going to change it to take into account both `executable` and `zip` files.

dcos/subcommand.py
tamarrow: I'll fix this in another pr. Created https://github.com/dcos/dcos-cli/issues/593

jsancio: Thanks!

On Wed, Apr 27, 2016 at 12:39 PM tamarrow <notifications@github.com> wrote:

> In dcos/subcommand.py
> <https://github.com/dcos/dcos-cli/pull/586#discussion_r61322519>:
>
> >      else:
> > -        raise DCOSException("Installation methods '{}' not supported".format(
> > -            install_operation.keys()))
> > +        install_operation = pkg.command_json(options)
>
> I'll fix this in another pr. Created #593
> <https://github.com/dcos/dcos-cli/issues/593>
>
> —
> You are receiving this because you commented.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/dcos/dcos-cli/pull/586/files/b49c3487208434d7e02ce57509db71ad5c624347#r61322519>
>
-- 
José Armando García Sancio
Mesosphere

dcos/subcommand.py
```diff
@@ -219,8 +247,30 @@ def _write_package_json(pkg):
...
+    :type resources: {}
+    :returns: cli binary
+    :rtype: str | None
+    """
+    cli = resources.get("cli")
+    if cli and cli.get("binaries"):
+        arch = platform.architecture()[0]
+        if arch != "64bit":
+            logger.debug("Unsupported archecture {}\n", arch)
+            return None
```
jsancio: What do you think about just throwing a DCOSException with the message above?

jsancio: I think I understand. We want to support both new and old for a now.

dcos/subcommand.py
```diff
@@ -219,8 +247,30 @@ def _write_package_json(pkg):
...
+        binary = cli["binaries"].get(system)
+        if binary:
+            return binary.get("x86-64")
+
+    return None
+
+
 def _install_env(pkg, options):
-    """ Install subcommand virtual env.
+    """Install subcommand virtual env.
```
jsancio: This comment doesn't look correct. This installs both binary and virtualenv, correct?

dcos/subcommand.py
```diff
@@ -230,20 +280,31 @@ def _install_env(pkg, options):
...
+            binary_cli)
     else:
-        raise DCOSException("Installation methods '{}' not supported".format(
-            install_operation.keys()))
+        install_operation = pkg.command_json(options)
+
+        virtualenv_dir = os.path.join(
+            pkg_dir,
+            constants.DCOS_SUBCOMMAND_VIRTUALENV_SUBDIR
+        )
```
jsancio: We can move this inside the block for `if 'pip' in install_operations:`.

dcos/subcommand.py
```diff
@@ -324,6 +385,64 @@ def _find_virtualenv(bin_directory):
...
+    """
+    :param package_name: the name of the package
+    :type package_name: str
+    :param package_dir: the path to the directory in which to install the
+                          package's binary cli
+    :type package_dir: str
+    :param binary_cli: binary cli to install
+    :type binary_cli: str
+    :rtype: None
+    """
```
jsancio: We need to make sure that we handle as many failure conditions as possible. We can use `tempfile` to help us with this. I'll highlight the operations that I know can fail.

dcos/subcommand.py
```diff
@@ -324,6 +385,64 @@ def _find_virtualenv(bin_directory):
...
+    binary_url, kind = binary_cli["url"], binary_cli["kind"]
+
+    try:
+        bin_dir = os.path.join(package_dir, os.path.join(BIN_DIRECTORY))
+        binary_name = "dcos-{}".format(package_name)
+        if kind == "executable":
+            if not os.path.exists(bin_dir):
+                os.makedirs(bin_dir)
+            binary_file = os.path.join(bin_dir, binary_name)
+            urllib.request.urlretrieve(binary_url, binary_file)
```
jsancio: This can fail.

dcos/subcommand.py
```diff
@@ -324,6 +385,64 @@ def _find_virtualenv(bin_directory):
...
+        elif kind == "zip":
+            contents = BytesIO(requests.get(binary_url, stream=True).content)
```
jsancio: This can fail and we shouldn't store it in memory.

dcos/subcommand.py
```diff
@@ -324,6 +385,64 @@ def _find_virtualenv(bin_directory):
...
+    :param binary_cli: binary cli to install
+    :type binary_cli: str
+    :rtype: None
+    """
+
+    binary_url, kind = binary_cli["url"], binary_cli["kind"]
+
+    try:
+        bin_dir = os.path.join(package_dir, os.path.join(BIN_DIRECTORY))
+        binary_name = "dcos-{}".format(package_name)
```
jsancio: Are we only going to allow one binary in the `zip` case?

dcos/subcommand.py
```diff
@@ -324,6 +385,64 @@ def _find_virtualenv(bin_directory):
...
+        # make binary executable
+        st = os.stat(binary_file)
+        os.chmod(binary_file, st.st_mode | stat.S_IEXEC)
+    except DCOSException:
+        shutil.rmtree(package_dir)
+        raise
+    except Exception as e:
+        logger.exception(e)
+        shutil.rmtree(package_dir)
+        raise _generic_error(package_name)
```
jsancio: Can we do the two `shutil.rmtree` into a `contextmanager`?

dcos/subcommand.py
```diff
@@ -121,6 +145,10 @@ def distributions():
             if os.path.isdir(
                 os.path.join(
                     subcommand_dir,
+                    subdir)) or
+            os.path.isdir(
+                os.path.join(
+                    subcommand_dir,
                     subdir,
                     constants.DCOS_SUBCOMMAND_VIRTUALENV_SUBDIR))
```
jsancio: Should we do the same thing for binary CLIs? I think we added the `env` folder so that we can store `package.json` without overriding anything that `virtualenv` does. I think we need to do the same for the `zip` case. That would simplify this code and the code that finds the executable.

jsancio: We can change the constant to `DCOS_SUBCOMMAND_ENV_SUBDIR`.

dcos/subcommand.py
tamarrow: Yes. If they don't have a binary that matches the platform, we can try pypi for now.

dcos/subcommand.py
tamarrow: yep, updated.

dcos/subcommand.py
tamarrow: done

dcos/subcommand.py
tamarrow: I like it, updating.

dcos/subcommand.py
tamarrow: no

tamarrow:
@jsancio updated

dcos/subcommand.py
```diff
@@ -219,8 +226,74 @@ def _write_package_json(pkg):
...
+        if expected_value != _hashfile(filename):
+            raise DCOSException(
+                "Downloaded CLI contents do not match expected contents. "
+                "Aborting...")
+        else:
+            return
+    else:
+        raise DCOSException(
+            "Hash algorithm specified is unsupported. "
+            "Please contact the project maintainer")
```
jsancio: Replace project maintainer with package maintainer. You use package maintainer in other places.

dcos/subcommand.py
```diff
@@ -219,8 +226,74 @@ def _write_package_json(pkg):
...
+    return hasher.hexdigest()
+
+
+def _check_hash(filename, content_hashes):
+    """Validates whether downloaded binary matches expected hash
+
+    :param filename: path to binary
+    :type filename: str
+    :param content_hashes: list of hash algorithms/value
+    :type content_hashes: [{}]
```
jsancio: Let's change this to `[{"algo": <string>, "value": <string>}]`. What do you think?

dcos/subcommand.py
```diff
@@ -219,8 +226,74 @@ def _write_package_json(pkg):
...
+
+    content_hash = next((contents for contents in content_hashes
+                        if contents.get("algo") == "sha256"),
+                        None)
+    if content_hash:
+        expected_value = content_hash.get("value")
+        if expected_value != _hashfile(filename):
+            raise DCOSException(
+                "Downloaded CLI contents do not match expected contents. "
+                "Aborting...")
```
jsancio: Let's print the two hashes. E.g. "The hash for the downloaded command (###) does not match the expected value (###). Aborting..."

dcos/subcommand.py
```diff
@@ -230,20 +303,31 @@ def _install_env(pkg, options):
...
-
-    if 'pip' in install_operation:
-        _install_with_pip(
+    if binary_cli:
+        _install_with_binary(
             pkg.name(),
-            env_dir,
-            install_operation['pip'])
+            pkg_dir,
+            binary_cli)
```
jsancio: I haven't looked at the implementation but why does this take `pkg_dir` and `_install_with_pip` take `env_dir`?

dcos/subcommand.py
```diff
@@ -261,7 +345,7 @@ def install(pkg, options):
 
     _write_package_json(pkg)
 
-    _install_env(pkg, options)
+    _install_cli(pkg, options)
```
jsancio: We don't use `options` let's remove it from `install` and `_install_cli`.

dcos/subcommand.py
```diff
@@ -230,20 +303,31 @@ def _install_env(pkg, options):
...
+        _install_with_binary(
             pkg.name(),
-            env_dir,
-            install_operation['pip'])
+            pkg_dir,
+            binary_cli)
     else:
-        raise DCOSException("Installation methods '{}' not supported".format(
-            install_operation.keys()))
+        install_operation = pkg.command_json()
```
jsancio: `command.json` is optional. So this will fail when `resource.cli` is set but `command.json` is missing. We currently use `has_command_definition` to determine if there is a CLI to install. We need to rework that. One option is to change that method's name to `has_cli_definition` and fix the implementation to look at both `resource.cli` and `command.json`. Even that is not 100% correct. I think the only reason we need that check is so that we can print `Installing CLI subcommand....`. Think about it and let me know how you want to solve this.

dcos/subcommand.py
```diff
@@ -324,6 +408,99 @@ def _find_virtualenv(bin_directory):
...
+    :param binary_cli: binary cli to install
+    :type binary_cli: str
+    :rtype: None
+    """
+
+    binary_url, kind = binary_cli.get("url"), binary_cli.get("kind")
+
+    try:
+        env_dir = os.path.join(package_dir,
+                               constants.DCOS_SUBCOMMAND_ENV_SUBDIR)
```
jsancio: We should just pass `env_dir` into this function and make it the same as the `pip` implementation.

dcos/subcommand.py
```diff
@@ -324,6 +408,99 @@ def _find_virtualenv(bin_directory):
...
+    :param dst: destination file or directory
+    :type dst: str
+    :returns: None
+    """
+
+    try:
+        if os.path.isfile(src):
+            util.sh_copy(src, dst)
+        else:
+            shutil.copytree(src, dst)
```
jsancio: We should use `shutil.move`. It will figure out if a filesystem move is possible and use that else it will perform a copy and remove. https://docs.python.org/2/library/shutil.html#shutil.move

dcos/subcommand.py
tamarrow: done

dcos/subcommand.py
tamarrow: sure

dcos/subcommand.py
tamarrow: done

dcos/subcommand.py
```diff
@@ -219,8 +226,74 @@ def _write_package_json(pkg):
...
+                "Aborting...")
+        else:
+            return
+    else:
+        raise DCOSException(
+            "Hash algorithm specified is unsupported. "
+            "Please contact the project maintainer")
+
+
+def _get_cli_binary(resources):
```
tamarrow: it returns `{"url": <str>, "kind": <str>, "contentHash": [{}]}`, so updated pydoc and renamed to `_get_cli_binary_info`

dcos/subcommand.py
tamarrow: updated

dcos/subcommand.py
tamarrow: Yes, we should change the method signature to `has_cli_definition` and can check for `None` in `pkg.command_json` to see if that file exists. This won't catch the case where we only have the cli defined in `resource.cli` but not for the user's platform. In that case we will print `Installing cli subcommand...` and then `Could not find a CLI subcommand for your platform` which seems appropriate to me.

dcos/subcommand.py
tamarrow: 👍

dcos/subcommand.py
jsancio: Depending on the answer we get from product, I think we should throw an `DCOSException` if `cli` is set but the current platform is missing.

dcos/subcommand.py
```diff
@@ -324,6 +405,95 @@ def _find_virtualenv(bin_directory):
...
+    content_hashes = binary_cli.get("contentHash")
+
+    try:
+        env_bin_dir = os.path.join(env_directory, BIN_DIRECTORY)
+
+        if kind == "executable":
+            binary_name = "dcos-{}".format(package_name)
+            with util.temptext() as file_tmp:
+                _, binary_tmp = file_tmp
+                urllib.request.urlretrieve(binary_url, binary_tmp)
```
jsancio: Did you look into using `requests` or even better our `dcos.http`? Is there are reason why you are using `urllib.request`?

dcos/subcommand.py
```diff
@@ -324,6 +405,95 @@ def _find_virtualenv(bin_directory):
...
+    :param dst: destination file or directory
+    :type dst: str
+    :returns: None
+    """
+
+    try:
+        shutil.move(src, dst)
+    except Exception as e:
+        logger.exception(e)
+        shutil.rmtree(dst)
```
jsancio: I don't think this is enough, no? I think we want a `contextmanager` for `pkg_dir` defined in line 305. We want to delete the `pkg_dir` if we had any failure during the installation, no?

tamarrow: done! util.path

dcos/subcommand.py
```diff
@@ -1,12 +1,17 @@
...
 import os
+import platform
 import shutil
+import stat
 import subprocess
 import sys
+import zipfile
 from subprocess import PIPE, Popen
 
+import dcos.http
```
tamarrow: using this import instead of `from dcos import http` to avoid circular import

dcos/subcommand.py
tamarrow: It was simpler, but updated to use `dcos.http`

dcos/subcommand.py
jsancio: What cycle is this? we have `from dcos import ...` below.

tamarrow: `http` imports `config` which imports `subcommand`. http://stackoverflow.com/questions/744373/circular-or-cyclic-imports-in-python

dcos/subcommand.py
```diff
@@ -219,40 +225,125 @@ def _write_package_json(pkg):
...
+def _get_cli_binary_info(cli_resources):
+    """Find compatible cli binary, if one exists
 
+    :param cli_resources: cli property of resource.json
+    :type resources: {}
+    :returns: {"url": <str>, "kind": <str>, "contentHash": [{}]}
+    :rtype: {} | None
+    """
+
+    if cli_resources.get("binaries"):
```
jsancio: `if 'binaries' in cli_resources:`

dcos/subcommand.py
```diff
@@ -219,40 +225,125 @@ def _write_package_json(pkg):
...
+            raise DCOSException(
+                "There is no compatible subcommand for your architecture [{}] "
+                "We only support x86-64. Aborting...".format(arch))
+        system = platform.system().lower()
+        binary = binaries.get(system)
+        if binary is None:
+            raise DCOSException(
+                "There is not compatible subcommand for your system [{}] "
+                "Aborting...".format(binary))
+        elif binary.get("x86-64") is not None:
```
jsancio: `elif 'x86-64' in binary`

dcos/subcommand.py
```diff
@@ -219,40 +225,125 @@ def _write_package_json(pkg):
...
+    :type pkg: PackageVersion
+    :rtype: None
+    """
+
+    pkg_dir = _package_dir(pkg.name())
+    with util.path(pkg_dir):
+        env_dir = os.path.join(pkg_dir, constants.DCOS_SUBCOMMAND_ENV_SUBDIR)
+
+        resources = pkg.resource_json()
+        binary = resources and resources.get("cli")
```
jsancio: I don't understand this code:
```
>>> resource = {}
>>> value = resource and resource.get('blah')
>>> value
{}
```

dcos/subcommand.py
jsancio: Which should remove the cycle. Please file a high priority issue and we can fix it right after this PR

tamarrow: It's actually causing other import issues. Filed #602, and will use `requests` until we fix.

dcos/subcommand.py
tamarrow: updated

dcos/subcommand.py
tamarrow: updated

dcos/subcommand.py
tamarrow: simplified this.

dcos/util.py
```diff
@@ -77,6 +77,21 @@ def temptext():
         shutil.rmtree(path, ignore_errors=True)
 
 
+@contextlib.contextmanager
+def path(path):
```
jsancio: Let's name this something else. Maybe `remove_path_on_error` so you can do `with remove_path_on_error(path) as path:`

dcos/util.py
```diff
@@ -77,6 +77,21 @@ def temptext():
...
+    """A context manager for modifying a specific path
+    `path` and all subpaths will be removed on error
+
+    :rtype: None
+    """
+
+    try:
+        yield
```
jsancio: `yield path` so you can use it as above.

dcos/subcommand.py
```diff
@@ -219,40 +225,125 @@ def _write_package_json(pkg):
...
+    """Install subcommand cli
 
-def install(pkg, options):
+    :param pkg: the package to install
+    :type pkg: PackageVersion
+    :rtype: None
+    """
+
+    pkg_dir = _package_dir(pkg.name())
+    with util.path(pkg_dir):
```
jsancio: If you implement the previous comment, you can `with util.remove_path_on_error(_package_dir(pkg.name())) as pkg_dir:`

dcos/subcommand.py
jsancio: Nice. We can probably just remove this method `_move` and use `shutil.move` directly, no?

jsancio:
LGTM @tamarrow !! :confetti_ball:

dcos/util.py
tamarrow: 👍

dcos/subcommand.py
tamarrow: yep!

jsancio:
@tamarrow I just realized that there aren't any tests for this PR. We should at least have tests for a working "executable" subcommand, a working "zip" subcommand and a invalid "zip" subcommand.

PR #587, opened 2016-04-25
Make pagination a toggle

lloesche:
The current dcos-cli behaviour of paginating all output is uncommon for a Unix CLI tool and something you usually only see for commands that output entire man pages.

This PR allows the user to disable pagination.

Default is pagination=True to not break current behaviour.


tamarrow:
@lloesche looks good, can you add tests for this please?

lloesche:
@tamarrow I was wondering about that. How could I test if the pagination toggle works? When run in a CI server it's never going to have a tty, does it?
Do you have any good ideas?

tamarrow:
How about using a mock? You can even just verify if `pydoc.pipepager` is called or not. see [here](https://github.com/dcos/dcos-cli/blob/master/cli/tests/unit/common.py)

tamarrow:
^ @lloesche

tamarrow:
taking over this pr #622

PR #589, opened 2016-04-25
remove CI test helper script

tamarrow:
In order to avoid exposing secret token, run commands in CI directly.


BenWhitehead:
:ship:

PR #591, opened 2016-04-27
intercept SSL errors and add helpful message to users

tamarrow:


dcos/config.py
```diff
@@ -213,6 +213,26 @@ def get_config_schema(command):
     return subcommand.config_schema(executable, command)
 
 
+def get_property_description(name, value):
+    """
+    :param name: name of paramater
+    :type name: str
```
jsancio: `name` is the config section or subcommand. We call it section in the rest of the code. `param` is the property within the section. We call it subkey in the rest of the code. Take a look at line 27 in this file.

dcos/config.py
```diff
@@ -213,6 +213,26 @@ def get_config_schema(command):
...
+    :type param: str
+    :returns: description of name.value, or title if no description
+    :rtype: str
+    """
+
+    schema = get_config_schema(name)
+    properties = schema.get("properties")
+    config = properties.get(value)
+    if config is not None:
+        return config.get("description", "title")
```
jsancio: Is this correct? Isn't `config` a regular Python `dict`? If so then this will try to get `description` or return string "title" if it can't find it.

dcos/config.py
```diff
@@ -213,6 +213,26 @@ def get_config_schema(command):
     return subcommand.config_schema(executable, command)
 
 
+def get_property_description(name, value):
```
jsancio: The rest of the public APIs takes a `name` that is a dot separated path. E.g. `unset` and `set_value`. Would be nice to keep that consistent. For example we can call `split_key` from here.

jsancio:
Looks good @tamarrow. Exciting!

dcos/config.py
tamarrow: No, it's the json-schema dict, so the properties are inside of 'properties'

dcos/config.py
tamarrow: updated!

dcos/config.py
```diff
@@ -213,6 +213,25 @@ def get_config_schema(command):
     return subcommand.config_schema(executable, command)
 
 
+def get_property_description(section, subkey):
+    """
+    :param section: section of config paramater
+    :type section: str
+    :param subkey: property within 'section'
+    :type param: str
```
jsancio: should be `:param subkey:`.

dcos/config.py
jsancio: Right but `get('key', 'default value')` on a `dict` tries to find the key 'key'. If it can't find it, it returns 'title', no?

dcos/config.py
```diff
@@ -213,6 +213,25 @@ def get_config_schema(command):
...
+    :returns: description of section.subkey, or title if no description
+    :rtype: str
+    """
+
+    schema = get_config_schema(section)
+    property_info = schema["properties"].get(subkey)
+    if property_info is not None:
+        return property_info.get("description", "title")
```
jsancio: Copy and paste of a previous comment:

> Right but get('key', 'default value') on a dict tries to find the key 'key'. If it can't find it, it returns 'title', no?

dcos/config.py
tamarrow: `:type subkey:`. updated

dcos/config.py
tamarrow: sorry, missed this comment. yes, updated

dcos/config.py
```diff
@@ -213,6 +213,26 @@ def get_config_schema(command):
...
+    :type subkey: str
+    :returns: description of section.subkey, or title if no description
+    :rtype: str
+    """
+
+    schema = get_config_schema(section)
+    property_info = schema["properties"].get(subkey)
+    if property_info is not None:
+        return property_info.get(
+            "description", property_info.get("title", "value"))
```
jsancio: Why don't we just return `None`? We can then check `None` in `dcos.http` so that we only printed if the value is found. What do you think?

```
description = config.get_property_description("core", "ssl_verify")
logger.exception("HTTP SSL Error")
msg = ("An SSL error occurred. To configure your SSL settings, "
            "please run: `dcos config set core.ssl_verify <value>`"
            "\n<value>: {}".format(description) if description else "")
raise DCOSException(msg)
```

tamarrow: I like it, updated!

jsancio:
LGTM!

PR #597, opened 2016-04-28
limit length of description in search table

tamarrow:
PrettyTable renders each row to be the width of the largest row.
Packages with long descriptions make the table very hard to read.


tamarrow:
before: 
screen shot 2016-04-28 at 4 51 03 pm
after: 
screen shot 2016-04-28 at 4 48 55 pm

jsancio:
LGTM! I noticed that your example is not sorting by `SELECTED`. I suspect that we have code in the CLI that sorts it by `NAME`. I think we should let Cosmos do the sorting. If Cosmos is not sorting by `SELECTED` and then `NAME`, let's file a bug against it.

PR #601, opened 2016-05-04
Make dcos task ls work for completed tasks

surdy:
This is for https://dcosjira.atlassian.net/browse/DCOS-81

I have tested for 2 cases and got expected results: 
1) If the task is completed and has entry on both master and slave --> ls successfully lists file
2) If the task is listed on master but the sandbox is deletec from the slave --> Cannot find a task with ID containing "driver-20160504203014-0002"

@tamarrow @jsancio I'm sure there would be cases I haves missed. Let me know if any comes to your mind and I will test for those.

### DIFF
--- a/cli/dcoscli/data/help/task.txt
+++ b/cli/dcoscli/data/help/task.txt
@@ -5,7 +5,7 @@ Usage:
     dcos task --info
     dcos task [--completed --json <task>]
     dcos task log [--completed --follow --lines=N] <task> [<file>]
-    dcos task ls [--long] <task> [<path>]
+    dcos task ls [--long --completed] <task> [<path>]
 
 Command:
     log
--- a/cli/dcoscli/task/main.py
+++ b/cli/dcoscli/task/main.py
@@ -50,7 +50,7 @@ def _cmds():
 
         cmds.Command(
             hierarchy=['task', 'ls'],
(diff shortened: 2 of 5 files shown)


surdy:
This is unexpected. `make test` had succeeded on my machine. I'll dig in to the error

tamarrow:
@surdy should we add this as `--completed` flag to be analogous to `dcos task` and `dcos task log`? Can you add tests for the two cases you specified above?

surdy:
I thought about adding `--completed` to `dcos task ls` but to me it seemed redundant. Because when you are doing a `dcos task ls` you are already are specifying the task ID (which you possibly got using `dcos task --completed`) 

So my point being when you are doing `dcos task ls` you already have the task id and do not really care if the task is completed or not. If you are looking for files in a task you want them without having to say `--completed`

I have not used these commands extensively but that's my thought on it.  I will work on adding the tests for it and update this.

surdy:
I'm sorry for the delay on this. I got tied up in some other pressing issue. I have added `--completed` flag to `task ls`. Please disreagrd my previous comment. There were cases I was missing in argument there.

Please do not merge yet. I still have to add a couple of tests for this.

surdy:
Closing this. Will merge into #601

PR #614, opened 2016-05-17
Make --completed great again

surdy:
As per documentation `dcos task --completed` was intended to show BOTH completed and running tasks. In practice it only shows completed tasks.

### DIFF
--- a/cli/tests/integrations/test_task.py
+++ b/cli/tests/integrations/test_task.py
@@ -77,13 +77,13 @@ def test_task_table():
 
 def test_task_completed():
     returncode, stdout, stderr = exec_command(
-        ['dcos', 'task', '--completed', '--json'])
+        ['dcos', 'task', '--completed', '--json', '*-app*'])
     assert returncode == 0
     assert stderr == b''
     assert len(json.loads(stdout.decode('utf-8'))) > NUM_TASKS
 
     returncode, stdout, stderr = exec_command(
-        ['dcos', 'task', '--json'])
+        ['dcos', 'task', '--json', '*-app*'])
     assert returncode == 0
     assert stderr == b''
     assert len(json.loads(stdout.decode('utf-8'))) == NUM_TASKS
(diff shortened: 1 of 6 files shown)


tamarrow:
thanks for fixing @surdy!

surdy:
@tamarrow I have merged #601 into this locally and am testing it. Once my local tests succeed I will update this PR.

surdy:
I have modified the test for completed task to be more resilient. Please do review the changes in test_task.py

cli/tests/integrations/test_task.py
```diff
@@ -25,12 +25,14 @@
...
 NUM_TASKS = len(INIT_APPS)
+TASK_ID_COMPLETED = ''
 
 
 def setup_module():
     # create a completed task
     with app(SLEEP_COMPLETED, 'test-app-completed'):
-        pass
+        global TASK_ID_COMPLETED
+        TASK_ID_COMPLETED = _get_completed_task_id()
```
tamarrow: You only use `TASK_ID_COMPLETED` in `test_ls_completed`. Can we just move `_get_completed_task_id` in that test?

surdy: How would I get the `TASK_ID_COMPLETED` outside of `setup_module` ? Only point in the code where its is running is in `setup_module` right ? By TASK_ID_COMPLETED,  I mean to get the full ID of the task including hash. 
I did this for `test_ls_completed` not to fail. It used to succeed in p27 tests and fail in p34 tests because at that point there would be multiple `test-app-completed` and I wanted a way to get just one of those.

surdy:
I updated the tests as per your comments @tamarrow. Have a look when you get a chance

cli/tests/integrations/test_task.py
```diff
@@ -257,6 +257,30 @@ def test_ls_bad_path():
         returncode=1)
 
 
+def test_ls_completed():
+    # create a completed task
+    with app(SLEEP_COMPLETED, 'test-app-completed'):
+        # get its task id
+        task_id_completed = _get_completed_task_id()
+
```
tamarrow: @surdy since we are creating `test-app-completed` here, can we remove it from the setup module?

surdy: It would lead to other tests (`*_completed`) that rely upon that to fail. Unless you move this test before the other tests. I do not like the idea of tests being order sensitive

tamarrow: ok, than we should change the name of this app so we don't create two apps with the same name which could cause unexpected behavior.

Issue #603, opened 2016-05-09
package describe --cli should look at both command.json and resource.json


tamarrow:
Please answer the following questions before submitting your issue. Thanks!

### What version of the DCOS CLI are you using (`dcos --version`)?
0.4.4

### What version of DCOS are you using?
1.7

### What operating system and version are you using?
osx

### What did you do?
If possible please provide a recipe for reproducing.
dcos package describe --cli returns nothing for helloworld in binary-cli branch in universe because it has no command.json but has cli binaries specified in resource.json instead. 

### What did you expect to see?
cli binaries specified in resource.json

### What did you see instead?
nothing

PR #620, opened 2016-05-20
update `package describe --cli` to look at command + resource json

tamarrow:
With universe updates to change where the cli is defined, we now first
look for the cli definition in resource.cli, and if not found, look for
command.json in the package definition.


tamarrow:
#603

jsancio:
LGTM.

PR #627, opened 2016-05-24
Enable dcos node ssh to run a command

surdy:
Because `echo <command> | dcos node ssh ` felt hacky and did not work so well for me

### DIFF
--- a/cli/tests/integrations/test_node.py
+++ b/cli/tests/integrations/test_node.py
@@ -161,6 +161,18 @@ def test_slave_arg_deprecation_notice():
                    returncode=1)
 
 
+def test_node_ssh_with_command():
+    leader_hostname = mesos.DCOSClient().get_state_summary()['hostname']
+    _node_ssh(['--leader', '--master-proxy', '/opt/mesosphere/bin/detect_ip'],
+              0, leader_hostname)
+
+
+def test_node_ssh_slave_with_command():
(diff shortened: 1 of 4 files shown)


cli/tests/integrations/test_node.py
```diff
@@ -161,6 +161,16 @@ def test_slave_arg_deprecation_notice():
                    returncode=1)
 
 
+def test_node_ssh_leader_command():
```
tamarrow: can we change this to something like `test_node_ssh_with_command`? Since `dcos node ssh` is also a command, the test name is confusing

cli/tests/integrations/test_node.py
```diff
@@ -161,6 +161,16 @@ def test_slave_arg_deprecation_notice():
...
+    _node_ssh(['--leader', '--master-proxy', 'hostname'], 0)
```
tamarrow: we should check that the output of the command

cli/tests/integrations/test_node.py
```diff
@@ -161,6 +161,16 @@ def test_slave_arg_deprecation_notice():
...
+
+
+def test_node_ssh_slave_command():
```
tamarrow: can we change this to something like test_node_ssh_slave_with_command

surdy:
Thanks @tamarrow  for the review comments. I have updated the tests.

tamarrow:
LGTM

Issue #632, opened 2016-05-25
ssl_verify bogus


m0ppers:
I might be completely stupid and I am certainly not a python developer but I am currently migrating our dcos subcommand to use the dcos http library (so OAuth is properly working). When doing so I immediately got the following error:

```
$ dcos-arangodb arangodb dbservers
No handlers could be found for logger "dcos.http"
cannot connect to 'xxx/service/arangodb/v1/endpoints.json' An SSL error occurred. To configure your SSL settings, please run: `dcos config set core.ssl_verify <value>`
<value>: Whether to verify SSL certs for HTTPS or path to certs
```

So far so good. I changed the config but the error kept popping up.

After checking the source it seems to me that the flag is in fact never used. Instead the environment variable `DCOS_SSL_VERIFY` is the one controlling the behaviour. Unless I am mistaken (which is totally possible) the ssl_verify flag is bogus and should either be removed (including the misleading error message) or properly implemented

tamarrow:
The `ssl_verify` config value gets set as an env variable in `set_ssl_info_env_vars` in `cli/dcoscli/main.py`. And then in `dcos/http` we just look for the env variable. However, since you are using the `dcos/http` library and not going through dcos-cli main, the config value is not working. Will fix.

PR #635, opened 2016-05-25
use core.ssl_verify config in dcos.http

tamarrow:
We use the `ssl_verify` config property to set the corresponding env_var in dcoscli.main. However, when someone uses our http module outside of the dcos-cli, the config property doesn't work. 
#632

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -43,8 +43,10 @@ def _verify_ssl(verify=None):
     :rtype: bool | str
     """
 
-    if verify is None and constants.DCOS_SSL_VERIFY_ENV in os.environ:
-        verify = os.environ[constants.DCOS_SSL_VERIFY_ENV]
+    ssl_verify_config = util.get_config().get("core.ssl_verify")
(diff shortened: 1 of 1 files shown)


dcos/http.py
```diff
@@ -43,8 +43,10 @@ def _verify_ssl(verify=None):
     :rtype: bool | str
     """
 
-    if verify is None and constants.DCOS_SSL_VERIFY_ENV in os.environ:
-        verify = os.environ[constants.DCOS_SSL_VERIFY_ENV]
+    ssl_verify_config = util.get_config().get("core.ssl_verify")
+    if verify is None and (constants.DCOS_SSL_VERIFY_ENV in os.environ or
+                           ssl_verify_config is not None):
+        verify = os.environ[constants.DCOS_SSL_VERIFY_ENV] or ssl_verify_config
```
jsancio: We should standardize the precedence order and have a mechanism that enforces this order. E.g. environment variable > config file. We have this pattern littered in a few places in the code.

tamarrow:
Closing this in favor of #643

PR #637, opened 2016-05-27
DC slash OS

surdy:
I was bored a couple of nights ago and needed to do something to get to sleep, hence replaced all the text (non-code) references of DCOS with DC/OS. 

I thought I might as well do a PR.


tamarrow:
Thanks @surdy :)

tamarrow:
thanks @surdy!

PR #638, opened 2016-05-31
Let's start a discussion around auth with the dcos-cli

iamthemuffinman:
Having to put in a username and password every 5 days for the dcos-cli is annoying when using it via an automated build process. Granted, one could use `expect` or similar, but there isn't any documentation on this and it feels weird to use it this way. In our automated build process, usernames/passwords are typically passed in through Bamboo (CI). I don't expect this to get merged, but this is how we currently use the DC/OS cli. So, how is the DC/OS team approaching this?

### DIFF
--- a/cli/dcoscli/data/help/node.txt
+++ b/cli/dcoscli/data/help/node.txt
@@ -10,6 +10,7 @@ Usage:
                   [--user=<user>]
                   [--master-proxy]
                   (--leader | --master | --mesos-id=<mesos-id> | --slave=<slave-id>)
+                  [<command>]
 
 Commands:
     log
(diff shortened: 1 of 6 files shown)


jsancio:
We support encoding the username and password in the `core.dcos_url` property in your config. For example `dcos config set core.dcos_url https://username:password@dcoshost/`. Let me know if that works for you.

iamthemuffinman:
I suppose that works. Does the DC/OS team intend on supporting any other methods in the near future?

jsancio:
@iamthemuffinman, is there a specific requirement that you had in mind?

PR #640, opened 2016-05-31
Address missing core.dcos_url when mesos_master_url  is provided

issenec:
Added condition to address when mesos_master_url is provided in get_dcos_url.
Fixed exception statement in get_dcos_url.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -51,8 +51,11 @@ class DCOSClient(object):
         """
         if self._dcos_url:
             return urllib.parse.urljoin(self._dcos_url, path)
+        elif self._mesos_master_url:
+            mesos_master_url = ':'.join(self._mesos_master_url.split(':')[:2])
+            return urllib.parse.urljoin(mesos_master_url + '/', path)
         else:
-            raise util.missing_config_exception('core.dcos_url')
+            raise util.missing_config_exception(['core.dcos_url'])
 
     def master_url(self, path):
         """ Create a master URL
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -78,7 +78,7 @@ class DCOSClient(object):
 
         """
 
-        if private_url:
+        if self._mesos_master_url:
             return urllib.parse.urljoin(private_url, path)
         else:
             return urllib.parse.urljoin(self._dcos_url,
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -30,14 +30,8 @@ class DCOSClient(object):
 
     def __init__(self):
         config = util.get_config()
-        self._dcos_url = None
-        self._mesos_master_url = None
-
-        mesos_master_url = config.get('core.mesos_master_url')
-        if mesos_master_url is None:
-            self._dcos_url = util.get_config_vals(['core.dcos_url'], config)[0]
-        else:
-            self._mesos_master_url = mesos_master_url
+        self._dcos_url = util.get_config_vals(['core.dcos_url'], config)[0]
+        self._mesos_master_url = config.get('core.mesos_master_url')
 
         self._timeout = config.get('core.timeout')
 
@@ -51,9 +45,6 @@ class DCOSClient(object):
         """
         if self._dcos_url:
             return urllib.parse.urljoin(self._dcos_url, path)
-        elif self._mesos_master_url:
-            mesos_master_url = ':'.join(self._mesos_master_url.split(':')[:2])
-            return urllib.parse.urljoin(mesos_master_url + '/', path)
         else:
             raise util.missing_config_exception(['core.dcos_url'])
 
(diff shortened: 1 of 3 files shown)


dcos/mesos.py
```diff
@@ -51,8 +51,11 @@ def get_dcos_url(self, path):
         """
         if self._dcos_url:
             return urllib.parse.urljoin(self._dcos_url, path)
+        elif self._mesos_master_url:
+            mesos_master_url = ':'.join(self._mesos_master_url.split(':')[:2])
+            return urllib.parse.urljoin(mesos_master_url + '/', path)
```
tamarrow: Instead of this `elif` we should just remove [this if](https://github.com/issenec/dcos-cli/blob/25bdc9561b9850824c291ec6a96468ab837d626a/dcos/mesos.py#L37) and always set `self._dcos_url`.

dcos/mesos.py
```diff
@@ -51,8 +51,11 @@ def get_dcos_url(self, path):
...
         else:
-            raise util.missing_config_exception('core.dcos_url')
+            raise util.missing_config_exception(['core.dcos_url'])
```
tamarrow: 👍

dcos/mesos.py
issenec: Agreed this would be the best fix: the documentation in both the Github page and DCOS specify that `dcos_url` has to be specified. Also, I'm pretty sure there's multiple places in the code that require that specific variable.

However, I don't know if there's any implications of doing that. If you guys can confirm that there's no issue, I can remove the if and elif statements.

tamarrow: looked again, and removing that `if` won't be a problem

dcos/mesos.py
```diff
@@ -30,14 +30,8 @@ class DCOSClient(object):
...
         config = util.get_config()
-        self._dcos_url = None
-        self._mesos_master_url = None
-
-        mesos_master_url = config.get('core.mesos_master_url')
-        if mesos_master_url is None:
-            self._dcos_url = util.get_config_vals(['core.dcos_url'], config)[0]
-        else:
-            self._mesos_master_url = mesos_master_url
+        self._dcos_url = util.get_config_vals(['core.dcos_url'], config)[0]
```
tamarrow: This can just be `config.get('core.dcos_url')`

tamarrow:
@issenec one small comment, and then LGTM. Thanks for fixing!

dcos/mesos.py
issenec: I considered doing this, but I think the call to get_config_vals acts a check to make sure the variable is provided. 

I think a better solution would be to remove all the checks for `self._dcos_url` in DCOSClient and leave the `util.get_config_vals` in the to confirm its existence. Sounds good? (Making changes as I post this)

tamarrow: Reading the config everytime seems a little overkill since we are passing the variable along; so either it already threw an exception or it exists. We can be more explicit here with something like:
```
self._dcos_url = config.get("dcos_url")
if self._dcos_url is None:
     raise config.missing_config_exception(['core.dcos_url'])
```

issenec:
@tamarrow Np :D 

I just don't want anybody else to go through what I did to understand how DCOS works. Having created the manual process of setting up my environment through Ansible, I can see why DCOS exists, but it's the little things (e.g. Where do I configure JAVA/HADOOP_HOME? Why does the UI keep tracking my dead Spark processes?) that make it hard for me to ever tell my managers "Hey, we should switch to DCOS instead of using baremetal Mesos."

As someone who fell into sysadmin and is still trying to understand how things work on Mesos, it seems that the documentation is great for the brand new system, but if there's something different (e.g. your storage drives don't include /var/lib) or you're coming from an older system, then there'll be a lot of head banging and profanity. I definitely naive and stupid, but if I have to understand an error, I might as well start looking at the code base.

dcos/mesos.py
```diff
@@ -30,14 +30,10 @@ class DCOSClient(object):
 
     def __init__(self):
         config = util.get_config()
```
tamarrow: we need to call this `toml_config` or something since we are using a method from the config module on line 35, and the config variable is overwriting the module import name in this method

tamarrow:
We are working to improve our documentation and hopefully every new DC/OS version will only makes things easier :)

dcos/mesos.py
issenec: My bad: I dummied and forgot to follow config to realize that it would not have the function. I changed out the config in line 35 with util to fix it and `config` is not used anywhere else inside the class (aside form the __init__).

tamarrow:
@issenec did you see the failing unit tests? We need to update them because of this code change

dcos/mesos.py
```diff
@@ -83,11 +78,11 @@ def slave_url(self, slave_id, private_url, path):
 
         """
 
-        if self._dcos_url:
+        if private_url:
```
tamarrow: @issenec this should be `if self._mesos_master_url`

PR #644, opened 2016-06-03
make clear resolution of ENV_VAR precedence over config property

tamarrow:


tamarrow:
fixes #632

dcos/config.py
```diff
@@ -52,24 +54,36 @@ def get_config(mutable=False):
...
+    if config is None:
+        config = get_config()
+
+    section, subkey = split_key(name.upper())
+
+    env_var = None
+    if section == "CORE":
+        env_var = "DCOS_{}".format(subkey)
+        if subkey.startswith("DCOS") and os.environ.get(subkey):
+            env_var = subkey
```
jsancio: This is a bit confusing to read. Maybe we can write this as:

```
if subkey.startswith("DCOS") and os.environ.get(subkey):
    env_var = subkey
else:
    env_var = "DCOS_{}".format(subkey)
```

Let's write a short comment as to why we are doing this.

jsancio:
LGTM @tamarrow . Do we need file an issue to print the help information around all the config properties and env variable. Maybe `dcos help config` or `dcos config describe`?

tamarrow:
created issue #650

dcos/config.py
tamarrow: updated

PR #651, opened 2016-06-08
Check the config value for core.ssl_verify

jsancio:


dcos/config.py
```diff
@@ -67,7 +67,7 @@ def get_config_val(name, config=None):
     :param config: config
     :type config: Toml
     :returns: value of 'name' parameter
-    :rtype: object | None
+    :rtype: str | None
```
jsancio: I think this is correct. It should always return `str` or `None`, no?

tamarrow: yes

dcos/http.py
```diff
@@ -44,9 +44,9 @@ def _verify_ssl(verify=None):
 
     if verify is None:
         verify = config.get_config_val("core.ssl_verify")
-        if verify.lower() == "true":
+        if verify and verify.lower() == "true":
```
tamarrow: this will fail if `verify` is a bool.

jsancio: If you believe the pydoc for `get_config_val` then at this point verify should be a `str` or `None`. In that case `verify` is true if `verify is not None and verify != ""`.

tamarrow: ah yes

PR #656, opened 2016-06-16
handle mesos tasks having ids that are substrings of each other

tamarrow:
Instead of crashing, if looking for a task by a fltr, if a task is an
exact match, return that task.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -331,7 +331,8 @@ class Master(object):
                                         'slave/{}/'.format(slave['id']))
 
     def slave(self, fltr):
-        """Returns the slave that has `fltr` in its ID.  Raises a
+        """Returns the slave that has `fltr` in its ID. If any slaves
+        are an exact match, returns that task, id not raises a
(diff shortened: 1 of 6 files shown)


dcos/mesos.py
```diff
@@ -346,10 +347,15 @@ def slave(self, fltr):
             raise DCOSException('No slave found with ID "{}".'.format(fltr))
 
         elif len(slaves) > 1:
-            matches = ['\t{0}'.format(slave['id']) for slave in slaves]
+
+            exact_matches = [s for s in slaves if s['id'] == fltr]
+            if len(exact_matches) == 1:
+                return exact_matches[0]
+
```
jsancio: Should we add an `else` to make it clear?

jsancio:
LGTM. One minor comment. Can we write a test for this?

tamarrow:
thanks! updated with a unit test.

jsancio:
:shipit:

jsancio:
Fixes #660

PR #658, opened 2016-06-18
default requests to cosmos to version 2, with a version 1 fallback

tamarrow:
#655

### DIFF
--- a/dcos/cosmospackage.py
+++ b/dcos/cosmospackage.py
@@ -1,11 +1,12 @@
+import base64
 import collections
 import functools
 
 import six
 from dcos import emitting, http, util
 from dcos.errors import (DCOSAuthenticationException,
-                         DCOSAuthorizationException, DCOSException,
-                         DCOSHTTPException, DefaultError)
+                         DCOSAuthorizationException, DCOSBadRequest,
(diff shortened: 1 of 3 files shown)


dcos/cosmospackage.py
```diff
@@ -451,31 +460,37 @@ def package_versions(self):
...
     :type request_type: str
+    :param verison: version of request
+    :type version: str
     :returns: header information
     :rtype: str
     """
 
     return ("application/vnd.dcos.package.{}+json;"
-            "charset=utf-8;version=v1").format(request_type)
+            "charset=utf-8;version={}").format(request_type, version)
```
jsancio: I think we should instead construct all of the possible content type ahead of time. For example most RPCs in Cosmos are going to stay as `v1`. The type of this my could be: `Map[String, List[String]`, the key is the `request_type` and the value is a list of supported content-type in preference order. What do you think?

tamarrow: yes, updated!

tamarrow:
@jsancio can you review please? I removed your commit so that tests will pass (once testing/master is updated to 1.8 we can merge your pr)

PR #659, opened 2016-06-20
Add explicit support for py3.5

ticosax:
Hi,
Since python3.5 works out of the box, I thought it would be useful to advertise it.

### DIFF
--- a/cli/tox.ini
+++ b/cli/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{34,35}-syntax, py{27,34,35}-unit, py{27,34,35}-integration
+envlist = py35-syntax, py{27,34,35}-unit, py{27,34,35}-integration
 
 [testenv]
 deps =
--- a/cli/tox.ini
+++ b/cli/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py34-syntax, py{27,34}-unit, py{27,34}-integration
+envlist = py{34,35}-syntax, py{27,34,35}-unit, py{27,34,35}-integration
 
 [testenv]
 deps =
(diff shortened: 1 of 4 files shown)


cli/tox.ini
```diff
@@ -10,7 +10,7 @@ deps =
   -e..
 passenv = DCOS_* CI_FLAGS CLI_TEST_SSH_KEY_PATH CLI_TEST_MASTER_PROXY
 
-[testenv:py34-syntax]
```
tamarrow: We should keep syntax checks in one python version, or else we'll have conflicts :)

tamarrow:
@ticosax Thanks for contributing! I'm not sure that all features work in python 35, but if all the tests pass will merge! Looks like we need to require a higher version of tox to include py35 support in `requirements.txt`

cli/tox.ini
ticosax: If you prefer, I think it is better to check with latest supported python version to catch `async` keyword used as a variable, for instance,

```python
def foo(async=True):
    pass
```
This snippet will pass with python3.4 but not with python3.5

ticosax:
> Looks like we need to require a higher version of tox to include py35 support in requirements.txt

According to what I [read](https://teamcity.mesosphere.io/viewLog.html?buildId=306305&buildTypeId=DcosIo_DcosCli_Ci&tab=buildLog). This is not a `tox` issue, it is because python3.5 is not available in testing environment.

tamarrow:
ok to test

tamarrow:
@ticosax I'm going to close this pr because it requires updating the CIs and we already have an issue to upgrade our supported version of python to 35: #766 

If you disagree, feel free to re-open the pr.

PR #664, opened 2016-06-24
use BatchMode instead of PasswordAuthentication for ssh

tamarrow:
DC/OS bumped the coreOS version, and just using PasswordAuthentication
is now no longer enough to turn off interactive mode. BatchMode covers
all types of authentication.


jsancio:
LGTM!

PR #648, opened 2016-06-04
remove code duplication

tamarrow:
#566

PR #667, opened 2016-06-28
revert "remove code duplication (#648)"

tamarrow:
With this change, we need to specify the imports in pyinstaller spec
file, so leave as is.


jsancio:
LGTM!

PR #669, opened 2016-06-29
only support python3 for creating binaries

tamarrow:
Also explicitly use python3 in creating virtualenv for make commands.


bin/env.sh
```diff
@@ -4,7 +4,7 @@ BASEDIR=`dirname $0`/..
 
 if [ ! -d "$BASEDIR/env" ]; then
 
-    virtualenv -q $BASEDIR/env --prompt='(dcos) '
+    virtualenv -p python3 -q $BASEDIR/env --prompt='(dcos) '
```
jsancio: It looks like `venv` is the modern way of creating environment in python3: https://docs.python.org/3/library/venv.html

jsancio:
Looks good. One minor comment.

bin/env.sh
tamarrow: It's currently broken on the ubuntu docker we're using: https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847

I've failed an issue #670

setup.py
```diff
@@ -10,6 +11,24 @@
...
+    'pypng==0.0.18',
+    'pystache>=0.5, <1.0',
+    'requests>=2.6, <3.0',
+    'six>=1.9, <2.0',
+    'toml>=0.9, <1.0',
+]
+if sys.version_info[0] == 2:
+    install_requires = [
+        "futures>=3.0, <4.0'"
+    ]
```
jsancio: @tamarrow, Did you mean:
```
if sys.version_info[0] == 2:
  install_requires = intall_requires + ['...']
```

cli/setup.py
```diff
@@ -10,6 +11,19 @@
...
+    'docopt>=0.6, <1.0',
+    'pkginfo==1.2.1',
+    'toml>=0.9, <1.0',
+    'virtualenv>=13.0, <14.0',
+    'rollbar>=0.9, <1.0',
+]
+if sys.version_info[0] == 2:
+    install_requires = [
+        "futures>=3.0, <4.0'"
+    ]
```
jsancio: @tamarrow, Did you mean:
```
if sys.version_info[0] == 2:
  install_requires = intall_requires + ['...']
```

setup.py
tamarrow: yes :)

PR #672, opened 2016-07-05
remove v2/info call from Marathon client

tamarrow:
We were using it to make sure we have a recent version of Marathon. Since
 DC/OS CLI 0.4.0+ is only compatible with DC/OS 1.6.1, the Marathon
 check is no longer necessary. It is also messing with ACLs since
 v2/apps is at a user level and v2/info is an admin level.

### DIFF
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -675,7 +675,7 @@ def test_bad_configuration():
     config_set('marathon.url', 'http://localhost:88888')
 
     returncode, stdout, stderr = exec_command(
-        ['dcos', 'marathon', 'app', 'list'])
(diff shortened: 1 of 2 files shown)


alberts:
:ship:

jsancio:
I see. What error do we expect to see if the user tries to use this CLI against a DC/OS cluster less than `1.6.1`?

tamarrow:
@jsancio package commands will get the "Please downgrade to 0.3.x errors". Marathon commands should still work - DC/OS 1.6 is already at marathon 0.18.x. But users of old versions of DC/OS should hopefully never end up with 0.4.x CLI :)

jsancio:
Sounds good! :shipit:

PR #675, opened 2016-07-06
rename snapshot command

darkonie:
a new command to manage a set of cluster logs would be:
dcos node diagnostics ...

a zip file with collected logs is now called a `diagnostics bundle`


tamarrow:
@darkonie we're going to have to add deprecation notices on the `snapshot` commands

darkonie:
@tamarrow thanks, will do

malnick:
@tamarrow Is that neccessary? Have we actually released the snapshot command to the community yet? 1.8EA isn't until Monday. 

CC @spahl

jsancio:
I think we fix it before EA then we don't need to add a deprecation warning.

tamarrow:
okay

Issue #666, opened 2016-06-24
Tests: Compare metadata fields as Python dict


jsancio:
Instead of comparing metadata fields as base64 string we should compare them as python dicts.

Many of the test do raw string comparisons. We should instead decode the base64 string, parse the JSON and compare them as strings.

jsancio:
Let's fix the following places:

- [ ] https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_package.py#L391-L401
- [ ] https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_package.py#L258-L259
- [ ] https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_package.py#L197-L198

PR #681, opened 2016-07-08
Resolved Issue #666

larioj:
I added comparison of metadata as python dictionaries.

### DIFF
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -1,10 +1,11 @@
+import base64
 import collections
 import contextlib
 import json
 import os
 import subprocess
 import time
-import base64
+
 
 import six
 from dcos import config, http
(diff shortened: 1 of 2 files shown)


cli/tests/integrations/common.py
```diff
@@ -1,10 +1,12 @@
+import base64
 import collections
 import contextlib
 import json
 import os
 import subprocess
 import time
 
+
```
cruhland: Remove extra line?

cli/tests/integrations/test_package.py
```diff
@@ -256,7 +263,12 @@ def test_describe_options():
     actual_labels = stdout_.pop("labels", None)
 
     for label, value in expected_labels.items():
-        assert value == actual_labels.get(label)
+        if label == "DCOS_PACKAGE_METADATA":
+            # We can covert the metadata into a dictionary
```
cruhland: Could you rephrase this comment to explain why converting into a dictionary is a good idea? The fact that we're doing the conversion is clear from the code, but it's less clear why we would want to.

cruhland:
LGTM after addressing comments.

Issue #684, opened 2016-07-12
Make dcos config show/set error output consistent


larioj:
The config command output is not consistent

```
dcos-cli$ dcos config show core
Property 'core' doesn't fully specify a value - possible properties are:
core.dcos_acs_token
core.dcos_url
core.reporting
core.ssl_verify
core.timeout
```

```
dcos-cli$ dcos config set core.foo bar
Error: 'auieoau' is not a valid property. Possible properties are: pagination, dcos_url, reporting, mesos_master_url, dcos_acs_token, timeout, ssl_verify
```

PR #685, opened 2016-07-12
Resolving Issue 684

larioj:
Changed the formatting, see issue #684.

### DIFF
--- a/dcos/jsonitem.py
+++ b/dcos/jsonitem.py
@@ -64,10 +64,10 @@ def find_parser(key, schema):
 
     key_schema = schema['properties'].get(key)
     if key_schema is None:
-        keys = ', '.join(schema['properties'].keys())
+        keys = '\n'.join(schema['properties'].keys())
         raise DCOSException(
-            'Error: {!r} is not a valid property. '
-            'Possible properties are: {}'.format(key, keys))
+            'Property {!r} is invalid - '
+            'possible properties are: \n{}'.format(key, keys))
     else:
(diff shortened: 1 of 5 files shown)


dcos/jsonitem.py
```diff
@@ -64,10 +64,10 @@ def find_parser(key, schema):
 
     key_schema = schema['properties'].get(key)
     if key_schema is None:
-        keys = ', '.join(schema['properties'].keys())
+        keys = '\n'.join(schema['properties'].keys())
         raise DCOSException(
-            'Error: {!r} is not a valid property. '
-            'Possible properties are: {}'.format(key, keys))
+            'Property {!r} is invalid - '
+            'possible properties are: \n{}'.format(key, keys))
```
cruhland: Is the other occurrence of this code in another file? Would it be feasible to factor this out as a "keys formatting" method?

cruhland:
@larioj looks good, but the failing Linux Binary build is a bit concerning. Is it related to this change?

larioj:
The linux build failed for my other PR. Jose said that the only thing that mattered was the continuous-integration/teamcity

dcos/jsonitem.py
larioj: We can't really factor it out since the other version outputs the keys as core.foo while this one outputs foo. Also the beginning part of the string is different so the formatting function would be fairly small.

cruhland:
Ok, LGTM. :ship: it when you're ready.

PR #691, opened 2016-07-25
fix docopt resolution order in `dcos task`

tamarrow:
If no arguments were given for`dcos task log` and `dcos task ls`, we
were processing the command as `dcos task` with task-id log|ls.


jsancio:
Interesting. Good catch! Let's write a test for it otherwise it looks good to me.

PR #694, opened 2016-07-26
use config-schema datafiles for default commands

tamarrow:
We shouldn't have the binary need to call itself, especially for information it already has.

### DIFF
--- a/dcos/config.py
+++ b/dcos/config.py
@@ -323,9 +323,14 @@ def get_config_schema(command):
             pkg_resources.resource_string(
                 'dcos',
                 'data/config-schema/core.json').decode('utf-8'))
-
-    executable = subcommand.command_executables(command)
-    return subcommand.config_schema(executable, command)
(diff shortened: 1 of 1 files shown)


dcos/config.py
```diff
@@ -323,9 +323,14 @@ def get_config_schema(command):
...
                 'dcos',
                 'data/config-schema/core.json').decode('utf-8'))
-
-    executable = subcommand.command_executables(command)
-    return subcommand.config_schema(executable, command)
+    elif command in subcommand.default_subcommands():
+        return json.loads(
+            pkg_resources.resource_string(
+                'dcoscli',
+                'data/config-schema/{}.json'.format(command)).decode('utf-8'))
```
jsancio: @tamarrow this code is in the `dcos` package and doesn't have access to the `dcoscli` package's data files. If we want to do this then we should move the data files to the `dcos` package.

PR #695, opened 2016-07-26
fix issue 666

larioj:
* Add a function name base64_to_dict that converts a base64 string into
a python dictionary
* Change the base64 string comparisons in test_package.py to comparisons
in their python dictinary form. This will make error messages more
descriptive

### DIFF
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -1,3 +1,4 @@
+import base64
 import collections
 import contextlib
 import json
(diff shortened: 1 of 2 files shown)


cli/tests/integrations/common.py
```diff
@@ -549,3 +550,7 @@ def config_unset(key, env=None):
 
     assert returncode == 0
     assert stdout == b''
+
+
+def base64_to_dict(byte_string):
```
tamarrow: please add pydoc

tamarrow:
@larioj one small comment and then LGTM.

Issue #673, opened 2016-07-05
`dcos package describe` produces incorrect result


mgummelt:
Please answer the following questions before submitting your issue. Thanks!

### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?
If DC/OS CLI version < 0.4.5, DC/OS version can be found at `http://dcos_url/dcos-metada/dcos-version.json`

```
> dcos --version
dcoscli.version=SNAPSHOT
dcos.version=1.8-dev
dcos.commit=ef83e6e87d55f64ff847773d020379fea0d16402
dcos.bootstrap-id=817ce8f4c1248102a7733ba50d76b7a42209d5cb
```

### What operating system and version are you using?

```
> uname -a
Linux mg-mesos 4.2.0-36-generic #42-Ubuntu SMP Thu May 12 22:05:35 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

### What did you do?
If possible please provide a recipe for reproducing.

```
> dcos package describe spark
{
  "description": "Spark is a fast and general cluster computing system for Big Data.  Documentation: https://docs.mesosphere.com/usage/managing-services/spark/",
  "framework": null,
  "licenses": [
    {
      "name": "Apache License Version 2.0",
      "url": "https://raw.githubusercontent.com/apache/spark/master/LICENSE"
    }
  ],
  "maintainer": "support@mesosphere.io",
  "name": "spark",
  "packagingVersion": "2.0",
  "postInstallNotes": "DC/OS Spark is being installed!\n\n\tDocumentation: https://docs.mesosphere.com/usage/services/spark/\n\tIssues: https://docs.mesosphere.com/support/",
  "postUninstallNotes": "The Apache Spark DC/OS Service has been uninstalled and will no longer run.\nPlease follow the instructions at http://docs.mesosphere.com/services/spark/#uninstall to clean up any persisted state.",
  "preInstallNotes": null,
  "scm": "https://github.com/apache/spark.git",
  "selected": true,
  "tags": [
    "bigdata",
    "mapreduce",
    "batch",
    "analytics"
  ],
  "versions": [
    "1.6.0",
    "1.0.0-1.6.1-2",
    "1.5.0-multi-roles-v2"
  ],
  "website": "https://docs.mesosphere.com/usage/managing-services/spark/"
}
```

### What did you expect to see?

A single version in the `versions` field.

### What did you see instead?

[...]

PR #699, opened 2016-07-27
change `package describe` to output package.json without modification

tamarrow:
We were adding in all versions of a package, but this doesn't make sense
since different versions have different package.json contents. A user
can still see all versions of a package with `--package-versions` flag.


tamarrow:
#673

jsancio:
I am okay with this change. I worry that we may break scripts that use the old JSON format. Can you ping the Infinity and Velocity team to verify that it doesn't break any of their scripts?

tamarrow:
Yep, got the ok from both teams.

Issue #396, opened 2015-12-21
dcos-marathon should support --service=marathon-user


sttts:
Right now it is only possible to use one marathon framework instance with the marathon cli command. Compare https://github.com/mesosphere/dcos-cli/blob/master/dcos/marathon.py#L31.

As users might install multiple marathons, the cli should support this, i.e. switching via the command line.

aquamatthias:
+1

sttts:
Setting marathon.url in the dcos config works at least. But a user will never find out without reading the code.

```
[marathon]
url = "http://pool-c73d-elasticl-13c45jhfzyebl-1471334914.us-west-2.elb.amazonaws.com/service/marathon-user/"
```

ssk2:
+1

jsancio:
Adding both the `marathon` and `core` label since we may want to resolve it in a general so that all command that may use this feature.

PR #702, opened 2016-07-31
service name for dcos

kensipe:
a solution to using different service names for deployed packages with a solution for marathon

a proposed and working solution for https://github.com/dcos/dcos-cli/issues/396

I was going to just fix marathon, but it was mentioned it was a core issue.  I'm on the fence about using os.env for the value... but we need a way to pass around state without cascading params.  Another option would be adding to a config object.  Looking for feedback:

@jsancio @tamarrow

### DIFF
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -34,13 +34,19 @@ def _get_marathon_url(toml_config):
     :returns: marathon base url
     :rtype: str
     """
-
     marathon_url = config.get_config_val('marathon.url', toml_config)
     if marathon_url is None:
         dcos_url = config.get_config_val('core.dcos_url', toml_config)
         if dcos_url is None:
             raise config.missing_config_exception(['core.dcos_url'])
-        marathon_url = urllib.parse.urljoin(dcos_url, 'service/marathon/')
+
+        if util.get_nondefault_service_name() is None:
+            marathon_service_name = "marathon"
+        else:
+            marathon_service_name = util.get_nondefault_service_name()
+
+        marathon_service_url = 'service/{}/'.format(marathon_service_name)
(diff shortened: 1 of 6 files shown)


kensipe:
when we have agreement on the approach I will add additional tests.

dcos/util.py
```diff
@@ -654,6 +654,19 @@ def validate_png(filename):
...
+
+    :rtype: str or None
+
+    """
+
+    if constants.DCOS_SERVICE_NAME_ENV in os.environ:
+        return os.environ[constants.DCOS_SERVICE_NAME_ENV]
+    else:
+        return None
+
```
tamarrow: if we add an `env var` we need to also add a config variable for consistency. see https://github.com/dcos/dcos-cli/blob/master/dcos/config.py#L59

dcos/marathon.py
```diff
@@ -34,13 +34,19 @@ def _get_marathon_url(toml_config):
...
             raise config.missing_config_exception(['core.dcos_url'])
-        marathon_url = urllib.parse.urljoin(dcos_url, 'service/marathon/')
+
+        if util.get_nondefault_service_name() is None:
+            marathon_service_name = "marathon"
+        else:
+            marathon_service_name = util.get_nondefault_service_name()
+
+        marathon_service_url = 'service/{}/'.format(marathon_service_name)
+        marathon_url = urllib.parse.urljoin(dcos_url, marathon_service_url)
```
tamarrow: If we want to make this a more global feature, we should have something like `get_service_name` so its clear for third party developers to use this.

tamarrow:
@kensipe are you still working on this?

tamarrow:
@kensipe closing this branch since it's stale. If you are still working on it please reopen. Thanks!

PR #703, opened 2016-08-02
remove pypi universal config param

tamarrow:
Changing from using [universal wheels](https://packaging.python.org/distributing/#universal-wheels) to [pure python wheels](https://packaging.python.org/distributing/#pure-python-wheels)


jsancio:
@tamarrow Can you test that doing this doesn't break the existing install script?

tamarrow:
Going to hold off on this for now

PR #704, opened 2016-08-02
clean up tests so they pass locally + in the CI

tamarrow:


jsancio:
LGTM! :shipit:

README.rst
```diff
@@ -101,17 +101,37 @@ Running Tests
...
+directory to one of those two and follow the instructions.
+
+
+Initialization
+#######
+
+The `config` integration tests use static config files. To run these tests
+make sure you set owner only permissions on these files:
+
+:code:`chmod 600 cli/tests/data/dcos.toml`
```
kensipe: what prevents us from checking in the file with the correct permissions.  I like that we documented this... however I would love for it to be automated.

If there is reason to not check in a file with the correct attributes, then can we have a script that automate this?

README.rst
```diff
@@ -101,17 +101,37 @@ Running Tests
...
+:code:`echo "dcos/cluster/url dcos.snakeoil.mesosphere.com" >> /etc/hosts`
+
 
 Running
 #######
 
-Tox will run unit and integration tests in both Python environments using a
-temporarily created virtualenv.
+Tox will run unit and integration tests in Python 3.4 using a temporarily
+created virtualenv.
```
kensipe: is it ok to check in a `.python-version` with the project which can 1) provide documentation of expectation and auto use the correct version for those using `pyenv`

kensipe:
I have a few questions but otherwise lgtm.

README.rst
tamarrow: great idea, I added to the `make test` command

msabramo: Also see #969, which does it in Python, in case someone doesn't notice the `Makefile` and runs `tox` directly (like I did 😄 )

PR #710, opened 2016-08-10
Make sure that marathon URL ends with trailing slash

vespian:
This small patch makes sure that marathon urls always end with slash. Otherwise, in some cases urljoin can "eat" the last component of the path.

### DIFF
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -130,6 +130,8 @@ class Client(object):
     """
 
     def __init__(self, marathon_url, timeout=http.DEFAULT_TIMEOUT):
+        if not marathon_url.endswith('/'):
+            marathon_url += '/'
         self._base_url = marathon_url
         self._timeout = timeout
 


dcos/marathon.py
```diff
@@ -130,6 +130,8 @@ class Client(object):
     """
 
     def __init__(self, marathon_url, timeout=http.DEFAULT_TIMEOUT):
+        if not marathon_url.endswith('/'):
+            marathon_url += '/'
         self._base_url = marathon_url
```
tamarrow: How about `os.path.join(marathon_url, '')`? can we move this to `_get_marathon_url`? @vespian

vespian: I think that os.path.join will give you different results depending on platform ('/' on Linux vs '\' on Windows machines. 

This patch is meant only as quick fix, but if you want to fix this issue globally, instead of _get_marathon_url I would try to patch `_create_url` and substitue
```
return urllib.parse.urljoin(self._base_url, path)
```
with i.e. something similar to what @jgehrcke suggested:
```
'/'.join(c.strip('/') for c in components)
```
I was afraid that it can break other things so I provided a "localized" patch that poses very low risk of breaking stuff.

tamarrow: I think we should just do a more global patch so we don't end up having to make another quick fix laster.

`/`.join() won't add a trailing slash. 

something like this will work: urllib.parse.urljoin(self._base_url, urllib.parse, urljoin(path, '/'))`

vespian: After longer discussion on slack we agreed that for now we will implement this minimalistic patch and later we can try to fix it globally by first tracking all the places that use urljoin and then making sure that correct behavior will not brake user's tools.

PR #714, opened 2016-08-15
test updates due to cosmos adding screenshots in metadata label

tamarrow:
Screenshots used to be removed in the label if they weren't populated, but now they are reported with `screenshots: None`.


BenWhitehead:
:ship:

PR #718, opened 2016-08-15
Make sure that marathon URL ends with trailing slash

tamarrow:


tamarrow:
duplication of #710 so I can update the branch with the latest changes.

PR #719, opened 2016-08-16
Test tweaks

cruhland:
- Explicit version for pyvenv executable
- Increase sleep time for SSH to accomodate remote clients

### DIFF
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -630,7 +630,7 @@ def ssh_output(cmd):
     proc, master = popen_tty(cmd)
 
     # wait for the ssh connection
-    time.sleep(3)
+    time.sleep(5)
 
     proc.poll()
     returncode = proc.returncode
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -4,7 +4,7 @@ BASEDIR=`dirname $0`/..
 
 if [ ! -d "$BASEDIR/env" ]; then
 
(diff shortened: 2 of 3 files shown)


cruhland:
I have more improvements for local test execution in mind, but I'll save them for future PRs. I'll create issues for them.

cli/tests/integrations/common.py
```diff
@@ -630,7 +630,7 @@ def ssh_output(cmd):
     proc, master = popen_tty(cmd)
 
     # wait for the ssh connection
-    time.sleep(3)
+    time.sleep(5)
```
tamarrow: can we do some kind of loop to test for the connection instead of just increasing the sleep?

cruhland: I was going to file an issue to fix this "the right way", because it could be somewhat involved. But I can look into it now if you want.

cruhland:
@tamarrow some of the builds are failing:

- **OS X Binary**: a bunch of test failures. Any idea why these results are different from the Linux results?
- **default**: this appears to be the Windows binary build, and it looks like the CCM cluster didn't come up, or wasn't detected correctly.
- **Jenkins** - hanging build?

Looks like the OS X build is required, so I'll try to fix it. If the other two builds are useful, let's make them required and I can fix them too. If not, can I disable those builds? I'd rather not have them if they are always broken.

cruhland:
Closing this due to outdated checks, will resubmit.

PR #726, opened 2016-08-17
Test tweaks

cruhland:
- Explicit version for pyvenv executable
- Increase sleep time for SSH to accomodate remote clients

PR #729, opened 2016-08-17
Remove extra call to the Cosmos /package/render endpoint

### LINKED ISSUE #712: `package` describe should not call render unless specified
Render gets called here: https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/package/main.py#L263

That call should happen here: https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/package/main.py#L269

cruhland:
Fixes #712.

~~Depends on #726. **Do not merge before that!**~~

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -238,18 +238,19 @@ def _describe(package_name,
     :rtype: int
     """
 
-    # If the user supplied template options, they definitely want to
-    # render the template
-    if options_path:
-        render = True
-        # Expand ~ in the options file path
-        options_path = os.path.expanduser(options_path)
-
     if package_versions and \
        (app or cli or options_path or render or package_version or config):
         raise DCOSException(
             'If --package-versions is provided, no other option can be '
             'provided')
+
+    # If the user supplied template options, they definitely want to
+    # render the template
+    if options_path:
+        render = True
+
+    # Fail early if options file isn't valid
+    user_options = _user_options(options_path)
 
     package_manager = _get_package_manager()
     pkg = package_manager.get_package_version(package_name, package_version)
@@ -259,14 +260,11 @@ def _describe(package_name,
     if package_versions:
         emitter.publish(pkg.package_versions())
     elif cli or app or config:
-        user_options = _user_options(options_path)
-        options = pkg.options(user_options)
-
         if cli:
             emitter.publish(pkg.cli_definition())
         if app:
             if render:
-                app_output = pkg.marathon_json(options)
+                app_output = pkg.marathon_json(user_options)
             else:
                 app_output = pkg.marathon_template()
                 if app_output and app_output[-1] == '\n':
@@ -292,6 +290,9 @@ def _user_options(path):
     if path is None:
         return {}
     else:
+        # Expand ~ in the path
+        path = os.path.expanduser(path)
+
         with util.open_file(path) as options_file:
             return util.load_json(options_file)
 
@@ -348,9 +349,7 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
         # Install both if neither flag is specified
         cli = app = True
 
-    # Expand ~ in the options file path
-    if options_path:
-        options_path = os.path.expanduser(options_path)
+    # Fail early if options file isn't valid
     user_options = _user_options(options_path)
 
     package_manager = _get_package_manager()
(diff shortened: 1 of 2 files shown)


cli/dcoscli/package/main.py
```diff
@@ -366,9 +365,6 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
 
     if app and pkg.has_mustache_definition():
 
-        # render options before start installation
-        options = pkg.options(user_options)
```
cruhland: Since `/package/install` renders the options, this didn't seem necessary. It did require one of the tests to be modified, but the change seems harmless.

cli/tests/integrations/test_package.py
```diff
@@ -301,7 +301,11 @@ def test_describe_specific_version():
 
 def test_bad_install():
     args = ['--options=tests/data/package/chronos-bad.json', '--yes']
-    stdout = b""
+    stdout = (
+        b'We recommend a minimum of one node with at least 1 '
+        b'CPU and 2GB of RAM available for the Chronos '
+        b'Service.\n'
+        b'Installing Marathon app for package [chronos] version [2.4.0]\n')
```
cruhland: The last line here is new, because I removed [this line](https://github.com/dcos/dcos-cli/pull/729/files#diff-1649767c70a587014cbfef7156f395edL370). But otherwise the behavior is the same as before, with one less request made to Cosmos.

cli/dcoscli/package/main.py
tamarrow: We did this explicitly because we want to validate user options before we try installing. That way a user doesn't see the preInstallNotes and then a 'oops bad options' message.

cruhland: Ok, I'll revert this part, and add your explanation to the comment.

cruhland: @tamarrow updated!

cli/dcoscli/package/main.py
```diff
@@ -366,8 +365,9 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
...
+        # Even though package installation will check for template rendering
+        # errors, we want to fail early, before we display "Installing..."
```
tamarrow: before trying to install

cli/tests/integrations/test_package.py
```diff
@@ -301,7 +301,10 @@ def test_describe_specific_version():
 
 def test_bad_install():
```
tamarrow: why are these tests changing?

cruhland: Leftovers from my earlier changes. Looks like they don't really do anything, so I'll revert them too.

cruhland:
@tamarrow another update

cli/dcoscli/package/main.py
```diff
@@ -377,10 +377,7 @@ def _install(package_name, package_version, options_path, app_id, cli, app,
 
         emitter.publish(msg)
 
-        package_manager.install_app(
-            pkg,
-            options,
-            app_id)
+        package_manager.install_app(pkg, user_options, app_id)
```
tamarrow: why did we change this from options to user_options?

tamarrow:
@cruhland one small comment and then LGTM

cli/dcoscli/package/main.py
cruhland: So, `options` previously was returned from `pkg.options(user_options)`. If you look at the definition of that method, it just returns its argument unchanged (with the side effect of sending it in a request to Cosmos).

Maybe I should update that method to not return anything. What do you think?

tamarrow: yea, that sounds good. lets make the side effect clear.

cruhland:
@tamarrow updated, I'll merge once the builds finish.

dcos/cosmospackage.py
```diff
@@ -495,17 +495,14 @@ def has_mustache_definition(self):
...
 
     def options(self, user_options):
-        """Makes sure user supplied options are valid, and returns valid options
+        """Makes sure user supplied options are valid
 
-        :param options: the template options to use in rendering
-        :type options: dict
-        :rtype: dict
+        :param user_options: the template options to use in rendering
+        :type user_options: dict
```
tamarrow: :rtype: None

PR #738, opened 2016-08-23
handle darwin bug with PYVENV_LAUNCHER env var

tamarrow:
unset env var to specify correct virtualenv and override: https://github.com/pypa/setuptools/blob/ac325f34b73c534fd19204400afa598bf63ea301/setuptools/command/easy_install.py#L1940

### DIFF
--- a/dcos/subcommand.py
+++ b/dcos/subcommand.py
@@ -543,20 +543,22 @@ def _install_with_pip(
             for line in requirements:
                 print(line, file=requirements_file)
 
-        cmd = [
-            os.path.join(env_directory, BIN_DIRECTORY, 'pip'),
-            'install',
-            '--requirement',
-            requirement_path,
-        ]
+        # override setuptools finding executable
+        # so we can specify different path
+        with util.set_env('__PYVENV_LAUNCHER__', None):
+            cmd = [
+                os.path.join(env_directory, BIN_DIRECTORY, 'pip'),
+                'install',
+                '--requirement',
+                requirement_path,
+            ]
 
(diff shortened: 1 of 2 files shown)


dcos/util.py
```diff
@@ -35,6 +35,31 @@ def get_logger(name):
 
 
 @contextlib.contextmanager
+def set_env(env_var, new_value):
+    """A context manager for temporary updating env vars
+
+    :param env_var: name of environment variable to alter, None to remove
+    :type env_var: str | None
```
jsancio: The type should be `str`.

dcos/util.py
```diff
@@ -35,6 +35,31 @@ def get_logger(name):
...
+    :param new_val: new value to change env_var to
+    :type env_var: str
```
jsancio: `new_val` is the parameter name. `str | None` for the type.

dcos/subcommand.py
```diff
@@ -543,20 +543,20 @@ def _install_with_pip(
...
-        ]
-
-        if _execute_command(cmd)[2] != 0:
-            # We should remove the directory that we just created
-            if new_package_dir:
-                shutil.rmtree(env_directory)
-
-            raise _generic_error(package_name)
-
+        with util.set_env('__PYVENV_LAUNCHER__', None):
```
jsancio: Let's write a comment here as to why we are doing this. In the future we can put the `setuptools` GitHub issue when we have it.

jsancio:
LGTM. :ship: it

PR #739, opened 2016-08-23
unset __PYVENV_LAUNCHER__ for all commands run outside of main virtualenv

tamarrow:


jsancio:
This doesn't feel right. We should avoid this kind of code duplication. This write a method that does this.

jsancio:
Instead of modifying our own environment variables before forking/executing a command. We should just modify the environment variable for forked process. E.g. https://docs.python.org/3.5/library/subprocess.html#subprocess.Popen

jsancio:
Can you please explain what we need this change?

tamarrow:
@jsancio setuptools uses `__PYVENV_LAUNCHER__` to override the path to the executable. We want to make sure that the path we specify will be respected.

jsancio:
> @jsancio setuptools uses __PYVENV_LAUNCHER__ to override the path to the executable. We want to make sure that the path we specify will be respected.

My question is why do we need this for every single fork/exec operation? For example, we are changing the environment variable when calling `dcos <subcommand> --info`. Is that correct?

jsancio:
I think I understand. If we want to do this for **all** forks/execs then let's make it explicit in the code and not copy this environment change all over the code.

tamarrow:
Okay, I'll update the execs to use a decorator that will modify the environment.

jsancio:
LGTM! :shipit:

PR #743, opened 2016-08-26
Add `--cli` flag for `dcos package list`.

### LINKED ISSUE #626: Support `--cli` in `dcos package list`?

### What version of the DCOS CLI are you using (`dcos --version`)?
git @ d2d7acf3

### What version of DCOS are you using?
Would have been 1.7, but the server is disconnected at this point.

Ubuntu 15.10

### What did you do?
If possible please provide a recipe for reproducing.

I tried to view a list of the installed CLI submodules when no longer connected to a working DC/OS cluster.
a. `dcos package list` -> URL [oldurl] is unreachable: HTTPConnectionPool... Name or service not known
b. `dcos package list --cli` -> Command not recognized

I would've expected `dcos package list --cli` to show a list of custom-installed local client packages from previous `dcos package install`s, without attempting to check the server.

### What did you see instead?
Command not recognized

cruhland:
When the flag is provided, only the packages with subcommands are
listed. App instance information for those packages is still obtained
from the cluster.

~~Partially addresses #626. Let me know if I should extend this to support the offline use case.~~

Fixes #626.


cli/dcoscli/package/main.py
```diff
@@ -421,7 +423,7 @@ def _list(json_, app_id, package_name):
     if app_id is not None:
         app_id = util.normalize_app_id(app_id)
     results = package.installed_packages(
-        package_manager, app_id, package_name)
+        package_manager, app_id, package_name, not cli_only)
 
     # only emit those packages that match the provided package_name and app_id
     if results or json_:
```
cruhland: Question about the error message printed when there are no results (expand the code below): the message is a bit inaccurate, because if flags like `--cli` or `--app-id` are provided, there could be installed packages that aren't shown. Would it be worthwhile to have a different message when results are filtered out due to flags?

dcos/package.py
```diff
@@ -130,7 +80,7 @@ def _matches_package_name(name, command_name):
     return name is None or command_name == name
 
 
-def installed_packages(package_manager, app_id, package_name):
+def installed_packages(package_manager, app_id, package_name, cli_optional):
```
tamarrow: Lets change cli_optional to cli_only to be consistent

tamarrow:
@cruhland I think we should address the offline cause because as is now a user can use `jq` to see only the cli commands using `--json` or just eyeball the table view. We write the `package.json` in `~/.dcos/subcommands/package-name/package.json` so we already have all the information we need anyways :)

cli/dcoscli/package/main.py
tamarrow: probably, nice catch!

cruhland:
@tamarrow updated!

dcos/package.py
```diff
@@ -179,3 +127,45 @@ def installed_subcommands():
...
+    :type apps: [dict]
+    :param subcommands: information on each subcommand installed locally; must
+        have a 'name' key
+    :type subcommands: [dict]
+    :param app_required: if True, only returns elements that have an app
+    :type app_required: bool
+    :param cli_required: if True, only returns elements that have a subcommand
+    :type cli_required: bool
+    :returns: the resulting merged collection, with one element per package
+    :rtype: [dict]
```
tamarrow: [{}}

dcos/package.py
```diff
@@ -179,3 +127,45 @@ def installed_subcommands():
...
+    """
+
+    indexed_apps = {}
+    grouped_apps = itertools.groupby(apps, key=lambda app: app['name'])
+    for name, app_group in grouped_apps:
+        app_list = list(app_group)
+        pkg = app_list[0]
+        pkg['apps'] = [app['appId'] for app in app_list]
```
tamarrow: add sorted

cruhland:
@tamarrow I removed Chronos from the test, but it still takes a while...looks like Kafka is to blame.

With Chronos:
```
real    1m34.563s
user    0m19.597s
sys     0m1.693s
```

Without Chronos:
```
real    1m11.175s
user    0m23.550s
sys     0m2.077s
```

tamarrow:
@cruhland just realized that this call: https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/package/main.py#L420 queries the cluster for cosmos- so we need to update that method for `package list --cli` to work offline

cli/tests/integrations/test_package.py
```diff
@@ -665,11 +669,37 @@ def test_list_cli():
...
+def test_list_cli_flag():
+    kafka_path = 'tests/data/package/json/test_list_kafka_entry.json'
+    kafka_json = file_json_ast(kafka_path)
+    helloworld_path = 'tests/data/package/json/test_list_helloworld_entry.json'
+    helloworld_json = file_json_ast(helloworld_path)
+
+    def assert_test_case(args, expected_packages):
+        stdout = _list_stdout(args=['--json', '--cli'] + args)
+        actual_packages = json.loads(stdout.decode('utf-8'))
+        assert_same_elements(expected_packages, actual_packages)
```
cruhland: @tamarrow here's why I changed `_list()` to return `stdout` in my original PR: I needed to parse the JSON output. So I've created the `_list_stdout()` method to do that rather than change the behavior of `_list()`.

cruhland:
@tamarrow rebased and ready to go!

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,37 @@ def test_list_cli():
 
     stdout = file_json(
         'tests/data/package/json/test_list_helloworld_cli.json')
-    _list(stdout=stdout)
+    _list(args=['--json'], stdout=stdout)
 
     _uninstall_cli_helloworld()
 
 
+def test_list_cli_flag():
```
tamarrow: `test_list_cli_only`

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,37 @@ def test_list_cli():
...
+    kafka_path = 'tests/data/package/json/test_list_kafka_entry.json'
+    kafka_json = file_json_ast(kafka_path)
+    helloworld_path = 'tests/data/package/json/test_list_helloworld_entry.json'
```
tamarrow: lets just take the first element from`cli/tests/data/package/json/test_list_helloworld_cli.json`

cli/tests/integrations/test_package.py
tamarrow: This method is very specific to your use case and is only called once. Can we use exec_command, and just get out the stdout from there there? As is it's hard to know what this method is actually executing.

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,37 @@ def test_list_cli():
...
+    helloworld_json = file_json_ast(helloworld_path)
+
+    def assert_test_case(args, expected_packages):
+        stdout = _list_stdout(args=['--json', '--cli'] + args)
+        actual_packages = json.loads(stdout.decode('utf-8'))
+        assert_same_elements(expected_packages, actual_packages)
+
+    with _kafka():
```
tamarrow: lets remove the kafka check, this should be covered by the unit tests.

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,37 @@ def test_list_cli():
 
     stdout = file_json(
         'tests/data/package/json/test_list_helloworld_cli.json')
-    _list(stdout=stdout)
+    _list(args=['--json'], stdout=stdout)
 
     _uninstall_cli_helloworld()
```
tamarrow: lets add a test to verify that `dcos package list --cli` works offline

cruhland: TODO: Update config to a bogus URL

cruhland:
@tamarrow updated

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,35 @@ def test_list_cli():
...
 
 
+def test_list_cli_only():
+    helloworld_path = 'tests/data/package/json/test_list_helloworld_cli.json'
+    helloworld_json = file_json_ast(helloworld_path)[0]
```
tamarrow: you can actually just leave this as a list, since you only use it as such.

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +670,35 @@ def test_list_cli():
...
+
+    def assert_test_case(args, expected_packages):
```
tamarrow: we should use `assert_command` here instead. We can put in the expected stdout as a param.

cli/tests/integrations/test_package.py
```diff
@@ -941,56 +966,95 @@ def _chronos_package(
...
+
+
+def _helloworld_cli():
+    return _package(name='helloworld',
+                    args=['--yes', '--cli'],
+                    stdout=HELLOWORLD_CLI_STDOUT,
+                    uninstall_stderr=b'')
+
+
+def _kafka():
```
tamarrow: this is not used anymore

cli/tests/integrations/test_package.py
```diff
@@ -667,11 +679,29 @@ def test_list_cli():
...
+    _list(args=['--json'], stdout=stdout)
 
     _uninstall_cli_helloworld()
 
 
+def test_list_cli_only():
+    helloworld_path = 'tests/data/package/json/test_list_helloworld_cli.json'
+    helloworld_json = file_json(helloworld_path)
+
+    def assert_test_case(args, expected_stdout):
```
tamarrow: this function seems unnecessary

cli/tests/integrations/test_package.py
```diff
@@ -33,6 +33,15 @@ def setup_module(module):
         ['dcos', 'package', 'repo', 'add', 'test-universe', UNIVERSE_TEST_REPO]
     )
 
+    describe_command = ['dcos', 'package', 'describe', 'helloworld']
+    for _ in range(10):
```
tamarrow: lets add a comment as to why we're doing this

PR #744, opened 2016-08-26
Don't change the case of the username

jsancio:
We should take the username typed by the user as is. No need to lower
the case of the string.


tamarrow:
LGTM. ill merge once tests pass

Issue #746, opened 2016-08-29
The binary versions of the DC/OS CLI are affected by PyInstaller LD_LIBRARY_PATH issues:  #625, #1759


vishnu2kmohan:
Please answer the following questions before submitting your issue. Thanks!

### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?
If DC/OS CLI version < 0.4.5, DC/OS version can be found at `http://dcos_url/dcos-metada/dcos-version.json`

I first noticed that the latest version of the DC/OS Binary CLI `﻿⁠⁠⁠⁠0.4.11﻿⁠⁠⁠⁠` broke `﻿⁠⁠⁠⁠dcos node ssh`﻿⁠⁠⁠⁠:
```
$ dcos --version
dcoscli.version=0.4.11
dcos.version=1.8-dev
dcos.commit=9a34cc403b2ccdf96d8b6b65d2490d08ea271595
dcos.bootstrap-id=4526c8ec85de022ee8a559e99232505467e36151
$ dcos config show
core.dcos_acs_token=********
core.dcos_url=https://****
core.reporting=False
core.ssl_verify=false
$ dcos node ssh --master-proxy --leader
Running `ssh -A -t core@1.2.3.4 ssh -A -t core@5.6.7.8 `
ssh: symbol lookup error: ssh: undefined symbol: ECDSA_METHOD_set_sign
```
Running the command manually does work when I use the SSH I have on my box:
```
$ ssh -A -t core@1.2.3.4 ssh -A -t core@5.6.7.8
CoreOS stable (1068.9.0)
Last login: Sun Aug 28 18:22:10 2016 from 4.3.2.1
Update Strategy: No Reboots
Failed Units: 2
  locksmithd.service
  update-engine.service
```

I then noticed that it was also broken on DC/OS Binary CLI ﻿⁠⁠⁠⁠`0.4.10﻿⁠⁠⁠⁠`:
```
$ dcos --version
dcoscli.version=0.4.10
dcos.version=1.8-dev
dcos.commit=9a34cc403b2ccdf96d8b6b65d2490d08ea271595
dcos.bootstrap-id=4526c8ec85de022ee8a559e99232505467e36151
$ dcos node ssh --master-proxy --leader           
Running `ssh -A -t core@1.2.3.4 ssh -A -t core@5.6.7.8 `
ssh: symbol lookup error: ssh: undefined symbol: ECDSA_METHOD_set_sign
```
### What operating system and version are you using?
[...]

PR #747, opened 2016-08-31
add Subproc class to handle env change needed for subprocess

tamarrow:
Due to bugs in pyinstaller and setuptools we need to modify the
environment when forking a command.

### DIFF
--- a/dcos/subcommand.py
+++ b/dcos/subcommand.py
@@ -1,6 +1,5 @@
 from __future__ import print_function
 
-import functools
 import hashlib
 import json
 import os
@@ -11,11 +10,12 @@ import subprocess
 import sys
 import zipfile
 from distutils.version import LooseVersion
-from subprocess import PIPE, Popen
+from subprocess import PIPE
 
 import requests
 from dcos import constants, util
 from dcos.errors import DCOSException
+from dcos.subprocess import Subproc
 
 logger = util.get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


tamarrow:
#746

dcos/subcommand.py
```diff
@@ -686,6 +653,69 @@ def package_json(self):
...
+    """Represents a wrapper for subprocess
+    """
+
+    def __init__(self):
+        self._env = os.environ.copy()
+        lib_path = 'LD_LIBRARY_PATH'
+        lib_path_value = self._env.get(lib_path + '_ORIG')
+        # remove pyinstaller overriding `LD_LIBRARY_PATH`
+        # remvoe with stable release of fix:
+        # https://github.com/pyinstaller/pyinstaller/pull/2148
```
jsancio: Let's file an issue to remove the work around once that is fixed in pyinstaller. What release should we target?

dcos/subcommand.py
```diff
@@ -686,6 +653,69 @@ def package_json(self):
...
+        if lib_path_value is not None:
+            self._env[lib_path] = lib_path_value
+        elif self._env.get(lib_path):
+            del self._env[lib_path]
+
+        # Setuptools overrides path to executable from virtualenv,
+        # modify this so we can specify a different path
```
jsancio: Did we end up filing an issue against `setuptools` or asking them about the problem?

dcos/subcommand.py
```diff
@@ -686,6 +653,69 @@ def package_json(self):
             return util.load_json(package_json_file)
 
 
+class Subproc():
```
jsancio: We should move this out of the `subcommand` module. Are we using this for _all_ exec/fork now?

tamarrow: yes

dcos/subcommand.py
tamarrow: yes https://github.com/pypa/setuptools/issues/769

dcos/subcommand.py
tamarrow: created #749. I targeted `5.0` which should happen with DCOS 1.10 release.

PR #748, opened 2016-08-31
Pods integration tests

cruhland:
I've confirmed that the code runs to completion, but of course all tests currently fail.


cli/tests/data/marathon/pods/doubleplusgood.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    {
+      "name": "thing-2",
+      "resources": {
+        "cpus": 0.1,
+        "mem": 16.0
+      }
+    },
+    {
+      "name": "the-cat",
```
tamarrow: nice naming scheme :)

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
+import contextlib
+import json
+import re
+
+from .common import exec_command, file_bytes, file_json
+
+FILE_PATH_BASE = 'tests/data/marathon/pods/'
+
+GOOD_POD_ID = 'good-pod'
+GOOD_POD_FILE_PATH = FILE_PATH_BASE + 'good.json'
```
tamarrow: lets use `os.path.join` for these

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+_POD_ADD_CMD = _POD_BASE_CMD + ['add']
+_POD_LIST_CMD = _POD_BASE_CMD + ['list']
+_POD_REMOVE_CMD = _POD_BASE_CMD + ['remove']
+_POD_SHOW_CMD = _POD_BASE_CMD + ['show']
+_POD_UPDATE_CMD = _POD_BASE_CMD + ['update']
+
+
+def _pod_add_from_file(file_path):
+    cmd = _POD_ADD_CMD + [file_path]
+    exit_status, stdout, stderr = exec_command(cmd)
```
tamarrow: lets check the value of stdout and stderr

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def _pod_add_from_stdin(file_path):
+    cmd = _POD_ADD_CMD
+    exit_status, stdout, stderr = _exec_command_with_stdin(cmd, file_path)
+    return exit_status
+
+
+def _pod_list(json):
+    cmd = _POD_LIST_CMD + (['--json'] if json else [])
```
tamarrow: how about changing the param for this method to`args=[]` and then you can call `_pod_list(['--json'])`, that way you have to add the extra command explicitly

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+    exit_status, stdout, stderr = exec_command(cmd)
+    return exit_status, stdout
+
+
+def _pod_remove(pod_id, force):
```
tamarrow: I would make s/force/args or extra_args. If the test is testing the command with more flags, they should be more explicit.

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+
+    with _pod(GOOD_POD_ID, GOOD_POD_FILE_PATH), \
+            _pod(DOUBLE_POD_ID, DOUBLE_POD_FILE_PATH), \
+            _pod(TRIPLE_POD_ID, TRIPLE_POD_FILE_PATH):
+
+        exit_status, stdout = _pod_list(json=True)
+        assert exit_status == 0
+
+        parsed_stdout = json.loads(stdout.decode('utf-8'))
+        _assert_same_elements(parsed_stdout, expected_json)
```
tamarrow: how about sort the two lists (or make them sets) and then compare with ==

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+
+def test_pod_add_from_file_then_remove():
+    exit_status = _pod_add_from_file(GOOD_POD_FILE_PATH)
+    assert exit_status == 0
+
+    exit_status = _pod_remove(GOOD_POD_ID, force=False)
+    assert exit_status == 0
+
+
+def test_pod_add_from_stdin_then_force_remove():
```
tamarrow: this should be two tests

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+UPDATED_GOOD_POD_FILE_PATH = FILE_PATH_BASE + 'updated_good.json'
+
+DOUBLE_POD_ID = 'double-pod'
+DOUBLE_POD_FILE_PATH = FILE_PATH_BASE + 'double.json'
+
+TRIPLE_POD_ID = 'winston'
+TRIPLE_POD_FILE_PATH = FILE_PATH_BASE + 'doubleplusgood.json'
+
+
+def test_pod_add_from_file_then_remove():
```
tamarrow: this should be two tests

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+
+def _pod_show(pod_id):
+    cmd = _POD_SHOW_CMD + [pod_id]
+    exit_status, stdout, stderr = exec_command(cmd)
+    return exit_status, stdout
+
+
+def _pod_update_from_file(pod_id, file_path, force):
+    cmd = _POD_UPDATE_CMD + _force_args(force) + [pod_id, file_path]
+    exit_status, stdout, stderr = exec_command(cmd)
```
tamarrow: we should use `assert_command` and take in overrides for returncode,  stdout, stderr

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+DOUBLE_POD_ID = 'double-pod'
+DOUBLE_POD_FILE_PATH = FILE_PATH_BASE + 'double.json'
+
+TRIPLE_POD_ID = 'winston'
+TRIPLE_POD_FILE_PATH = FILE_PATH_BASE + 'doubleplusgood.json'
+
+
+def test_pod_add_from_file_then_remove():
+    exit_status = _pod_add_from_file(GOOD_POD_FILE_PATH)
+    assert exit_status == 0
```
tamarrow: lets us returncode to be consistent with the rest of the code base

cli/tests/integrations/test_marathon_pods.py
```diff
@@ -0,0 +1,173 @@
...
+    return exit_status, stdout
+
+
+def _pod_update_from_file(pod_id, file_path, force):
+    cmd = _POD_UPDATE_CMD + _force_args(force) + [pod_id, file_path]
+    exit_status, stdout, stderr = exec_command(cmd)
+    return exit_status, stdout
+
+
+def _pod_update_from_stdin(pod_id, file_path, force):
```
tamarrow: instead of returning, this should check the return params

tamarrow:
@cruhland looks good. my comments mostly revolve around how we are checking the values returned by exec-ing a command, and how we are constructing the command for readability. Let me know if you have questions.

cli/tests/integrations/test_marathon_pods.py
cruhland: In this case, it's not straightforward in Python because the lists contain dicts. So I can't turn the lists into sets because dicts aren't hashable, and sorting is tricky because dict entries are unordered. I could (_recursively_!) convert the dicts into lists of key-value tuples sorted by key, but at that point `assert_same_elements` is simpler.

If you were wondering why I added this method in my previous PR, now you know :stuck_out_tongue_closed_eyes:

cli/tests/integrations/test_marathon_pods.py
cruhland: Could you be more specific? I need to add a pod anyway in order to test removal.

cli/tests/integrations/test_marathon_pods.py
cruhland: Could you be more specific? I need to add a pod anyway in order to test removal.

cruhland:
@tamarrow updated! I replied to some of your comments that I wasn't able to address.

PR #752, opened 2016-09-05
execute janitor clean when a package is uninstalled

anistal:
#### Description
With the purpose of destroy volumes, unreserve resources and clean ZK paths, this feature adds a new flag --force to the uninstall command. The logic is the same that the janitor script: https://docs.mesosphere.com/1.7/usage/managing-services/uninstall/#framework-cleaner

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -1,16 +1,19 @@
 import json
 import os
 import sys
+from urllib.parse import urljoin
 
 import dcoscli
 import docopt
 import pkg_resources
 from dcos import (cmds, config, cosmospackage, emitting, http, options,
-                  package, subcommand, util)
+                  package, subcommand, util, janitorpackage)
 from dcos.errors import DCOSException
 from dcoscli import tables
 from dcoscli.subcommand import default_command_info, default_doc
 from dcoscli.util import decorate_docopt_usage
+
+import requests
 
 logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
@@ -87,7 +90,7 @@ def _cmds():
 
         cmds.Command(
             hierarchy=['package', 'uninstall'],
-            arg_keys=['<package-name>', '--all', '--app-id', '--cli', '--app'],
+            arg_keys=['<package-name>', '--all', '--app-id', '--cli', '--app', '--force'],
             function=_uninstall),
 
         cmds.Command(
@@ -488,7 +491,7 @@ def _search(json_, query):
     return 0
 
 
-def _uninstall(package_name, remove_all, app_id, cli, app):
+def _uninstall(package_name, remove_all, app_id, cli, app, force):
     """Uninstall the specified package.
 
     :param package_name: The package to uninstall
@@ -502,8 +505,21 @@ def _uninstall(package_name, remove_all, app_id, cli, app):
     """
 
     package_manager = _get_package_manager()
-    err = package.uninstall(
-        package_manager, package_name, remove_all, app_id, cli, app)
+
+    if(force == True):
+        toml_config = config.get_config()
+        dcos_url = config.get_config_val("core.dcos_url", toml_config)
+        master_url = "{0}:5050/master".format(dcos_url)
+        exhibitor_url = "{0}:8181".format(dcos_url)
+
+        janitor = janitorpackage.Janitor(package_manager, emitter)
+
+        janitor.destroy_volumes(master_url, "{0}-role".format(package_name), "{0}-principal".format(package_name))
+        janitor.unreserve_resources(master_url, "{0}-role".format(package_name), "{0}-principal".format(package_name))
+        janitor.delete_zk_node(exhibitor_url, "dcos-service-{0}".format(package_name))
+
+    err = package.uninstall(package_manager, package_name, remove_all, app_id, cli, app)
+
     if err is not None:
         emitter.publish(err)
         return 1
--- a/dcos/janitorpackage.py
+++ b/dcos/janitorpackage.py
@@ -1,0 +1,176 @@
+import json
+from urllib.parse import urljoin
+
+import requests
+from dcos import (http)
+
+
+class Janitor():
+    """Implementation of janitor"""
+
+    def __init__(self, package_manager, emitter):
+        self.package_manager = package_manager
+        self.emitter = emitter
+
+    def destroy_volumes(self, master_url, role, principal):
+        state = self.get_state_json(master_url)
+        if not state or not 'slaves' in state.keys():
+            return False
+        all_success = True
+        for slave in state['slaves']:
+            if not self.destroy_volume(slave, master_url, role, principal):
+                all_success = False
+        return all_success
+
+    def destroy_volume(self, slave, master_url, role, principal):
+        volumes = []
+        slaveId = slave['id']
+
+        reserved_resources_full = slave.get('reserved_resources_full', None)
+        if not reserved_resources_full:
+            self.emitter.publish('No reserved resources for any role on slave {}'.format(slaveId))
+            return True
+
+        reserved_resources = reserved_resources_full.get(role, None)
+        if not reserved_resources:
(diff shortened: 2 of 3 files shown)


nickbp:
Would it make sense for this to be a separate `cleanup` command rather than a feature on the `uninstall` command? Then it can be performed following any previous uninstall.

cli/dcoscli/package/main.py
```diff
@@ -502,8 +505,21 @@ def _uninstall(package_name, remove_all, app_id, cli, app):
...
+
+    if(force == True):
+        toml_config = config.get_config()
+        dcos_url = config.get_config_val("core.dcos_url", toml_config)
+        master_url = "{0}:5050/master".format(dcos_url)
+        exhibitor_url = "{0}:8181".format(dcos_url)
+
+        janitor = janitorpackage.Janitor(package_manager, emitter)
+
+        janitor.destroy_volumes(master_url, "{0}-role".format(package_name), "{0}-principal".format(package_name))
```
nickbp: The default values are dependent on the service name that the user chose. For example if a user named their Cassandra service `foo` at install then these would be `foo-role`/`foo-principal`/`dcos-service-foo`. Therefore it'd make sense for this to be a parameter that the user is required to provide.

cli/dcoscli/package/main.py
```diff
@@ -502,8 +505,21 @@ def _uninstall(package_name, remove_all, app_id, cli, app):
...
+        janitor.unreserve_resources(master_url, "{0}-role".format(package_name), "{0}-principal".format(package_name))
```
nickbp: The principal and role may be customized by the user (beyond just the above variance based on the service name), so it should be possible for the user to optionally specify any custom role/principal when running the command.

dcos/janitorpackage.py
```diff
@@ -0,0 +1,176 @@
...
+            return response.json()
+        except requests.exceptions.Timeout:
+            self.emitter.publish('HTTP GET request timed out.')
+        except requests.exceptions.ConnectionError as err:
+            self.emitter.publish('Network error:', err)
+        except requests.exceptions.HTTPError as err:
+            self.emitter.publish('Invalid HTTP response:', err)
+        return None
+
+    def extract_version_num(self, slaves_json):
```
nickbp: This has a bug with `1.0.0` => 0

dcos/janitorpackage.py
```diff
@@ -0,0 +1,176 @@
...
+
+import requests
+from dcos import (http)
+
+
+class Janitor():
+    """Implementation of janitor"""
+
+    def __init__(self, package_manager, emitter):
+        self.package_manager = package_manager
```
nickbp: It doesn't look like anything is using this member?

dcos/janitorpackage.py
```diff
@@ -0,0 +1,176 @@
...
+            return False
+
+        if 200 <= response.status_code < 300:
+            self.emitter.publish('Successfully deleted znode \'{}\' (code={}), if znode existed.'.format(
+                znode, response.status_code))
+            return True
+        else:
+            self.emitter.publish('ERROR: HTTP DELETE request returned code:', response.status_code)
+            self.emitter.publish('Response body:', response.text)
+            return False
```
nickbp: Missing trailing newline

cli/dcoscli/package/main.py
```diff
@@ -502,8 +505,21 @@ def _uninstall(package_name, remove_all, app_id, cli, app):
     """
 
     package_manager = _get_package_manager()
-    err = package.uninstall(
-        package_manager, package_name, remove_all, app_id, cli, app)
+
+    if(force == True):
+        toml_config = config.get_config()
+        dcos_url = config.get_config_val("core.dcos_url", toml_config)
+        master_url = "{0}:5050/master".format(dcos_url)
```
nickbp: This assumes that the master may be reached directly from the CLI machine

cli/dcoscli/package/main.py
```diff
@@ -1,17 +1,20 @@
...
 import pkg_resources
 from dcos import (cmds, config, cosmospackage, emitting, http, options,
-                  package, subcommand, util)
+                  package, subcommand, util, janitorpackage)
 from dcos.errors import DCOSException
 from dcoscli import tables
 from dcoscli.subcommand import default_command_info, default_doc
 from dcoscli.util import decorate_docopt_usage
 
+import requests
```
nickbp: This package isn't used in this file?

nickbp:
Hey, thanks for the PR! This has been a pain point for a while.

I think it'd make more sense for the cleanup logic to instead be implemented as part of the per-service CLI modules, i.e. the code that's invoked when [`dcos cassandra ...`](https://github.com/mesosphere/dcos-cassandra-service/blob/master/cli/dcos-cassandra/main.go) or [`dcos kafka ...`](https://github.com/mesosphere/dcos-kafka-service/blob/master/cli/dcos-kafka/main.go) commands are run by the user.
The reason for doing it in the per-service CLI modules is because the main DC/OS CLI can't really assume where individual services have placed their data, ie the role/principal/zk node being deleted here. The assumptions in this PR around `x-role`/`x-principal`/`dcos-service-x` is specific to a handful of services. But those services all use the same common code in [dcos-commons](https://github.com/mesosphere/dcos-commons/) for handling these resources. So the cleanup operation could be put in dcos-commons' common CLI code.

The logic could go like this (using cassandra as an example, but this would apply to kafka/hdfs as well):
- User calls `dcos cassandra uninstall [service-name]` (eg `dcos cassandra uninstall my-cassandra`)
- Command [invokes](https://github.com/mesosphere/dcos-commons/blob/master/cli/runcli.go#L12) `dcos package uninstall --app-id=<service-name> --app cassandra` (`--app` ensures that the CLI module isn't uninstalled while it's still running)
- Command invokes Janitor logic (this part is some work -- it'd involve porting the janitor script from Python to Go)
- Assuming the above succeeds, command prints something like: `Uninstall successful! Run "dcos package uninstall --cli cassandra" to remove this command`

The per-service CLI modules use this code as a common library (this is where the main 'uninstall' operation could be implemented): https://github.com/mesosphere/dcos-commons/tree/master/cli
And then each CLI module using dcos-commons could be updated to support that common operation: https://github.com/mesosphere/dcos-cassandra-service/blob/master/cli/dcos-cassandra/main.go and https://github.com/mesosphere/dcos-kafka-service/blob/master/cli/dcos-kafka/main.go

anistal:
Hi @nickbp !

First of all: thanks for your detailed feedback.

I would like to ask some things about your last comment:

- If we move Janitor's logic to dcos-commons and the final uninstall command goes here, a framework that don't extends from dcos-commons will not execute this uninstall proccess, right?. Then it will not work in all cases.
- Could it have sense to follow these steps?: 
  1. If the framework to uninstall has been implemented with dcos-commons, then the uninstall proccess is done following the way that you explained. (executing dcos uninstall <framework> that finally runs dcos-commons uninstall with the janitor logic). This is the most clean case.
  2. For other cases  we can use the --force flag that assumes that principal, role and zk path are the same that the name of the framework. (executing dcos uninstall <framework-name> --force). If for example zk path and role have changed we can pass more parameters (such as dcos uninstall <framework-name> --force --zk-path <zk-path> --role <role>). In this case we change the zk path and the role but principal is the same that the framework name.

Let me know what to you think or if I am saying anything wrong or if I forgot something more.

Thanks in advance!

jsancio:
Hi @anistal,

Excuse the delay response. We are interested in a solution to uninstall that is available to both the CLI and the UI or any script that a DC/OS user may write to interact with the package manager (Cosmos) or service manager (Marathon). Currently DC/OS doesn't have a mechanism for doing this. For DC/OS Packages, we would like for the package maintainer to include a task that would get executed when the Marathon AppDefinition gets removed or deleted. To implement this would need support from Marathon and Jobs (metronome).

We have this feature in our roadmap and will be addressing it in up coming releases.

I am going to close this PR for now.

PR #753, opened 2016-09-06
when adding new config value, only check that section of config

tamarrow:
For example, if you have a spark section in the config, and you uninstall the spark subcommand, trying to add anything to the config will result in a 'spark is not a dcos command' error. 

Instead, if the user is not changing the 'spark' section in the toml, we shouldn't be validating it.

### DIFF
--- a/cli/tests/integrations/test_config.py
+++ b/cli/tests/integrations/test_config.py
@@ -151,6 +151,25 @@ def test_set_new_output(env):
     config_set('core.dcos_url', 'http://dcos.snakeoil.mesosphere.com', env)
 
 
+def test_set_nonexistent_subcommand(env):
+    assert_command(
+        ['dcos', 'config', 'set', 'foo.bar', 'baz'],
+        stdout=b'',
+        stderr=b"'foo' is not a dcos command.\n",
+        returncode=1,
+        env=env)
+
+
+def test_set_when_extra_section():
+    env = os.environ.copy()
+    path = os.path.join('tests', 'data', 'config', 'invalid_section.toml')
+    env['DCOS_CONFIG'] = path
+    os.chmod(path, 0o600)
+
+    config_set('core.dcos_url', 'http://dcos.snakeoil.mesosphere.com', env)
+    config_unset('core.dcos_url', env)
+
+
 def test_unset_property(env):
     config_unset('core.reporting', env)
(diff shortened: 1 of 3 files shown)


jsancio:
Can we write tests for this? We may already have a test for this but it would be nice to see what happens if the user specifies a section that doesn't exists.

tamarrow:
added a test!

cli/tests/integrations/test_config.py
```diff
@@ -151,6 +151,15 @@ def test_set_new_output(env):
     config_set('core.dcos_url', 'http://dcos.snakeoil.mesosphere.com', env)
 
 
+def test_set_nonexistent_subcommand(env):
+    assert_command(
+        ['dcos', 'config', 'set', 'foo.bar', 'baz'],
+        stdout=b'',
+        stderr=b"'foo' is not a dcos command.\n",
+        returncode=1,
+        env=env)
```
jsancio: Great. How about a test for what this bug is actually fixing. The config file has an entry for "missing.something" but the "missing" section/subcommand doesn't exists.

tamarrow: added

cli/tests/integrations/test_config.py
```diff
@@ -151,6 +151,25 @@ def test_set_new_output(env):
...
+
+
+def test_set_when_extra_section():
+    env = os.environ.copy()
+    path = os.path.join('tests', 'data', 'config', 'invalid_section.toml')
+    env['DCOS_CONFIG'] = path
+    os.chmod(path, 0o600)
```
jsancio: Can we instead store it in git with the right permission?

cli/tests/integrations/test_config.py
```diff
@@ -295,6 +314,7 @@ def test_timeout(env):
 def test_parse_error():
     env = os.environ.copy()
     path = os.path.join('tests', 'data', 'config', 'parse_error.toml')
+    os.chmod(path, 0o600)
```
jsancio: Can we instead store it in git with the right permission?

tamarrow: Git doesn't store file permissions (only the executable bit) and I wanted to minimize configuration for running the tests locally.

jsancio:
LGTM! :shipit:

PR #761, opened 2016-09-12
Implement `pod remove`

cruhland:


dcos/marathon.py
```diff
@@ -747,3 +757,19 @@ def _default_marathon_error(message=""):
     return ("Marathon likely misconfigured. Please check your proxy or "
             "Marathon URL settings. See dcos config --help. {}").format(
                 message)
+
+
+def _normalize_marathon_id_path(id_path):
```
tamarrow: this should be in the marathon class

dcos/marathon.py
```diff
@@ -733,8 +722,29 @@ def get_leader(self):
...
+        """Completely removes the requested pod.
+
+        :param pod_id: the ID of the pod to remove
+        :type pod_id: str
+        :param force: whether to override running deployments
+        :type force: bool
+        :rtype: None
+        """
+
+        path = 'v2/pods' + _normalize_marathon_id_path(pod_id)
```
tamarrow: lets use os.path.join here

cli/tests/unit/test_marathon_pod.py
```diff
@@ -40,3 +53,28 @@ def resource_reader(path):
         subcmd.pod_add('does/not/matter')
 
     assert exception_info.value == exception
+
+
+def _assert_pod_remove_invoked_successfully(pod_id, force):
+    def resource_reader(path):
+        pass
```
tamarrow: instead of passing, can we throw an exception to make sure this isn't doing anything?

dcos/marathon.py
cruhland: I actually found another implementation of this in `dcos.util`, so I moved this implementation there, and updated all usages of it.

dcos/marathon.py
cruhland: How about `urllib.parse.urljoin()`, since we're combining URL paths, not filesystem paths?

cli/tests/unit/test_marathon_pod.py
cruhland: Will do

dcos/marathon.py
cruhland: Argh, this is frustrating! The normalized path ID has a leading slash, which causes problems:

```python
os.path.join('v2/pods', '/foo/bar') == '/foo/bar'
os.path.join('v2/pods/', '/foo/bar') == '/foo/bar'
urllib.parse.urljoin('v2/pods', '/foo/bar') == '/foo/bar'
urllib.parse.urljoin('v2/pods/', '/foo/bar') == '/foo/bar'
```

I might just write a custom method for this.

cruhland:
@tamarrow updated! Please take a look at the commit diff, I made some non-trivial changes.

cruhland:
@tamarrow added another commit that puts an assertion in the `resource_reader` as you requested.

cli/tests/unit/test_marathon_pod.py
```diff
@@ -57,7 +57,7 @@ def resource_reader(path):
 
 def _assert_pod_remove_invoked_successfully(pod_id, force):
     def resource_reader(path):
-        pass
+        assert False, "should not be called"
```
tamarrow: lets use parens to be consistent with the rest of the code

dcos/marathon.py
```diff
@@ -741,10 +743,30 @@ def remove_pod(self, pod_id, force=False):
...
         params = {'force': 'true'} if force else None
         self._rpc.http_req(http.delete, path, params=params)
 
+    @staticmethod
+    def _marathon_id_path_join(url_path, id_path):
+        """Concatenates a URL path with a Marathon "ID path", ensuring the
+        result is well-formed.
+
+        The path and the ID will be joined with a single forward slash (/),
+        all trailing slashes in the ID will be removed, and the ID will have
```
tamarrow: we explicitly want the trailing slash for functionality with admin router.

dcos/marathon.py
```diff
@@ -741,10 +743,30 @@ def remove_pod(self, pod_id, force=False):
...
+        :param url_path: the path portion of a URL
+        :type url_path: str
+        :param path_id: a Marathon "ID path", e.g. app ID or group ID
+        :type path_id: str
+        :returns: the path with the ID appended
+        :rtype: str
+        """
+
+        normalized_id_path = urllib.parse.quote(id_path.strip('/'))
+        return posixpath.join(url_path, normalized_id_path)
```
tamarrow: will this work on windows?

dcos/util.py
```diff
@@ -618,16 +618,20 @@ def get_ssh_options(config_file, options):
...
+    urllib.parse.quote().
+
+    :param id_path
+    :type id_path: str
+    :returns: normalized path
     :rtype: str
     """
 
-    return urllib.parse.quote('/' + app_id.strip('/'))
+    return urllib.parse.quote('/' + id_path.strip('/'))
```
tamarrow: If we rename a library function we should explicitly call it out in the commit message.

dcos/marathon.py
cruhland: Could you give an example? Looking at the existing code I don't see anything ensuring trailing slashes are present. Did you mean a _leading_ slash, i.e. we want `/foo/bar` instead of `foo/bar`?

tamarrow: see #718, JIRA: DCOS-6961

dcos/marathon.py
cruhland: Yep, see the second "Note" in the [`os.path` docs](https://docs.python.org/3.4/library/os.path.html).

dcos/util.py
cruhland: Ok, I'll make sure it's in the message for the squashed commit.

dcos/marathon.py
cruhland: Ah, notice in #718 that you added the trailing slash to the Marathon base URL, which always has the endpoint path appended to it afterward anyway. Jan-Philip points out the source of the problem in DCOS-6961: it's the counterintuitive behavior of `urllib.parse.urljoin()`, which will drop the last segment of the first path argument:

```python
urllib.parse.urljoin('v2/pods', 'foo/bar') == 'v2/foo/bar' # !!!
urllib.parse.urljoin('v2/pods', '/foo/bar') == '/foo/bar'
urllib.parse.urljoin('v2/pods/', 'foo/bar') == 'v2/pods/foo/bar'
urllib.parse.urljoin('v2/pods/', '/foo/bar') == '/foo/bar'
```

So (1) it's fine to remove trailing slashes from the ID, because that wasn't the cause of DCOS-6961, and (2) the correct behavior will be preserved by this, because `posixpath.join()` is being used instead.

There's a few test cases in this PR which cover exactly this behavior, but if you think there's a gap I'm happy to add more :smile:

dcos/marathon.py
tamarrow: Based on the note, `os.path` works for all platforms, and `posixpath` only for unix:
```
The os.path module is always the path module suitable for the operating system Python is running on, and therefore usable for local paths. However, you can also import and use the individual modules if you want to manipulate a path that is always in one of the different formats. They all have the same interface:
posixpath for UNIX-style paths
ntpath for Windows paths
macpath for old-style MacOS paths
```

dcos/util.py
tamarrow: thanks. we're trying to make breaking changes more obvious

cli/tests/unit/test_marathon_pod.py
cruhland: If I do that, I get this error from the syntax tests:

```
tests/unit/test_marathon_pod.py:60:9: F999 assertion is always true, perhaps remove parentheses?
tests/unit/test_marathon_pod.py:72:9: F999 assertion is always true, perhaps remove parentheses?
flake8.main.application   MainProcess   1770 INFO     Found a total of 12 violations and reported 2
ERROR: InvocationError: '/home/charles/repos/dcos/dcos-cli/cli/.tox/py34-syntax/bin/flake8 --verbose dcoscli tests setup.py'
```

I think `assert` is actually an element of Python syntax, so putting parentheses around its arguments just means we're `assert`ing on a tuple, instead of passing two arguments to `assert`.

Yep: https://docs.python.org/3.4/reference/simple_stmts.html#grammar-token-assert_stmt

dcos/marathon.py
cruhland: Hmm, I read this sentence differently:

> However, you can also import and use the individual modules if you want to manipulate a path that is always in one of the different formats. 

To me, that says _on any platform_, you are able to use the module for plaform X if your program (for whatever reason) needs to manipulate paths for X.

I did write tests to cover this behavior, and they've passed on the Windows build, so I think we're good?

cruhland:
@tamarrow I think I've addressed all of your comments.

dcos/marathon.py
```diff
@@ -765,7 +763,7 @@ def _marathon_id_path_join(url_path, id_path):
         """
 
         normalized_id_path = urllib.parse.quote(id_path.strip('/'))
-        return posixpath.join(url_path, normalized_id_path)
+        return url_path.rstrip('/') + '/' + normalized_id_path
```
tamarrow: 👍

tamarrow:
merge when tests pass

PR #765, opened 2016-09-15
Implement `pod list`

cruhland:


cli/dcoscli/marathon/main.py
```diff
@@ -851,7 +851,10 @@ def pod_list(self, json_):
         :rtype: int
         """
 
-        raise DCOSException('Not implemented')
+        marathon_client = self._create_marathon_client()
+        pods = marathon_client.list_pod()
+        emitter.publish(pods)
```
tamarrow: this should depend on whether `json_` is T/F so it can print a table or `json`

cruhland:
@tamarrow updated!

cruhland:
@tamarrow added new commit that gets the integration test working.

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -1,26 +1,17 @@
...
-GOOD_POD_ID = 'good-pod'
-GOOD_POD_FILE_PATH = os.path.join(FILE_PATH_BASE, 'good.json')
-UPDATED_GOOD_POD_FILE_PATH = os.path.join(FILE_PATH_BASE, 'updated_good.json')
-
-DOUBLE_POD_ID = 'double-pod'
-DOUBLE_POD_FILE_PATH = os.path.join(FILE_PATH_BASE, 'double.json')
-
-TRIPLE_POD_ID = 'winston'
-TRIPLE_POD_FILE_PATH = os.path.join(FILE_PATH_BASE, 'doubleplusgood.json')
+from ..fixtures.marathon import (DOUBLE_POD_FILE_PATH, DOUBLE_POD_ID,
```
cruhland: Well, that's kind of why I moved those constants to `fixtures.marathon`, because that module only defines test data. Should I just do a wildcard import of `fixtures.marathon`? Or do you think I should make a `constants.marathon` module?

tamarrow: I would just import `fixtures.marathon`

cli/tests/unit/test_marathon_pod.py
```diff
@@ -41,6 +43,38 @@ def test_pod_show_propagates_exceptions_from_show_pod():
         Exception('Oops!'))
 
 
+def test_pod_list_success_case():
```
cruhland: Sure, there will be more assertions in a single test method that way, but the test output makes it fairly easy to determine which one failed. I'll combine them.

cruhland:
@tamarrow comments addressed in code, except the one about the constants (see my reply). Sorry for the delay!

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -129,6 +121,7 @@ def _assert_pod_show(pod_id):
 
     pod_json = json.loads(stdout.decode('utf-8'))
     assert pod_json['id'] == util.normalize_marathon_id_path(pod_id)
+    return pod_json
```
tamarrow: This helper is a little confusing since it is named `_assert_x`, but also returns a value.

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -50,16 +41,17 @@ def test_pod_add_from_stdin_then_force_remove():
...
-    paths = [GOOD_POD_FILE_PATH, DOUBLE_POD_FILE_PATH, TRIPLE_POD_FILE_PATH]
-    expected_json = [file_json(path) for path in paths]
     expected_table = file_bytes('tests/unit/data/pod.txt')
 
     with _pod(GOOD_POD_ID, GOOD_POD_FILE_PATH), \
             _pod(DOUBLE_POD_ID, DOUBLE_POD_FILE_PATH), \
             _pod(TRIPLE_POD_ID, TRIPLE_POD_FILE_PATH):
 
+        pod_ids = [GOOD_POD_ID, DOUBLE_POD_ID, TRIPLE_POD_ID]
+        expected_json = [_assert_pod_show(pod_id) for pod_id in pod_ids]
```
tamarrow: we should read expected_json from a file so we are validating a known value.

cruhland:
@tamarrow back to you!

PR #770, opened 2016-09-22
Improved help for dcos node diagnostics create.

joerg84:
The keyword usage was misleading before.

### DIFF
--- a/cli/dcoscli/data/help/node.txt
+++ b/cli/dcoscli/data/help/node.txt
@@ -25,7 +25,7 @@ Commands:
         cluster.
     diagnostics create
         Create a diagnostics bundle. Nodes can be: ip address, hostname, mesos ID
-        or key words "all", "masters", "agents".
+        or keywords "all", "masters", "agents" (notice the quotes around the keywords).
     diagnostics download
         Download a diagnostics bundle.
     diagnostics delete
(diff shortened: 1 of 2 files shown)


cli/dcoscli/data/help/node.txt
```diff
@@ -25,7 +25,7 @@ Commands:
         cluster.
     diagnostics create
         Create a diagnostics bundle. Nodes can be: ip address, hostname, mesos ID
-        or key words "all", "masters", "agents".
+        or key words "all", "masters", "agents" (notice the quotes around the keywords).
```
jsancio: We need to make the same change here: https://github.com/dcos/dcos-cli/blob/master/cli/tests/data/help/node.txt#L26-L28

tamarrow:
the need to update the expected text in the test as well:  https://github.com/dcos/dcos-cli/blob/master/cli/tests/data/help/node.txt

cli/dcoscli/data/help/node.txt
joerg84: Wondered why it appeared twice :-).

joerg84:
Updated, I suppose in case this lands I also need to update: https://dcos.io/docs/1.8/usage/cli/command-reference/. Or is there some automagic update happening?

tamarrow:
@joerg84 no magic :( you have to make a pr against docs for that.

PR #776, opened 2016-09-26
Change 409 response status error message to be more understandable

cruhland:
Per our discussion on Friday @tamarrow. Open to suggestions on the wording.

### DIFF
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -109,9 +109,14 @@ class RpcClient(object):
             return template.format(request_method, request_url, reason, suffix)
 
         if status_code == 409:
-            return ('App, group, or pod already exists and cannot be changed '
-                    'at this time. Try again later or use --force if it is'
-                    'available for your command.')
+            _, _, path, _, _, _ = urllib.parse.urlparse(request_url)
+            for resource_name in ['app', 'group', 'pod']:
+                if '/v2/{}s'.format(resource_name) in path:
+                    break
+            else:
+                resource_name = 'resource'
+
+            return '{} already exists.'.format(resource_name.capitalize())
 
         if json_body is None:
             template = 'Error decoding response from [{}]: HTTP {}: {}'
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -79,7 +79,7 @@ class RpcClient(object):
         self._timeout = timeout
 
     ERROR_JSON_VALIDATOR = jsonschema.Draft4Validator(load_error_json_schema())
-    RESOURCE_NAMES = ['app', 'group', 'pod']
+    RESOURCE_TYPES = ['app', 'group', 'pod']
 
     @classmethod
     def response_error_message(cls, status_code, reason, request_method,
@@ -111,7 +111,7 @@ class RpcClient(object):
 
         if status_code == 409:
             path = urllib.parse.urlparse(request_url).path
-            path_name = (name for name in cls.RESOURCE_NAMES if name in path)
+            path_name = (name for name in cls.RESOURCE_TYPES if name in path)
             resource_name = next(path_name, 'resource')
 
             return '{} already exists.'.format(resource_name.capitalize())
--- a/dcos/marathon.py
+++ b/dcos/marathon.py
@@ -109,8 +109,9 @@ class RpcClient(object):
             return template.format(request_method, request_url, reason, suffix)
 
         if status_code == 409:
-            return ('App, group, or pod is locked by one or more deployments. '
-                    'Override with --force.')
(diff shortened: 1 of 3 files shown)


dcos/marathon.py
```diff
@@ -109,8 +109,9 @@ def response_error_message(cls, status_code, reason, request_method,
             return template.format(request_method, request_url, reason, suffix)
 
         if status_code == 409:
-            return ('App, group, or pod is locked by one or more deployments. '
-                    'Override with --force.')
+            return ('App, group, or pod already exists and cannot be changed '
+                    'at this time. Try again later or use --force if it is'
+                    'available for your command.')
```
tamarrow: Lets keep it simple: `App, group, or pod already exists`. Users shouldnt be using force if they dont know what it'll do. 

We can even be fancier, and since we have the `request_method` tell the user <i>which</i> resource already exists.

tests/test_marathon.py
```diff
@@ -406,16 +406,22 @@ def test_response_error_message_with_400_status_json():
...
-                'Override with --force.')
-    assert actual == expected
+        pattern = r'(.*) already exists\.'
+        _assert_matches_with_groups(pattern, actual, (expected_resource,))
+
+    test_case('http://marathon/v2/apps', 'App')
+    test_case('http://marathon/v2/groups/group-id', 'Group')
+    test_case('http://marathon/v2/pods/', 'Pod')
+    test_case('http://marathon/v2/thingies/foo', 'Resource')
+    test_case('http://dcos/service/marathon/v2/apps/bar', 'App')
```
cruhland: Turns out I already wrote a generic unit test for this error, so I was able to test it for all resource types. Yay unit tests!

cruhland:
@tamarrow strange test failure: https://teamcity.mesosphere.io/viewLog.html?buildId=420010&buildTypeId=DcosIo_DcosCli_Ci&tab=buildLog#_focus=2129

I'll rerun and see if it happens again.

tamarrow:
@cruhland thats from a ccm error. if you look at step 2 + 3 you will see it errored. and at the top of step 4: `raise InvalidURL("Invalid URL %r: No host supplied" % url)`

dcos/marathon.py
```diff
@@ -109,8 +109,14 @@ def response_error_message(cls, status_code, reason, request_method,
             return template.format(request_method, request_url, reason, suffix)
 
         if status_code == 409:
-            return ('App, group, or pod is locked by one or more deployments. '
-                    'Override with --force.')
+            _, _, path, _, _, _ = urllib.parse.urlparse(request_url)
```
tamarrow: `path = urllib.parse.urlparse(request_url).path`

dcos/marathon.py
```diff
@@ -109,8 +109,14 @@ def response_error_message(cls, status_code, reason, request_method,
...
+            for resource_name in ['app', 'group', 'pod']:
+                if '/v2/{}s'.format(resource_name) in path:
```
tamarrow: I think it's good enough if we just look for the resource in the path. We know this is a marathon url and only one resource is in their endpoint names. 

we can then do something like `resource_name = next((resource for resource in ['app', 'group', 'pod'] if resource in path), 'Resource')`

https://docs.python.org/3.4/library/functions.html#next

dcos/marathon.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, base_url, timeout=http.DEFAULT_TIMEOUT):
         self._timeout = timeout
 
     ERROR_JSON_VALIDATOR = jsonschema.Draft4Validator(load_error_json_schema())
+    RESOURCE_NAMES = ['app', 'group', 'pod']
```
tamarrow: s/names/types

dcos/marathon.py
```diff
@@ -109,12 +110,9 @@ def response_error_message(cls, status_code, reason, request_method,
...
         if status_code == 409:
-            _, _, path, _, _, _ = urllib.parse.urlparse(request_url)
-            for resource_name in ['app', 'group', 'pod']:
-                if '/v2/{}s'.format(resource_name) in path:
-                    break
-            else:
-                resource_name = 'resource'
+            path = urllib.parse.urlparse(request_url).path
+            path_name = (name for name in cls.RESOURCE_NAMES if name in path)
+            resource_name = next(path_name, 'resource')
```
tamarrow: i would just put the last two lines in one- no need to try to find all resources in path.

cruhland: I'd have to do some line wrapping if I combined them. The current code is semantically the same -- the [generator expression](https://docs.python.org/3.4/reference/expressions.html#generator-expressions) is lazy.

PR #777, opened 2016-09-26
All pods integration tests pass against pods-enabled Marathon and Mesos

cruhland:
The tests are still marked as skipped to prevent CI breakage.


dcos/marathon.py
```diff
@@ -753,7 +774,14 @@ def update_pod(self, pod_id, pod_json, force=False):
...
         """
 
-        return self._update('pods', pod_id, pod_json, force)
+        response = self._update_req('pods', pod_id, pod_json, force)
+        header_name = 'Marathon-Deployment-Id'
+
+        try:
+            return response.headers[header_name]
+        except KeyError:
+            template = 'Error: missing "{}" header from Marathon response'
```
tamarrow: lets use response.headers.get('header_name') instead of using the try

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -85,14 +88,19 @@ def test_pod_update_from_stdin_force_true():
 
 def _pod_add_from_file(file_path):
     cmd = _POD_ADD_CMD + [file_path]
-    return exec_command(cmd)
+    outputs = exec_command(cmd)
+
+    watch_all_deployments()
+    return outputs
```
tamarrow: lets just exec in this method, and let the `wait` happen in the caller

cli/tests/data/marathon/pods/updated_good.json
```diff
@@ -15,5 +15,10 @@
         "mem": 16.0
       }
     }
+  ],
+  "networks": [
```
tamarrow: do we have a test if a user omits this field?

dcos/marathon.py
cruhland: Any particular reason why? It's about the same length, and Python encourages [EAFP](https://docs.python.org/3.4/glossary.html#term-eafp).

```python
deployment_id = response.headers.get(header_name)
if deployment_id is None:
    template = 'Error: missing "{}" header from Marathon response'
    raise DCOSException(template.format(header_name))

return deployment_id
```

cli/tests/integrations/test_marathon_pod.py
cruhland: Just for this method, `_pod_add_from_file()`? Or everywhere I'm using `watch_all_deployments()`? I'd prefer not to have to remember to call it each time.

cli/tests/data/marathon/pods/updated_good.json
cruhland: Not specifically for this field. There's a unit test that ensures we display an error message when we get a 422 back from Marathon, which is what happens when we omit this field.

I don't want to write tests that rely on the details of Marathon's validation, since they're more likely to be brittle. On the other hand, having an integration test covering the error handling code path might be a good idea, even though it's covered by a unit test already. What do you think?

cruhland:
@tamarrow responded to comments.

cli/tests/integrations/test_marathon_pod.py
tamarrow: just this one. since we're returning all outputs of `exec_command` it alludes that we're doing the checking of the values (and thus `watch_all_deployments`) elsewhere

dcos/marathon.py
tamarrow: just a style thing- we havent been using that style. we only do try/except in cases where we can't guarantee there will be no exceptions (ie reading a file)

cli/tests/data/marathon/pods/updated_good.json
tamarrow: You added a new commit that added just the networks field, so my question is when did you realize this field was necessary? If you realized it because existing tests failed, than no need to add a new one imo.

cruhland: Yeah, the integration tests failed because they expected a 200 status but got a 422.

cruhland:
@tamarrow the commit I just pushed permanently enables the pods integration tests, so I'll link to a custom CI run with a pods-enabled cluster once that seems to be working.

cli/tests/data/marathon/pods/updated_good.json
tamarrow: 👍

tamarrow:
@cruhland can you link to the passing CI? if thats green, then merge away.

cruhland:
@tamarrow some tests are still failing against the pods-enabled DC/OS cluster. Will update once they're all fixed.

cruhland:
@tamarrow getting a strange error with latest changes: https://teamcity.mesosphere.io/viewLog.html?buildId=423675&buildTypeId=DcosIo_DcosCli_Ci&tab=buildLog

The weird thing is that those tests pass locally using the CCM cluster for pods I created earlier today. I'm creating another CCM cluster now to see if anything's changed.

tamarrow:
@cruhland looks like ccm failed to create again :/

cruhland:
@tamarrow progress, but now getting this failure: https://teamcity.mesosphere.io/viewLog.html?buildId=423735&buildTypeId=DcosIo_DcosCli_CiOpen&tab=buildLog#_state=82,74,66&focus=1189

I was able to reproduce locally using a manually created CCM cluster, investigating now.

cruhland:
@tamarrow CI passes! https://teamcity.mesosphere.io/viewLog.html?buildId=424114&

You should review my latest commit though, had to make some non-trivial test changes.

cruhland:
@tamarrow I decided to go ahead with a separate pods CI after all. Sorry! But given there's several additional small PRs I want to make today, not having to put thought into which build failures can be ignored or not seemed like the right way to go.

cruhland:
@tamarrow should be all set!

cli/tests/integrations/test_marathon.py
```diff
@@ -144,16 +144,24 @@ def test_show_relative_app_version():
 
 
 def test_show_missing_relative_app_version():
+    app_id = 'zero-instance-app'
```
tamarrow: this could be a global `ZERO_INSTANCE_APP_ID` since we use it all over this file

cli/tests/integrations/test_marathon.py
```diff
@@ -767,9 +776,20 @@ def _list_versions(app_id, expected_count, max_count=None):
...
 
+    # Marathon persists app versions indefinitely by ID, so there may be extras
+    assert len(result) >= expected_min_count
+
+    if max_count is not None:
+        assert len(result) <= max_count
+
+
+def _assert_matches_with_groups(pattern, text, groups):
+    match = re.fullmatch(pattern, text, flags=re.DOTALL)
```
tamarrow: fullmatch is not supported for 2.7

cli/tests/integrations/test_marathon.py
```diff
@@ -767,9 +776,20 @@ def _list_versions(app_id, expected_count, max_count=None):
...
+    assert match
+    assert match.groups() == groups
```
tamarrow: lets just put the app-id where it should be in the str output

cli/tests/integrations/test_marathon.py
cruhland: Want me to replace all literal occurrences of it?

cli/tests/integrations/test_marathon.py
cruhland: I thought our tests only run using Python 3.4?

cli/tests/integrations/test_marathon.py
cruhland: Yeah, that's less code, although it will be more work to change if `ZERO_INSTANCE_APP_ID` ever changes (unlikely).

cruhland:
@tamarrow partially addressed your comments. Let me know if you want more changes.

cruhland:
:tada:

PR #780, opened 2016-09-28
Improving tox config: TeamCity test output, flake8-import-sort, pep8-naming

cmaloney:
Rolls what was #781, #782 into one big PR since otherwise they all conflict slightly with oneanother when merging.

This adds the teamcity-messages packages which extends py.test and flake8 in
order to make it so messages from the tools will be picked up by the CI system
directly.

This also passes TEAMCITY_VERSION through the tox.ini environment variables. The
teamcity-messages automatically enables / starts itself when TEAMCITY_VERSION is
in the environment, which is only set on TeamCity build agents, making for
minimal CI changes being necessary, while leaving it compatible with other CI
system message formatting plugins.

This moves environment variables to one per line which makes them easier to work with.

Switch to using flake8-import-sort from isort
     - Does a better job sorting
     - Requires less configuration
     - Makes structured error messages CI can see / pop up rather than digging into logs

All the existing files are updated / adjusted to fit the new style rules.

Iterating on re-building this PR until the messages are all nicely structured.

### DIFF
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -2,4 +2,5 @@ sphinx>=1.3.1, <2.0
 tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
+teamcity-messages>=1.20
 -e .. # Install the DC/OS package
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -5,10 +5,10 @@ import re
 import sys
 import threading
 
-from dcos import constants
-
 import pytest
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+
+from dcos import constants
 
 from .common import (app, assert_command, assert_lines,
                      exec_command, list_deployments, popen_tty,
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -829,7 +829,7 @@ def _zero_instance_app():
 def _zero_instance_app_through_http():
     class JSONRequestHandler (BaseHTTPRequestHandler):
 
-        def do_GET(self):
+        def do_GET(self):  # noqa: N802
             self.send_response(200)
             self.send_header("Content-type", "application/json")
             self.end_headers()
--- a/tox.ini
+++ b/tox.ini
@@ -1,21 +1,30 @@
 [tox]
 envlist = py34-syntax, py34-unit
 
+[flake8]
+application-import-names=dcos
+import-order-style=smarkets
+
 [testenv]
 deps =
   mock
   pytest
   pytest-cov
-passenv = DCOS_* CI_FLAGS CLI_TEST_SSH_KEY_PATH
+  teamcity-messages
+
+passenv =
+  DCOS_*
+  CI_FLAGS
+  CLI_TEST_SSH_KEY_PATH
+  TEAMCITY_VERSION
 
 [testenv:py34-syntax]
 deps =
   flake8
-  isort
+  flake8-import-order
 
 commands =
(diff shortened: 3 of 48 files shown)


tamarrow:
@cmaloney can you include a screenshot of how messages will be displayed now?

cmaloney:
This PR's output:
image

New "tests" tab from TeamCity" which includes duration over time
image

The code coverage summary seems to be off because there are multiple independent test runs and TeamCity doesn't seem to quite deal with that right.

From another project with this same setup, here are what failed tests look like:
image

cmaloney:
Note the OS X binary, Linux Binary jobs would need more changes to do the same thing. I dug a little bit but it's quite a bit of wrappers to get to it / out of scope of the simple improvement aimed at.

tox.ini
```diff
@@ -5,13 +5,20 @@ envlist = py34-syntax, py34-unit
 deps =
   pytest
   pytest-cov
-passenv = DCOS_* CI_FLAGS CLI_TEST_SSH_KEY_PATH
+  teamcity-messages
```
tamarrow: newline

tox.ini
```diff
@@ -5,13 +5,20 @@ envlist = py34-syntax, py34-unit
...
+  CI_FLAGS
+  CLI_TEST_SSH_KEY_PATH
+  TEAMCITY_VERSION
 
 [testenv:py34-syntax]
 deps =
   flake8
   isort
-
+  teamcity-messages
```
tamarrow: newline

tox.ini
```diff
@@ -5,13 +5,20 @@ envlist = py34-syntax, py34-unit
...
+passenv =
+  TEAMCITY_VERSION
```
tamarrow: newline

tamarrow:
please also make this change to `cli/tox.ini`

cmaloney:
Verified the `Subproc().Popen()` -> `Subproc().popen()` by hand-auditing the results of `git grep 'Subproc'`

cmaloney:
@tamarrow should be ready for another review pass

cli/tests/integrations/test_marathon.py
```diff
@@ -829,7 +829,7 @@ def _zero_instance_app():
 def _zero_instance_app_through_http():
     class JSONRequestHandler (BaseHTTPRequestHandler):
 
-        def do_GET(self):
+        def do_GET(self):  # noqa: N802
```
tamarrow: is this to ignore the pep8 check?

cmaloney: Yes, and it ignores only the pep8 naming check. The `_GET` and friends of the BaseHTTPRequestHandler sadly do not follow pep8, so have to silence it.

cli/requirements.txt
```diff
@@ -2,4 +2,5 @@ sphinx>=1.3.1, <2.0
 tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
+teamcity-messages>=1.20
```
tamarrow: do we also need this now in `requirements.txt`?

cmaloney: I don't _think_ so. 

This causes the OS X CLI build which does a `make test-binary` which means it only gets what is in cli/requirements.txt installed to it, which includes pytest but not teamcity-messages which resulted in it not showing nice structured test results.

The top level requirements.txt doesn't have pytest, just tox, and so I think not having it there is good / right (there is nothing that would use teamcity-messages, just tox, and if someone runs tox, that adds the dependency).

tamarrow: Yes, we only run integration tests for the binary CIs which are in the 'cli' directory. Okay, makes sense to me.

tamarrow:
@cmaloney Waiting for the windows build to pass and then i'll merge. Thanks for contributing!

PR #781, opened 2016-09-28
Switch to using flake8-import-sort from isort

cmaloney:
First commit does the swap, second commit fixes all the newly found problems so it's easy to see what the tooling change is vs. the import reordering.

 - Does a better job sorting
 - Requires less configuration
 - Makes structured error messages CI can see / pop up rather than digging into logs

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-not_skip = __init__.py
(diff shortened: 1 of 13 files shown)


tamarrow:
@cmaloney please also update `cli/tox.ini`

cmaloney:
Now part of #780

PR #782, opened 2016-09-28
Add pep8 naming

cmaloney:
- Subprocess().Popen() -> Subprocess().popen() per PEP-8 conventions (member function names should be all lower case)
 - Move cosmos_error out of class because it isn't a member function per PEP-8
 - Add pep8-naming plugin to flake8 so PEP-8 errors are caught during syntax checks


cmaloney:
Now part of #780. Verified the `Subproc().Popen()` -> `Subproc().popen()` by hand-auditing the results of `git grep 'Subproc'`

PR #784, opened 2016-09-29
Support pods in the `marathon deployment list` table view

cruhland:
- Add a new POD column to the table from "affectedPods" field
- Correctly handle the occurrence of "app" xor "pod" in actions

### DIFF
--- a/cli/dcoscli/tables.py
+++ b/cli/dcoscli/tables.py
@@ -156,9 +156,8 @@ def deployment_table(deployments):
         path_id = action.get('app') or action.get('pod')
 
         if path_id is None:
-            msg = 'Expected "app" or "pod" field in action: {}'.format(action)
-            logger.exception(msg)
-            raise ValueError(msg)
+            template = 'Expected "app" or "pod" field in action: %s'
+            logger.exception(template, action)
 
         return path_id
 
@@ -177,7 +176,9 @@ def deployment_table(deployments):
                 raise ValueError(
                     'Unknown Marathon action: {}'.format(action['action']))
 
-            if multiple_apps:
+            if None in multiple_apps:
+                ret.append('N/A')
+            elif multiple_apps:
                 path_id = resource_path_id(action)
                 ret.append('{0} {1}'.format(action_display, path_id))
             else:
--- a/cli/dcoscli/tables.py
+++ b/cli/dcoscli/tables.py
@@ -156,10 +156,9 @@ def deployment_table(deployments):
         path_id = action.get('app') or action.get('pod')
 
         if path_id is None:
-            logger.exception('Missing app or pod entry')
-
-            template = 'Expected "app" or "pod" field in action: {}'
-            raise ValueError(template.format(action))
+            msg = 'Expected "app" or "pod" field in action: {}'.format(action)
+            logger.exception(msg)
+            raise ValueError(msg)
 
         return path_id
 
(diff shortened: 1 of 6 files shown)


cli/tests/unit/data/deployment/app.txt
```diff
@@ -0,0 +1,2 @@
+APP              POD  ACTION  PROGRESS  ID                                    
+/cassandra/dcos       scale     1/2     bebb8ffd-118e-4067-8fcb-d19e44126911
```
tamarrow: we should put a `-` under POD if it doesnt exist. same for app.

cli/dcoscli/tables.py
```diff
@@ -145,9 +149,20 @@ def deployment_table(deployments):
 
     """
 
+    def resource_path_id(action):
+        path_id = action.get('app') or action.get('pod')
+
+        if path_id is None:
+            logger.exception('Missing app or pod entry')
```
tamarrow: lets log `action` too

cli/dcoscli/tables.py
```diff
@@ -145,9 +149,20 @@ def deployment_table(deployments):
...
+
+            template = 'Expected "app" or "pod" field in action: {}'
+            raise ValueError(template.format(action))
```
tamarrow: instead of raising an exception, we should put 'N/A' under `ACTION`

cli/dcoscli/tables.py
cruhland: You mean log if the `action` field does not exist?

cli/dcoscli/tables.py
cruhland: So if any of the actions in a particular deployment have neither `"app"` nor `"pod"`, set the `ACTION` column for that row to `N/A`?

cli/tests/unit/data/deployment/app.txt
cruhland: Ok, will do.

cli/dcoscli/tables.py
tamarrow: log the json as part of the exception

cli/dcoscli/tables.py
tamarrow: yep.

cli/dcoscli/tables.py
```diff
@@ -149,21 +149,23 @@ def deployment_table(deployments):
 
     """
 
+    def join_path_ids(deployment, affected_resources_key):
```
tamarrow: pydoc

cli/dcoscli/tables.py
```diff
@@ -149,21 +149,23 @@ def deployment_table(deployments):
...
+        path_ids = deployment.get(affected_resources_key)
+        return '\n'.join(path_ids) if path_ids else '-'
+
     def resource_path_id(action):
```
tamarrow: pydoc

cli/dcoscli/tables.py
```diff
@@ -175,7 +177,9 @@ def get_action(deployment):
                 raise ValueError(
                     'Unknown Marathon action: {}'.format(action['action']))
 
-            if multiple_apps:
+            if None in action_path_ids:
```
tamarrow: should this be `if resource_path_id(action) is None`

cruhland:
OS X tests were being flaky again.

PR #785, opened 2016-09-29
Make Marathon 409 response status error message less confusing

cruhland:
- State that changes cannot be made
- Refer to an existing deployment as the reason


tamarrow:
merge when (pod) tests pass

PR #787, opened 2016-09-30
Remove `<properties>` arguments from `marathon pod update`

### LINKED ISSUE #779: Remove support for command-line JSON properties in `pod update`
Why? The UX is really bad. Here's what the command line looks like for a simple change:

```
dcos marathon pod update good-pod 'scaling={"instances":3,"kind":"fixed"}' \
    'id=/good-pod' \
    'containers=[{"name":"good-container","resources":{"cpus":0.1,"mem":16.0}}]' \
    'networks=[{"mode":"host"}]'
```

This is less of a problem with `app update` and `group update` because most of their useful properties are at the top levels of their JSON documents.

The solution is to remove this feature from `pod update` and display an error message if the user attempts to use it. The JSON can still be provided via stdin. cc @tamarrow @pyronicide @sascala

cruhland:
Pod descriptor JSONs have too many important nested fields, making
command-line properties awkward to specify. Removal improves the
user experience of the command.

Fixes #779.

### DIFF
--- a/cli/tests/integrations/test_marathon_pod.py
+++ b/cli/tests/integrations/test_marathon_pod.py
@@ -64,31 +64,23 @@ def test_pod_show():
         _assert_pod_show(GOOD_POD_ID, expected_json)
 
 
-@pytest.mark.skipif(not _PODS_ENABLED, reason="Requires pods")
-def test_pod_update_from_properties():
-    expected_json = file_json_ast(UPDATED_GOOD_POD_FILE_PATH)
-    containers_json_str = json.dumps(expected_json['containers'])
-    networks_json_str = json.dumps(expected_json['networks'])
-    properties = ['id=/{}'.format(GOOD_POD_ID),
-                  'containers={}'.format(containers_json_str),
-                  'networks={}'.format(networks_json_str)]
+def test_pod_update_does_not_support_properties():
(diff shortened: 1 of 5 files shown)


cli/tests/integrations/test_marathon_pod.py
```diff
@@ -64,31 +64,23 @@ def test_pod_show():
...
 
-@pytest.mark.skipif(not _PODS_ENABLED, reason="Requires pods")
-def test_pod_update_from_properties():
-    expected_json = file_json_ast(UPDATED_GOOD_POD_FILE_PATH)
-    containers_json_str = json.dumps(expected_json['containers'])
-    networks_json_str = json.dumps(expected_json['networks'])
-    properties = ['id=/{}'.format(GOOD_POD_ID),
-                  'containers={}'.format(containers_json_str),
-                  'networks={}'.format(networks_json_str)]
+def test_pod_update_does_not_support_properties():
```
tamarrow: we can remove this test

cruhland: I was thinking it would be good to keep around to prevent people from adding properties back to `pod update` in the future. I could see someone doing that in an effort to keep all the `update` commands consistent.

Issue #783, opened 2016-09-29
Remove pods gating once pods have landed in testing/master


cruhland:
- Remove `@pytest.mark.skipif` decorators in `cli/tests/integrations/test_marathon_pod.py`
- Delete this TeamCity build: https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcosIo_DcosCli_CliAgainstLatestDcOsWithPods

PR #792, opened 2016-10-04
Fix accidental skipping of pods integration tests in CI builds

cruhland:
Add PODS_ENABLED to Tox env var whitelist

You'll be happy/relieved to know that all tests still pass! I also updated #783 so that we remember to undo this change as well.


tamarrow:
@cruhland can we change the env var to DCOS_PODS_ENABLED? than it will be automagically picked up by tox and we'll have less things to do later?

cruhland:
@tamarrow updated! I also updated the pods CI build to use the new env var.

PR #797, opened 2016-10-06
Update `marathon pod show` to use `GET /v2/pods/{pod-id}::status`

cruhland:
Example output can be found in `cli/tests/data/marathon/pods/good_status.json` and `cli/tests/data/marathon/pods/updated_good_status.json`.


dcos/marathon.py
```diff
@@ -758,7 +758,7 @@ def show_pod(self, pod_id):
         :rtype: dict
         """
 
-        path = self._marathon_id_path_join('v2/pods', pod_id)
+        path = self._marathon_id_path_join('v2/pods', pod_id) + '::status'
```
tamarrow: lets add `::status` in the path join

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -220,7 +259,7 @@ def _pod(pod_id, file_path):
             _assert_pod_remove(pod_id, extra_args=['--force'])
 
 
-def _wait_for_instances(expected_instances, max_attempts=10):
+def _wait_for_instances(expected_instances, max_attempts=20):
```
tamarrow: :/ how this been flaky?

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -111,26 +113,63 @@ def _assert_pod_list_json_subset(expected_json, actual_json):
...
-    actual_containers = actual_pod['containers']
+    assert actual_pod_status['id'] == expected_pod_status['id']
+    assert actual_pod_status['status'] == expected_pod_status['status']
+    assert len(actual_pod_status['instances']) == \
+        len(expected_pod_status['instances'])
+
+    _assert_pod_spec_json(expected_pod_status['spec'],
+                          actual_pod_status['spec'])
+
+    for i in range(len(expected_pod_status['instances'])):
```
tamarrow: shouldn't the output of each of the instances be identical here?

dcos/marathon.py
cruhland: I tried, but that method escapes colons, so it ends up as e.g. `/v2/pods/foo%3A%3Astatus`.

cli/tests/integrations/test_marathon_pod.py
cruhland: Yes, your point being?

cli/tests/integrations/test_marathon_pod.py
cruhland: I had to increase the time because the `pod update` test now spins up six instances.

cli/tests/integrations/test_marathon_pod.py
tamarrow: that we can probably make this simpler

cruhland: I could change the "expected" file to have only one instance, and update the code to compare that instance to every instance in the "actual" JSON. But then the "expected" file won't mirror the structure of the "actual" JSON anymore. Is that okay?

cli/tests/integrations/test_marathon_pod.py
cruhland: With that change, the total time of the pods integration tests is about one minute.

tamarrow: ... should we just create fewer instances?

dcos/marathon.py
tamarrow: 👍

cli/tests/integrations/test_marathon_pod.py
cruhland: Sure, I don't even really need to test that scaling works at all. I'll just remove that from the update JSON.

cli/tests/integrations/test_marathon_pod.py
tamarrow: The expected file is already a subset, so it's fine. We should make it clear that these should be identical. We could even

dcos/marathon.py
```diff
@@ -758,7 +758,8 @@ def show_pod(self, pod_id):
         :rtype: dict
         """
 
-        path = self._marathon_id_path_join('v2/pods', pod_id) + '::status'
+        normalized_pod_id = urllib.parse.quote(pod_id.strip('/'))
+        path = 'v2/pods/{}::status'.format(normalized_pod_id)
```
tamarrow: why cant we use `_marathon_id_path_join` here?

cruhland:
@tamarrow Updateed, got the pods integration tests down to 30 seconds! And the flakiness is hopefully gone.

tamarrow:
@cruhland looks like the pod_list test just failed :/

cruhland:
@tamarrow actually it's a marathon groups test -- failed with connection timeout. I'll re-run.

cruhland:
Cluster creation failed...

PR #803, opened 2016-10-11
Implement `pod instance kill`

cruhland:


tests/test_marathon.py
```diff
@@ -152,6 +152,44 @@ def test_update_pod_raises_dcos_exception_if_deployment_id_missing():
...
+
+    _assert_kill_pod_instances_executes_successfully(
+        pod_id='foo',
+        instance_ids=['i1', 'i2', 'i3'],
+        path='v2/pods/foo::instance',
+        response_json={'some': ['instance', 'status']})
+
+    _assert_kill_pod_instances_executes_successfully(
+        pod_id='foo',
+        instance_ids=['instance1', 'instance2'],
```
tamarrow: the differences here seem to be validating the we can handle variations in the url rather than than the kill command is doing what we expect. Lets have tests that cover url variations (if we don't already) and ones that just test the `pod kill command`

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -92,6 +94,34 @@ def test_pod_update_from_stdin():
...
+        kill_1, keep, kill_2 = assert_show_and_get_instance_ids()
+
+        remove_args = [INSTANCE_KILL_ID, kill_1, kill_2]
+        assert_command(_POD_INSTANCE_KILL_CMD + remove_args)
+
+        new_instance_ids = assert_show_and_get_instance_ids()
+        assert keep in new_instance_ids
+        assert kill_1 not in new_instance_ids
+        assert kill_2 not in new_instance_ids
+        assert len(new_instance_ids) == 3
```
tamarrow: we should check the length before the kill as well.

tamarrow: if there are 3 instances to start with, and we kill 2, why do we end up with 3?

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -92,6 +94,34 @@ def test_pod_update_from_stdin():
         watch_all_deployments()
 
 
+@pytest.mark.skipif(not _PODS_ENABLED, reason="Requires pods")
+def test_pod_instance_kill():
```
tamarrow: we should also test the case where the user doesn't input any `instance-ids`

cli/tests/integrations/test_marathon_pod.py
cruhland: I'll look into if we need a `--scale` parameter. I'll also add a comment here.

tests/test_marathon.py
cruhland: I'll remove the redundant cases.

cruhland:
@tamarrow pods CCM clusters are broken right now, I'll find out when I can get a clean build.

cruhland:
@tamarrow it might be a while, there's a Marathon issue being debugged right now. Feel free to review my changes though!

cli/tests/integrations/test_marathon_pod.py
```diff
@@ -92,6 +94,35 @@ def test_pod_update_from_stdin():
         watch_all_deployments()
 
 
+@pytest.mark.skipif(not _PODS_ENABLED, reason="Requires pods")
+def test_pod_kill():
+    def get_instance_ids():
```
tamarrow: lets make this a first class function

tamarrow:
@cruhland one small change and then LGTM once tests pass

PR #804, opened 2016-10-12
Add yaml format for resource file

duythinht:
I just think the yaml file is very clean, simple and effective than json file. Just add some code to load resource from .yaml or .yml file.
Please consider about it.

### DIFF
--- a/cli/dcoscli/job/main.py
+++ b/cli/dcoscli/job/main.py
@@ -806,6 +806,8 @@ def _get_resource(resource):
     if resource is not None:
         if os.path.isfile(resource):
             with util.open_file(resource) as resource_file:
+                if resource.endswith('.yaml') or resource.endswith('.yml'):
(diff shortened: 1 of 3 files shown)


tamarrow:
@duythinht thanks for the pr! I agree that there are benefits to YAML, but unfortunately we have many commands that load resources, and we want to be consistent with the resource format- which is `JSON`.

Feel free to add a feature request to support YAML [here](https://dcosjira.atlassian.net/projects/DCOS/).

tamarrow:
@duythinht closing this pr. Please re-open if you don't agree.

krzkaczor:
I would love to see support for yaml. Especially things like comments and increased readability are nice. We already use `yaml` and do "ad-hoc" translation. Something like `yaml2json pod-definition.yml |  dcos marathon pod add` works just fine ;)

PR #806, opened 2016-10-12
pod add should return the deployment id

kensipe:


cli/dcoscli/marathon/main.py
```diff
@@ -865,7 +865,8 @@ def pod_add(self, pod_resource_path):
         self._ensure_pods_support(marathon_client)
 
         pod_json = self._resource_reader.get_resource(pod_resource_path)
-        marathon_client.add_pod(pod_json)
+        deployment_id = marathon_client.add_pod(pod_json)
+        emitter.publish(deployment_id)
```
tamarrow: emitter.publish('Deployment-id: {}'.format(deployment_id)

cruhland:
@kensipe there are some unit tests that will fail due to your changes. Please make sure to update them to test the new behavior.

Also, the CI builds don't exercise the pod integration tests by default, but we have a non-triggered build specifically for pods, which I just started manually: https://teamcity.mesosphere.io/viewLog.html?buildId=443018&buildTypeId=DcosIo_DcosCli_CliAgainstLatestDcOsWithPods&guest=1

kensipe:
@cruhland thanks!  looking at it now.

tests/test_marathon.py
```diff
@@ -541,7 +536,7 @@ def _assert_add_pod_returns_parsed_response_body(response_json):
     marathon_client, rpc_client = _create_fixtures()
     rpc_client.http_req.return_value = mock_response
 
-    assert marathon_client.add_pod({'some': 'json'}) == response_json
+    assert marathon_client.add_pod({'some': 'json'}) startswith "Marathon-Deployment-ID"
```
cruhland: Did you try running this locally? I don't see how this will work, since you haven't set the header value for `Marathon-Deployment-Id` on the mock response object.

Since `response_json` isn't used anymore, could you rename it to `deployment_id` and use it to pass in the expected header value?

tests/test_marathon.py
```diff
@@ -21,11 +21,6 @@ def test_add_pod_returns_parsed_response_body():
     _assert_add_pod_returns_parsed_response_body(["another", "pod", "json"])
 
 
-def test_add_pod_raises_dcos_exception_for_json_parse_errors():
-    _assert_method_raises_dcos_exception_for_json_parse_errors(
-        lambda marathon_client: marathon_client.add_pod({'some': 'json'}))
```
cruhland: Since you're now returning the header, let's add a test to cover the case when the header value is missing (we want to make sure the user doesn't see a Python stacktrace when things fail). See the `update_pod()` method in `dcos/marathon.py` for code that handles it already (you might want to factor that out into its own method), and the corresponding unit test in this file: `test_update_pod_raises_dcos_exception_if_deployment_id_missing()`.

kensipe: makes sense

tests/test_marathon.py
kensipe: yeah I should have waited to push...

cruhland:
@tamarrow @kensipe for consistency, should we output `Created deployment ...` instead of `Deployment-id: ...`? A number of marathon commands already use that wording, e.g. https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/marathon/main.py#L664

kensipe:
good question regarding deployment id...

kensipe:
I think I like the `Deployment-id:` over `Created deployment`.     doesn't the `:` make the output more parsable?    I strongly agree with consistency.    I can change to Created deployment if that is the consensus.

kensipe:
the fail test isn't related... :(

cruhland:
@kensipe those tests are flaky, so I've restarted the builds. We've been working on improving them but haven't had time to get to all of them yet.

tamarrow:
@kensipe lets stick with `Created deployment`. Consistency is always the least confusing.

kensipe:
2 is quorum :)   SGTM   updated and ready to go assuming tests pass.

tamarrow:
thanks @kensipe. These integration tests needs to be updated with the new stdout: https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_marathon_pod.py#L34
and
https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_marathon_pod.py#L47

cruhland:
@kensipe Marathon updated their endpoint name, so the pods tests will fail until this gets merged: https://github.com/dcos/dcos-cli/pull/809

kensipe:
@cruhland ah... that explains it.  thanks!!

PR #809, opened 2016-10-13
Marathon endpoint for `pod kill` now ends with plural `::instances`

cruhland:
Pod tests for PR #806 won't pass until this is merged.

### DIFF
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -701,3 +701,41 @@ def base64_to_dict(byte_string):
     :rtype dict
     """
     return json.loads(base64.b64decode(byte_string).decode('utf-8'))
+
+
+UNIVERSE_REPO = "https://universe.mesosphere.com/repo"
+UNIVERSE_TEST_REPO = "http://universe.marathon.mesos:8085/repo"
+
+
+def setup_universe_server():
+    # add universe-server with static packages
+    assert_command(
+        ['dcos', 'marathon', 'app', 'add', 'tests/data/universe-v3-stub.json'])
+    watch_all_deployments()
+
+    assert_command(
+        ['dcos', 'package', 'repo', 'remove', 'Universe'])
+
+    assert_command(
+        ['dcos', 'package', 'repo', 'add', 'test-universe', UNIVERSE_TEST_REPO]
+    )
+
+    # Give the test universe some time to become available
+    describe_command = ['dcos', 'package', 'describe', 'helloworld']
+    for _ in range(10):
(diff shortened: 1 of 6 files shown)


cruhland:
@tamarrow I made some significant test changes, please take a look!

cli/tests/integrations/test_task.py
```diff
@@ -328,3 +328,15 @@ def _get_completed_task_id(app_id='test-app-completed'):
     assert len(tasks) == 1
     task_id = tasks[0]['id']
     return task_id
+
+
+def _read_lines_once_available(file_obj):
```
tamarrow: rename to `_read_lines(file_obj)`, we don't need to expose implementation. Please add a pydoc :)

cli/tests/integrations/common.py
```diff
@@ -701,3 +701,41 @@ def base64_to_dict(byte_string):
...
+    assert_command(
+        ['dcos', 'package', 'repo', 'remove', 'Universe'])
+
+    assert_command(
+        ['dcos', 'package', 'repo', 'add', 'test-universe', UNIVERSE_TEST_REPO]
+    )
+
+    # Give the test universe some time to become available
+    describe_command = ['dcos', 'package', 'describe', 'helloworld']
+    for _ in range(10):
```
tamarrow: the problem is sometimes this takes > 10 seconds.

cli/tests/integrations/common.py
```diff
@@ -701,3 +701,41 @@ def base64_to_dict(byte_string):
...
+        returncode, _, _ = exec_command(describe_command)
+        if returncode == 0:
+            break
+        time.sleep(1)
+    else:
+        teardown_universe_server()
```
tamarrow: this call is unnecessary. should be done by the teardown module

cruhland: Nope! When the assertion is raised on the following line, PyTest does not call `teardown_module()` afterwards. This happened to me with local testing and I had to manually run the teardown commands. The library must assume that the setup code will not raise an exception.

cli/tests/integrations/common.py
cruhland: I'll bump it up to 30 seconds. If it's slower than that, something must be really wrong. And if it's faster than that, this code won't wait for the full time anyway.

cli/tests/integrations/common.py
cruhland: I'll add a comment to clarify.

tamarrow: please link to documentation

PR #810, opened 2016-10-13
app and group add returns deployment id

kensipe:


dcos/marathon.py
```diff
@@ -328,7 +328,15 @@ def add_app(self, app_resource):
 
         response = self._rpc.http_req(http.post, 'v2/apps', json=app_json)
 
-        return response.json()
+        try:
+            body_json = response.json()
+            return body_json['deployments'][0]['id']
```
tamarrow: lets use `.get` instead of catching an exception.

dcos/marathon.py
```diff
@@ -711,7 +719,15 @@ def create_group(self, group_resource):
             group_json = group_resource
 
         response = self._rpc.http_req(http.post, 'v2/groups', json=group_json)
-        return response.json()
+        try:
+            body_json = response.json()
+            return body_json["deploymentId"]
```
tamarrow: lets use `.get` instead of catching an exception.

tamarrow:
@kensipe looks like you need to rebase and then we can merge :)

PR #813, opened 2016-10-17
Encode to utf-8 if not a terminal

TerryHowe:
Printing anything redirected or piped that is unicode will cause an exception if it is not encoded to UTF-8.

### DIFF
--- a/dcos/emitting.py
+++ b/dcos/emitting.py
@@ -154,7 +154,7 @@ def _page(output, pager_command=None):
     output = six.text_type(output)
 
     if not sys.stdout.isatty() or util.is_windows_platform():
-        print(output)
+        print(output.encode('utf-8'))
         return
 
     num_lines = output.count('\n')


TerryHowe:
Ha! Well that broke about every test in the world.

tamarrow:
@TerryHowe can you give an example of what you are trying to do?

TerryHowe:
With a unicode character in a job

```
dcos marathon app list | grep foo
```

Piping to grep forces it down that not tty path and boom.  I need to have a look at the tests on this, I though it was going to be simple and then yesterday was a crazy day.  Maybe today.

tamarrow:
Are you using python2.7? In python3.4 it defaults to printing as utf-8: http://stackoverflow.com/questions/4545661/unicodedecodeerror-when-redirecting-to-file

If you specify `PYTHONIOENCODING` you can override it to `utf-8`

TerryHowe:
Yes, py27 and sure there are some work arounds.

TerryHowe:
too much work to fix those tests for this.

PR #815, opened 2016-10-21
pods: read from task log not executor

tamarrow:
https://mesosphere.atlassian.net/browse/DCOS-10828

### DIFF
--- a/cli/tests/integrations/common.py
+++ b/cli/tests/integrations/common.py
@@ -151,17 +151,17 @@ def wait_for_service(service_name, number_of_services=1, max_count=300):
         count += 1
 
 
-def add_job(app_path):
+def add_job(job_path):
     """ Add a job, and wait for it to deploy
 
-    :param app_path: path to job's json definition
-    :type app_path: str
+    :param job_path: path to job's json definition
+    :type job_path: str
     :param wait: whether to wait for the deploy
     :type wait: bool
     :rtype: None
     """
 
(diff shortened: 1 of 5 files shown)


kensipe:
@tamarrow a couple of doc issues but otherwise looks really good!

tamarrow:
thanks @kensipe! fixed the docs issues

cli/tests/integrations/common.py
```diff
@@ -522,15 +538,78 @@ def app(path, app_id, wait=True):
...
+    """Context manager that deploys pods on entrance, and removes
+    them on exit.
+
+    :param pods: dict of path/to/pod/json -> pod id
+    :type pods: {}
+    :rtype: None
+    """
+
+    for pod_path in pods:
+        add_pod(pod_path)
```
cruhland: Don't we want `wait=False` here?

cli/tests/integrations/common.py
```diff
@@ -151,11 +151,11 @@ def wait_for_service(service_name, number_of_services=1, max_count=300):
         count += 1
 
 
-def add_job(app_path):
+def add_job(job_path):
```
kensipe: the issue with the build is app_path was changed to "job_path" on line 154, however line 164 also needs to change to job_path.   passes syntax check after the change.

cli/tests/integrations/common.py
kensipe: @tamarrow I totally agree with charles here as well!

PR #820, opened 2016-10-26
python27: stop publishing wheel

tamarrow:
We want to be able to use python3 features.

For users on DC/OS 1.8+: They should be using the binary CLI- so this should not affect them
For legacy users (DC/OS < 1.8): With this change, they will download the latest supported 2.7 version when following the UI instructions from their cluster.  

For developers: python3 wheels for `dcos` and `dcoscli` packages will be published. If they want to use python2.7 they can still use the older 2.7 published wheels for this package.

### DIFF
--- a/bin/install/install-optout-dcos-cli.sh
+++ b/bin/install/install-optout-dcos-cli.sh
@@ -152,7 +152,7 @@ if [ -z "$DCOS_CLI_VERSION" ]; then
     if $(compare_version 1.6.1); then
         "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli<0.4.0"
     else
-        "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli"
+        "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli==0.4.12"
     fi
 else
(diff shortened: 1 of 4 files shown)


bin/install/install-optout-dcos-cli.sh
```diff
@@ -152,7 +152,7 @@ if [ -z "$DCOS_CLI_VERSION" ]; then
     if $(compare_version 1.6.1); then
         "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli<0.4.0"
     else
-        "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli"
+        "$VIRTUAL_ENV_PATH/bin/pip" install --quiet "dcoscli==0.4.12"
```
jsancio: Maybe we can make the `0.5.0` release the release that drops `p27` support. If we do this we can change this to `dcoscli<0.5`.

jsancio:
LGTM. One minor comment.

bin/install/install-optout-dcos-cli.sh
tamarrow: The reason for this version is 1. versions afterwards use features that aren't compatible with dc/os 1.7 (which should be the last version to use non-binary clis). 2. a 2.7 bug got introduced after 0.4.12 :)

PR #823, opened 2016-10-28
Reduce requests

### LINKED ISSUE #818: DCOS Sub Command CLI has problems with POSTing multipart
### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?

dcoscli.version=0.4.10
dcos.version=1.8.4
dcos.commit=e64024af95b62c632c90b9063ed06296fcf38ea5
dcos.bootstrap-id=5b4aa43610c57ee1d60b4aa0751a1fb75824c083

OS X 10.11.6
### What did you do?

I'm writing a custom sub command and I'm using the DC/OS Python http library (0.4.14). When performing a POST request that uploads files I receive a bad request status back. The reason for this is because the DC/OS implementation of the post request makes several attempts. The first attempt is without authorisation, and the subsequent ones are with authorisation. Because I'm uploading a multipart file, and the stream associated with the request isn't reset, no content will be sent on the subsequent requests. Here's how I prepare the request:

```
        files.append(('bundle', (bundle_name, open(bundle_file, 'rb'))))
```

That my load request succeed even on subsequent attempts.
### What did you see instead?

The first load correctly streams the parts, but the subsequent ones do not. The first one fails though because authorisation is required.
---

The solution is for your flavour of the requests post method to scan through the `files` data structure and reset any stream found by calling `seek(0)`.

Even better, why even retry? Why not just one request with authorisation always? I'd state that you've broken the contract of the request post method, and that perhaps this should be some higher level function.

huntc:
Reduce the requests made to HEAD requests in order to determine the type of authentication required, and actually authenticate. Once authenticated, then make the request. This strategy preserves the content of the request object, particularly when they contains stream objects as in the case of multipart/formdata requests.

Fixes #818

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -157,7 +157,6 @@ def _request_with_auth(response,
         else:
             auth = AUTH_CREDS[creds]
 
-    # try request again, with auth
     response = _request(method, url, is_success, timeout, auth,
                         verify, **kwargs)
 
@@ -174,9 +173,6 @@ def _request_with_auth(response,
                        "Please run: `dcos auth login`")
                 raise DCOSException(msg)
 
-    if response.status_code == 401:
-        raise DCOSAuthenticationException(response)
-
     return response
 
 
@@ -186,8 +182,14 @@ def request(method,
             timeout=None,
             verify=None,
             **kwargs):
-    """Sends an HTTP request. If the server responds with a 401, ask the
-    user for their credentials, and try request again (up to 3 times).
+    """Sends an HTTP request. We first send a HEAD request for the
+    supplied URL so that we can determine the type of authentication
+    required (if any). If authentication is required then we again
+    use a HEAD request asking the user for their credentials, and
+    try request again (up to 3 times). Once authenticated, we issue
+    the request passed in. We are careful to execute the request
+    passed in just once given that it may be stateful e.g. any files
+    object containing a stream may only be evaluated once.
 
     :param method: method for the new Request object
     :type method: str
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -216,20 +216,26 @@ def request(method,
     if verify is not None:
         silence_requests_warnings()
 
-    response = _request('head', url, is_success, timeout,
-                        verify=verify)
-    i = 0
-    while i < 3 and response.status_code == 401:
-        auth_response = _request_with_auth(response, 'head', url, is_success,
-                                           timeout, verify=verify)
-        response.status_code = auth_response.status_code
-        i += 1
+    dcos_url = config.get_config_val("core.dcos_url")
+    auth_url = urllib.parse.urljoin(dcos_url, 'exhibitor/')
+    response = _request(
+        'GET', auth_url, is_success, timeout, verify=verify, **kwargs)
 
     if response.status_code == 401:
-        raise DCOSAuthenticationException(response)
+        i = 0
+        while i < 3 and response.status_code == 401:
+            auth_response = _request_with_auth(
+                response, 'GET', auth_url, is_success, timeout,
+                verify=verify, **kwargs)
+            response.status_code = auth_response.status_code
+            i += 1
 
-    response = _request_with_auth(response, method, url, is_success,
-                                  timeout, verify, **kwargs)
+        response = _request_with_auth(response, method, url, is_success,
+                                      timeout, verify=verify, **kwargs)
+        if response.status_code == 401:
+            raise DCOSAuthenticationException(response)
+    else:
+        response = _request(method, url, is_success, timeout, verify, **kwargs)
 
     if is_success(response.status_code):
         return response
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -126,7 +126,7 @@ def _request_with_auth(response,
                        timeout=None,
                        verify=None,
                        **kwargs):
-    """Try request (3 times) with credentials if 401 returned from server
+    """Request with credentials
 
     :param response: requests.response
     :type response: requests.Response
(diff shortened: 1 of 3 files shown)


huntc:
@tamarrow Any chance you could look at this and advise on if/when a release/merge would be possible? Thanks.

dcos/http.py
```diff
@@ -146,37 +146,33 @@ def _request_with_auth(response,
...
+    creds = (hostname, auth_scheme, realm)
 
-        i += 1
+    with lock:
+        if creds not in AUTH_CREDS:
+            auth = _get_http_auth(response, parsed_url, auth_scheme)
+        else:
+            auth = AUTH_CREDS[creds]
+
+    # try request again, with auth
```
tamarrow: we should remove this comment

dcos/http.py
```diff
@@ -146,37 +146,33 @@ def _request_with_auth(response,
...
+        elif response.status_code == 401 and \
+                auth_scheme in ["acsjwt", "oauthjwt"]:
+
+            if config.get_config_val("core.dcos_acs_token") is not None:
+                msg = ("Your core.dcos_acs_token is invalid. "
+                       "Please run: `dcos auth login`")
+                raise DCOSException(msg)
 
     if response.status_code == 401:
         raise DCOSAuthenticationException(response)
```
tamarrow: we don't want to raise here anymore since we are looping outside this function.

dcos/http.py
```diff
@@ -218,12 +214,20 @@ def request(method,
     if verify is not None:
         silence_requests_warnings()
 
-    response = _request(method, url, is_success, timeout,
-                        verify=verify, **kwargs)
+    response = _request('head', url, is_success, timeout,
```
tamarrow: lets add a comment to explain why we are requesting with `head.

tamarrow:
ok to test

dcos/http.py
huntc: I left it there because the prev implementation was responsible for throwing it ie I've preserved its contract. I don't think it is good to rely on the outer behaviour unless otherwise specified in its contract. 

That said, I'll change it if that's what you want. Please confirm.

tamarrow: We can update the pydoc with this information

dcos/http.py
```diff
@@ -218,12 +214,20 @@ def request(method,
...
+                        verify=verify)
+    i = 0
+    while i < 3 and response.status_code == 401:
+        auth_response = _request_with_auth(response, 'head', url, is_success,
```
huntc: You don't want to overwrite the orig response given that the www-authenticate header will be lost.

dcos/http.py
huntc: ok

dcos/http.py
huntc: Looking further, I guess it is an internal function so perhaps no need to doc further. I'll remove the raising of the exception.

huntc:
PR feedback incorporated. I previously ran `tox` in order to test things so it was unexpected to see that the integration tests also required running. Also, I'm unable to get the integration tests running locally so hopefully the change that I made there is sufficient.

huntc:
@tamarrow I may need your help in getting the integration tests done. Thanks.

BTW things are now working well with my DC/OS CLI sub command integration given these changes. I've been using the changes here against our production environment.

tamarrow:
Yay, glad to hear it! 

Re: integration tests: `test_http_auth.test_request_with_bad_auth_basic` needs to be updated because we changed `http._request_with_auth`

huntc:
I'm really sorry to ask - would you mind updating the CLI tests... I'm having great difficulty running them and I feel that we're going to play ping-pong here otherwise. Sorry again to ask. I've granted you access to my repo:

tamarrow:
@huntc happy to help, I'll get to it as soon as I can. If you want to talk about difficulties with running the tests locally, free free to join our open slack account: http://chat.mesosphere.com/ and we can chat there.

huntc:
Thanks! I'll post my troubles to your gitter channel today.

dcos/http.py
```diff
@@ -216,20 +216,26 @@ def request(method,
...
-    i = 0
-    while i < 3 and response.status_code == 401:
-        auth_response = _request_with_auth(response, 'head', url, is_success,
-                                           timeout, verify=verify)
-        response.status_code = auth_response.status_code
-        i += 1
+    dcos_url = config.get_config_val("core.dcos_url")
+    auth_url = urllib.parse.urljoin(dcos_url, 'exhibitor/')
+    response = _request(
+        'GET', auth_url, is_success, timeout, verify=verify, **kwargs)
```
huntc: I'd still use a `HEAD` here - it is cheaper given the lack of content in reply. You'll still get the headers you need, and your service can potentially optimise in terms of handling `HEAD`.

dcos/http.py
```diff
@@ -216,20 +216,26 @@ def request(method,
...
 
     if response.status_code == 401:
-        raise DCOSAuthenticationException(response)
-
-    response = _request_with_auth(response, method, url, is_success,
-                                  timeout, verify, **kwargs)
+        i = 0
+        while i < 3 and response.status_code == 401:
+            auth_response = _request_with_auth(
+                response, 'GET', auth_url, is_success, timeout,
```
huntc: Why's `response` being passed as the request?

Also, `HEAD` should be cheaper as mentioned. `HEAD` should always be interchangeable with `GET`.

tamarrow:
@huntc see #824

tamarrow:
Closing since this bug is fixed with #824

PR #824, opened 2016-11-02
http: don't retry requests if response is 401

tamarrow:
We were retrying requests with 401 so that we can use the response
headers to determine type of authentication. Instead, always send
request with auth if we already have credentials (from
core.dcos_acs_token).

This also fixes multi-part requests that currently break if we have to
retry the request (since the stream gets opened by the initial request).

### DIFF
--- a/cli/tests/unit/test_http_auth.py
+++ b/cli/tests/unit/test_http_auth.py
@@ -103,8 +103,9 @@ def test_request_with_bad_auth_basic(mock, req_mock, auth_mock):
     req_mock.return_value = mock
 
     with pytest.raises(DCOSException) as e:
-        http._request_with_auth(mock, "method", mock.url)
-    assert e.exconly().split(':')[1].strip() == "Authentication failed"
+        http.request_with_auth("method", mock.url)
+    msg = "Authentication failed. Please run `dcos auth login`"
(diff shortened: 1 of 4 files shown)


cli/tests/unit/test_http_auth.py
```diff
@@ -103,8 +103,9 @@ def test_request_with_bad_auth_basic(mock, req_mock, auth_mock):
     req_mock.return_value = mock
 
     with pytest.raises(DCOSException) as e:
-        http._request_with_auth(mock, "method", mock.url)
-    assert e.exconly().split(':')[1].strip() == "Authentication failed"
+        http.request_with_auth("method", mock.url)
+    msg = "Authentication failed. Please run `dcos auth login`"
```
vespian: Some people will disagree with me, some not - please treat it only as recommendation/advice.

Generally I am against using exception text parsing in tests. The reason is that this is meant for humans and we can express the same information with i.e. "You must run `dcos auth login because authentication has failed.` This will fail the test but will trigger the same action in human. IMO the fact that we need to resort to parsing messages means that our exception hierarchy/types is not granular enough and we need to add more.

2cents apply.

huntc:
Thanks!

PR #834, opened 2016-11-17
Fix failing test caused by requests library update

cruhland:
Replace an actual Response object with a mock to
ensure we have the correct state for the test


jsancio:
LGTM! :shipit:

jsancio:
@tamarrow the Binary tests are failing. Have you seen this before?

tamarrow:
Windows wasn't run since it was triggered during Jenkins upgrade. Since unit tests only run on "continuous-integration", it's okay.

PR #835, opened 2016-11-18
better reporting of json processing failures for metronome

kensipe:
add disk to json if not provided.  
provide back meaningful message for json processing failures.

### DIFF
--- a/cli/dcoscli/job/main.py
+++ b/cli/dcoscli/job/main.py
@@ -576,20 +576,9 @@ def _add_job(job_file):
         del full_json['schedules']
 
     # iterate and post each schedule
-    job_added = False
-    try:
-        _post_job(full_json)
-        job_added = True
-    except DCOSHTTPException as e:
-        if e.response.status_code == 422:
-            message = e
-            if e.response is not None and e.response.text is not None:
-                message = e.response.text
-            raise DCOSException(message)
-        else:
-            raise DCOSException(e)
+    _post_job(full_json)
 
-    if schedules is not None and job_added is True:
+    if schedules is not None:
         return _add_schedules(job_id, schedules)
 
     return 0
--- a/cli/dcoscli/job/main.py
+++ b/cli/dcoscli/job/main.py
@@ -484,17 +484,7 @@ def _add_schedules(job_id, schedules_json):
         return 1
 
     for schedule in schedules_json:
-        try:
-            _post_schedule(job_id, schedule)
-        except DCOSHTTPException as e:
-            if e.response.status_code == 404:
-                emitter.publish("Job ID: '{}' does NOT exist.".format(job_id))
-            elif e.response.status_code == 409:
-                emitter.publish("Schedule already exists.")
-            else:
-                raise DCOSException(e)
-        except DCOSException as e:
-            raise DCOSException(e)
+        _post_schedule(job_id, schedule)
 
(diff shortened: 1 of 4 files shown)


cli/dcoscli/job/main.py
```diff
@@ -575,6 +575,9 @@ def _add_job(job_file):
     if 'id' not in full_json:
         raise DCOSException("Jobs JSON requires an ID.")
 
+    if 'disk' not in full_json['run']:
+        full_json['run']['disk'] = 0
```
tamarrow: should we check if `run` exists?

cli/dcoscli/job/main.py
```diff
@@ -588,8 +591,11 @@ def _add_job(job_file):
         _post_job(full_json)
         job_added = True
     except DCOSHTTPException as e:
-        if e.response.status_code == 409:
-            emitter.publish("Job ID: '{}' already exists".format(job_id))
+        if e.response.status_code == 422:
```
tamarrow: Since we are no longer having a custom message here, we should remove the `try/except`. we catch `DCOSHTTPException` in `http.request`.

tamarrow:
thanks @kensipe, some small comments.

cli/dcoscli/job/main.py
kensipe: run is required as part of the schema and there is no such thing as a job with a run command.  If you want me to provide the check I will.  thoughts?

cli/dcoscli/job/main.py
kensipe: this is the main point of this code... the message for http 422 provides a generic processing error in the dcos exception message... however what we want is the information that is in the body of the http response which includes the reason why the json process failed.

kensipe:
@tamarrow 
Thanks for the feedback!!!!  this is a MUCH better way of handling this and now works for all situations of JSON entity processing.   The issue with JSON processing failure is it is too hard to know why.  now the information comes back from the service is provided to the user.  For example:

`dcos.errors.DCOSUnprocessableException: Error while fetching [http://ken-qyxk6-elasticl-ots5puw3smfz-1917677140.us-west-2.elb.amazonaws.com/service/metronome/v1/jobs/pikachu/schedules]: HTTP 422: {"message":"Object is not valid","details":[{"path":"","errors":["Wrong type. Expected object, was string."]}]}`

or 
`raise DCOSUnprocessableException(response)
dcos.errors.DCOSUnprocessableException: Error while fetching [http://ken-qyxk6-elasticl-ots5puw3smfz-1917677140.us-west-2.elb.amazonaws.com/service/metronome/v1/jobs]: HTTP 422: {"message":"Object is not valid","details":[{"path":"/run","errors":["Property disk missing"]}]}`

This needs to be combined with 836.

kensipe:
@tamarrow rebased!

cli/dcoscli/job/main.py
tamarrow: we can now just have `except DCOSUnprocessableException` here :)

kensipe: you are right!

kensipe:
@tamarrow you are correct on the try block with the new change.   rebased and ready.

tamarrow:
thanks @kensipe will merge when tests pass :)

PR #837, opened 2016-11-18
auth: new flow allowing auth from CLI for all configured providers

tamarrow:


cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+    :rtype: str | None
+    """
+
+    password = None
+    if password_str:
+        password = password_str
+    elif password_env:
+        password = os.environ.get(password_env)
+    elif password_file:
+        password = util.read_file(password_file)
```
jgehrcke: We should validate the permissions of that file and error out if they are too open. Do we already do that? See my corresponding comment further below, in the context of the service account login, where I propose something in the lines of `util.read_file_secure(path)`.

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    :rtype: str
+    """
+
+    msg = "\n{}\n\n    {}\n\nEnter {} Token:".format(
+          "Please go to the following link in your browser:",
+          url,
+          token_type)
+    sys.stderr.write(msg)
+    sys.stderr.flush()
+    token = sys.stdin.readline().strip()
```
jgehrcke: I may not understand enough of the CLI architecture, but why don't we use Python's `input([prompt])` built-in?

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    return token
+
+
+def _get_dcostoken_by_post_with_creds(dcos_url, creds):
+    """
+    Get DC/OS Authentication token by POST to `acs/api/v1/auth/login`
+    with specific credentials (either OIDC token or uid/password)
```
jgehrcke: `OIDC token` is the OIDC ID token used for open DC/OS w/ the implicit OIDC flow, correct? If that's true, let's add more commentary here while this is fresh in our heads :).

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    :param hostname: hostname for credentials
+    :type hostname: str
+    :returns: username, password
+    :rtype: str, str
+    """
+
+    if username is None:
+        sys.stdout.write("{}'s username: ".format(hostname))
+        sys.stdout.flush()
+        username = sys.stdin.readline().strip()
```
jgehrcke: Again... why not `input()`? :)

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+
+    :param dcos_url: url to cluster
+    :type dcos_url: str
+    :param creds: credentials to login endpoint
+    :type creds: {}
+    :returns: DC/OS Authentication Token
+    :rtype: str
+    """
+
+    url = urllib.parse.urljoin(dcos_url, 'acs/api/v1/auth/login')
```
jgehrcke: I think `urljoin()` is not the right tool. It's meant to perform much more complex logic than required here (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin). Therefore, it may lead to subtle bugs (we once had that with Marathon API URL construction).

From my understanding, this is more resilient: ensure(!) that `dcos_url` has a trailing slash and simply do `dcos_url + 'acs/api/v1/auth/login'` _or_ ensure that `dcos_url` does not have a trailing slash and do `dcos_url + '/acs/api/v1/auth/login'` (I like the latter more).

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    :param dcos_url: url to cluster
+    :type dcos_url: str
+    :rtype: None
+    """
+
+    dcos_token = _get_dcostoken_by_oidc_implicit_flow(dcos_url)
+    if not dcos_token:
+        raise DCOSException("Authentication failed")
+    else:
+        return
```
jgehrcke: Should `dcos_token` be `dcos_auth_token`?

Do we need to return it from this function?

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    Get DC/OS Authentication Token by OIDC implicit flow
+
+    :param dcos_url: url to cluster
+    :type dcos_url: str
+    :returns: DC/OS Authentication Token
+    :rtype: str
+    """
+
+    oauth_login = 'login?redirect_uri=urn:ietf:wg:oauth:2.0:oob'
+    url = urllib.parse.urljoin(dcos_url, oauth_login)
```
jgehrcke: I again believe that a simple string concatenation would remove a whole lot of magic and complexity here, and therefore room for bugs.

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    oidc_token = _prompt_user_for_token(url, "OpenID Connect ID")
```
jgehrcke: "OpenID Connect ID Token", as in http://openid.net/specs/openid-connect-core-1_0.html#IDToken

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    :param key_path: path to service key
+    :param key_path: str
+    :rtype: None
+    """
+
+    creds = {
+        'uid': username,
+        'token': jwt.encode(
+            {
+                'exp': int(time.time()+5*60),
```
jgehrcke: Can we add a comment that this `'token'` key in `creds` contains a "short-lived service login token"?

We _could_ make this live for just 10 seconds, but I think we do not want to require the clocks of the local machine to be _perfectly_ in sync with that of DC/OS nodes. So, I think that having a lifetime of 5 minutes is fine. Still, do we need or wand to add a help text / warning / comment somewhere that this method of login requires the time of the local machine to be (largely) in sync with that of the DC/OS nodes?

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+                'uid': username
+            },
+            util.read_file(key_path),
+            algorithm='RS256')
```
jgehrcke: Note that this might fail if `pyjwt` cannot import `cryptography`: https://github.com/jpadilla/pyjwt/blob/5caa1af9d57c3621596a12aeaa4693bda5f15596/jwt/algorithms.py#L31

Is that the first time that we need to bundle `cryptography` with the CLI binary? `cryptography` itself depends, by default, on OpenSSL, which also needs to be bundles with the CLI binary. We're already doing that with the security CLI.

Main point: this change may have a significant impact on the CLI binary build properties, and it adds new error paths for how the CLI may fail in certain environments.

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    :param dcos_url: url to cluster
+    :type dcos_url: str
+    :param provider_info: info about provider to auth with
+    :param provider_info: str
+    :rtype: None
+    """
+
+    start_flow_url = provider_info["client"]["start_flow_url"]
+    if not urlparse(start_flow_url).netloc:
+        start_flow_url = urljoin(dcos_url, start_flow_url)
```
jgehrcke: Again, I am strongly in favor of controlling the trailing slash behavior of `dcos_url` elsewhere plus using simple string concatenation here.

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+
+    # hit protected endpoint which will prompt for auth if cluster has auth
+    endpoint = '/pkgpanda/active.buildinfo.full.json'
+    url = urllib.parse.urljoin(dcos_url, endpoint)
+    response = http._request('HEAD', url)
+    auth_scheme = _get_auth_scheme(response)
+
+    i = 0
+    while i < 3 and response.status_code == 401:
```
jgehrcke: I think we should clean up this loop logic and add some commentary :-).

In the loop header we currently check `response.status_code`'s value, but in the body we only set it in one place where we immediate leave with a `return`, so this code is inconsistent. Maybe as a suggestion, Python's `for ... else` might be useful here, in the lines of:

```
for _ in range(3):
    ...
    if ....:
        break

else:
    # No break occurred 
    raise DCOSAuthenticationException
```

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+        if auth_scheme == "oauthjwt":
```
jgehrcke: I believe that for the case distinctions (as in `if auth_scheme == "oauthjwt":`) we should add some comments about what we believe to have learned about the cluster from here.

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+    Returns dict of providers configured on cluster
+
+    :returns: configured providers
+    :rtype: {}
+    """
+
+    dcos_url = config.get_config_val("core.dcos_url")
+    endpoint = 'acs/api/v1/auth/providers'
+    url = urllib.parse.urljoin(dcos_url, endpoint)
```
jgehrcke: Same thing.. :)

cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+    :param password_env: name of environment variable with password
+    :type password_env: str
+    :param password_file: path to file with password
+    :type password_file: bool
+    :returns: password or None if no password specified
+    :rtype: str | None
+    """
+
+    password = None
+    if password_str:
```
tamarrow: the user would have to specify `--password_str=""` to input an empty string, but I guess thats valid.

cli/dcoscli/auth/main.py
tamarrow: Good catch, adding.

cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+
+    password = _get_password(password_str, password_env, password_file)
+    if provider is None:
+        if password and username:
+            auth.dcos_uid_password_auth(dcos_url, username, password)
+        else:
+            try:
+                providers = auth.get_providers()
+                # Let users know if they have non-default providers configured
+                if len(providers) > 2:
```
tamarrow: 👍

cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+                    msg = ("Your cluster has several authentication providers "
+                           "enabled. Run `dcos auth list-providers` to see all"
+                           " providers and `dcos auth login --provider` to "
```
tamarrow: 👍

cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+    client_method = provider_info.get("client_method")
+    dcos_url = config.get_config_val("core.dcos_url")
+
+    if client_method == "browser-prompt-authtoken":
+        auth.by_browser_prompt_auth()
+    elif client_method == "browser-prompt-oidcidtoken-get-authtoken":
+        auth.oidc_implicit_flow_auth(dcos_url)
+    elif client_method == "dcos-credential-post-receive-authtoken":
+        if not username or not password:
+            msg = "Please specify --username and --password for provider [{}]"
```
tamarrow: i like it, updating

cli/dcoscli/auth/main.py
```diff
@@ -68,34 +75,139 @@ def _info(info):
...
+            raise DCOSException(msg.format(provider))
+        auth.dcos_uid_password_auth(dcos_url, username, password)
+    elif client_method == "dcos-servicecredential-post-receive-authtoken":
+        if not username or not key_path:
+            msg = "Please specify --username and --key_path for provider [{}]"
```
tamarrow: good catch

cli/dcoscli/data/help/auth.txt
```diff
@@ -18,5 +24,17 @@ Options:
         Print usage.
     --info
         Print a short description of this subcommand.
+    --password=<password>
+        Password specified as plaintext
```
tamarrow: yes, thanks for paying attention to this :)

dcos/auth.py
tamarrow: The two are not exactly interchangeable. `readline` reads an input stream and `input` reads a line and returns it as a str. `input` also returns carriage return in windows, and `readline` leaves in newline. 

Specifically for the CLI, i don't know why readline vs input, but we use readline for all other prompts, I was just being consistent. The CLI specific thing to be aware of is that subcommands can be forked processes, but I would expect `input` to still work.

dcos/auth.py
tamarrow: done

dcos/auth.py
tamarrow: 👍

dcos/auth.py
tamarrow: yep

dcos/auth.py
tamarrow: fair point

dcos/auth.py
tamarrow: i had `_prompt_user_for_token` append "Token" but I realized that's pretty dense so now the whole string should be a parameter

dcos/auth.py
```diff
@@ -0,0 +1,308 @@
...
+
+def servicecred_auth(dcos_url, username, key_path):
+    """
+    Get DC/OS Authentication token by browser prompt
+
+    :param dcos_url: url to cluster
+    :type dcos_url: str
+    :param username: username user for authentication
+    :type username: str
+    :param key_path: path to service key
```
tamarrow: 👍

dcos/auth.py
tamarrow: comment added. If we see this becomes a problem we can rethink how much we communicate to the user.

dcos/auth.py
tamarrow: This is the first time that we require cryptography :/ I need to add tests that call that library so we can verify that it works on all platforms (especially windows). Depending on how those go, we may need to include more dependencies on the CIs, but hopefully this be minimal, or at least one time.

dcos/auth.py
tamarrow: done

dcos/auth.py
tamarrow: added!

dcos/auth.py
tamarrow: much cleaner :)

tamarrow:
@jgehrcke all your comments should be addressed!

cli/dcoscli/data/help/auth.txt
```diff
@@ -18,5 +24,17 @@ Options:
...
+    --password=<password>
+        Specify password on the command line (insecure).
+    --password-env=<password_env>
+        Specify environment variable name containing password.
+    --password-file=<password_file>
+        Specify path to file containing password.
+    --provider=<provider_id>
+        Specify authentication provider to login with.
+    --service-key=<key_path>
+        Specify path to file containing private key.
```
jgehrcke: containing _the_ private key? :)

cli/dcoscli/data/help/auth.txt
```diff
@@ -18,5 +24,17 @@ Options:
...
+    --username=<username>
+        Username to login with
```
jgehrcke: Maybe "Specify username to login with."?

cli/dcoscli/data/help/auth.txt
tamarrow: done

cli/dcoscli/data/help/auth.txt
tamarrow: done

andreimc:
This supports CLI only login that can be used in CI?

tamarrow:
@andreimc yes

andreimc:
@tamarrow great work would be great to get this in, we are in this crazy cycle of reseting the dcos auth token every 5 days on our CI :)

jgehrcke:
Hello @andreimc! Define "can be used in CI" :-) It all depends on which type of secret you are willing to store for how long where. I am not convinced that this change here (and also the bigger picture, including the corresponding back-end refactoring that we're planning to do in open DC/OS) will help you getting around _actively_ ensuring that you have a valid authentication token available in the CI.

If you'd like to, you could bring up your use case on the DC/OS users mailing list (or did you already?) for further discussion.

andreimc:
@jgehrcke so our CI is set up to do auto-deployments, once our tests pass we want to deploy our apps to production. CLI login should be an available feature because I imagine a lot of ppl would love to automate deployments.

cli/dcoscli/data/help/auth.txt
```diff
@@ -18,5 +24,17 @@ Options:
         Print usage.
     --info
         Print a short description of this subcommand.
+    --password=<password>
+        Specify password on the command line (insecure).
+    --password-env=<password_env>
+        Specify environment variable name containing password.
```
joel-hamill: ...`variable name that contains the password.`

cli/dcoscli/data/help/auth.txt
```diff
@@ -18,5 +24,17 @@ Options:
...
+    --password-file=<password_file>
+        Specify path to file containing password.
+    --provider=<provider_id>
+        Specify authentication provider to login with.
```
joel-hamill: ...`provider to use for login.`

tamarrow:
@joel-hamill thanks for the docs improvements :)

jgehrcke:
@joel-hamill, thanks from my side, too :-)!

Tamar, I have just one cosmetic question about the `dcos auth list-providers` table: Do we want to have the `-` in the headings `PROVIDER-ID` and `AUTHENTICATION-TYPE` or do we instead want to display a space instead? I tend towards displaying a space, since this output is not meant to be machine-readable.

jgehrcke:
:rocket:

PR #840, opened 2016-11-22
dcos experimental package add

larioj:


dcos/cosmospackage.py
```diff
@@ -296,16 +299,34 @@ def remove_repo(self, name):
         response = self.cosmos_post("repository/delete", params=params)
         return response.json()
 
+    def experimental_package_add(self, dcos_package):
```
jsancio: I would change this to just `package_add`. As we are implementing these commands I would keep in mind that we would like to make it very easy to move them from `dcos experimental ...` to `dcos ...`.

dcos/cosmospackage.py
```diff
@@ -296,16 +299,34 @@ def remove_repo(self, name):
...
+        request = "add"
+        headers = self._request_preferences().get(request).pop(0)
+        headers['X-Dcos-Content-Length'] = str(_file_length(dcos_package))
+        data = _chunk_file(dcos_package)
+        response = self._post(request, params=None, headers=[headers], data=data)
+        return response
+
     @cosmos_error
-    def _post(self, request, params, headers=None):
+    def _post(self, request, params, headers=None, data=None):
```
jsancio: This means that `params` should be optional, no?

dcos/cosmospackage.py
```diff
@@ -296,16 +299,34 @@ def remove_repo(self, name):
...
+        Adds a DC/OS package to DC/OS
+
+        :param dcos_package: path to the DC/OS package
+        :type dcos_package: str
+        :return: Response to the package add request
+        :rtype: Response
+        """
+        request = "add"
+        headers = self._request_preferences().get(request).pop(0)
+        headers['X-Dcos-Content-Length'] = str(_file_length(dcos_package))
```
cruhland: We don't need this anymore; it was a workaround to provide the content length for chunked requests.

dcos/cosmospackage.py
```diff
@@ -576,12 +600,18 @@ def _get_cosmos_header(request_name, version):
...
-    request_name = request_name.replace("/", ".")
-    return {"Accept": _get_header("{}-response".format(request_name),
-                                  version),
-            "Content-Type": _get_header("{}-request".format(request_name),
-                                        "v1")}
+    if request_name == "add":
+        return {
+            'Accept': 'application/vnd.dcos.package.add-response+json;charset=utf-8;version=v1',
+            'Content-Type': 'application/vnd.dcos.universe.package+zip;version=v1',
+            'Transfer-Encoding': 'chunked',
```
cruhland: This should be left out.

dcos/cosmospackage.py
```diff
@@ -704,3 +734,38 @@ def _v2_package_to_v1_package_json(package_info):
...
+    :type filename: str
+    :return: length of the file whose path is filename
+    :rtype: int
+    """
+    with util.open_file(filename, 'r+b') as file:
+        file.seek(0, 2)
+        return file.tell()
+
+
+def _chunk_file(filename, chunk_size=200):
```
cruhland: Replace with file object?

dcos/cosmospackage.py
```diff
@@ -562,8 +583,11 @@ def _get_header(request_type, version):
     :rtype: str
     """
 
-    return ("application/vnd.dcos.package.{}+json;"
-            "charset=utf-8;version={}").format(request_type, version)
+    if request_type == 'add-response':
+        return 'application/vnd.dcos.universe.package+json;charset=utf-8;version=v3'
```
jsancio: Why are we mapping `add-reponse` to `.../vnd.dcos.universe.package...`? This should not be a valid type returned by Cosmos. The add command in Cosmos should  always return a `application/vnd.dcos.package.add-response+json;charset=utf-8;version=v1`. cc @cruhland

dcos/cosmospackage.py
```diff
@@ -576,12 +600,18 @@ def _get_cosmos_header(request_name, version):
...
+            'Accept': 'application/vnd.dcos.package.add-response+json;charset=utf-8;version=v1',
+            'Content-Type': 'application/vnd.dcos.universe.package+zip;version=v1',
+            'Transfer-Encoding': 'chunked',
+        }
+    else:
+        request_name = request_name.replace("/", ".")
+        return {"Accept": _get_header("{}-response".format(request_name),
+                                      version),
+                "Content-Type": _get_header("{}-request".format(request_name),
+                                            "v1")}
```
jsancio: Not related to this PR but can you please fix the indentation to match your indentation style.

dcos/cosmospackage.py
```diff
@@ -576,12 +600,18 @@ def _get_cosmos_header(request_name, version):
...
-    return {"Accept": _get_header("{}-response".format(request_name),
-                                  version),
-            "Content-Type": _get_header("{}-request".format(request_name),
-                                        "v1")}
+    if request_name == "add":
+        return {
+            'Accept': 'application/vnd.dcos.package.add-response+json;charset=utf-8;version=v1',
+            'Content-Type': 'application/vnd.dcos.universe.package+zip;version=v1',
+            'Transfer-Encoding': 'chunked',
+        }
```
jsancio: Okay for now. In the near future we are going have to both dcos package add and universe coordinate add.

dcos/cosmospackage.py
```diff
@@ -704,3 +734,38 @@ def _v2_package_to_v1_package_json(package_info):
...
+    Gets the length of a file
+
+    :param filename: a path to a file
+    :type filename: str
+    :return: length of the file whose path is filename
+    :rtype: int
+    """
+    with util.open_file(filename, 'r+b') as file:
+        file.seek(0, 2)
+        return file.tell()
```
jsancio: You can use `https://docs.python.org/3.6/library/os.path.html#os.path.getsize`. When we move to now using chunked encoding please investigate if requests is able to populate the `content-length` automatically when you pass a file object.

dcos/cosmospackage.py
larioj: Yes, this is a workaround for the current version. @cruhland has already told me this will change.

dcos/cosmospackage.py
```diff
@@ -310,13 +310,14 @@ def experimental_package_add(self, dcos_package):
         """
         request = "add"
         headers = self._request_preferences().get(request).pop(0)
-        headers['X-Dcos-Content-Length'] = str(_file_length(dcos_package))
-        data = _chunk_file(dcos_package)
-        response = self._post(request, params=None, headers=[headers], data=data)
+        with util.open_file(dcos_package, 'r+b') as pkg:
```
jsancio: Use `'rb'`. You don't need to write to the file: https://docs.python.org/2/library/functions.html#open

dcos/cosmospackage.py
```diff
@@ -582,9 +583,8 @@ def _get_header(request_type, version):
     :returns: header information
     :rtype: str
     """
-
-    if request_type == 'add-response':
-        return 'application/vnd.dcos.universe.package+json;charset=utf-8;version=v3'
+    if request_type == 'add-request':
```
jsancio: FYI, this won't work when we have to implement universe install. There we are going to use `add-request`.

dcos/cosmospackage.py
```diff
@@ -296,16 +299,35 @@ def remove_repo(self, name):
...
+        :type dcos_package: str
+        :return: Response to the package add request
+        :rtype: Response
+        """
+        request = "add"
+        headers = self._request_preferences().get(request).pop(0)
+        with util.open_file(dcos_package, 'rb') as pkg:
+            headers['Content-MD5'] = util.md5_hash_file(pkg)
+        with util.open_file(dcos_package, 'rb') as data:
+            response = self._post(request, headers=[headers], data=data)
```
jsancio: How about `return self._post(...)`?

larioj:
The integration test won't pass locally until the response type gets changed to add-response rather than universe-package. I could write tests, and ignore them, but I don't see much difference between that and writing them once everything has been integrated. Either way is fine by me.

dcos/package.py
```diff
@@ -169,3 +169,34 @@ def merge_installed(apps, subcommands, app_only, cli_only):
...
+            raise config.missing_config_exception(["core.dcos_url"])
+    return cosmos_url
+
+
+def get_package_manager():
+    """Returns type of package manager to use
+
+    :returns: PackageManager instance
+    :rtype: PackageManager
+    """
```
jsancio: Didn't we move this from `.../package/main.py` or something like that? Let's delete that code and change all of the call sites.

dcos/package.py
```diff
@@ -169,3 +169,34 @@ def merge_installed(apps, subcommands, app_only, cli_only):
         merged.extend(indexed_subcommands.values())
 
     return merged
+
+
+def get_cosmos_url():
+    """
+    :returns: cosmos base url
+    :rtype: str
+    """
```
jsancio: Didn't we move this from `.../package/main.py` or something like that? Let's delete that code and change all of the call sites.

larioj: Done

dcos/package.py
larioj: Done

PR #841, opened 2016-11-22
Add some missing newlines between import sections

cmaloney:
Builtin, third party, and first party are all supposed to have a newline between
them. Most of the codebase was doing this, but some had snuck through. This
updates those ones to follow the coding style. These are caught by recent
changes in the code style checking plugin making it more strict / accurate.


jsancio:
Thanks @cmaloney! Merged.

PR #842, opened 2016-11-23
Fix readme bug

larioj:


larioj:
Making this against master

PR #846, opened 2016-11-28
Refactor cosmospackage.py

larioj:
Factored out code for talking to cosmos into its own class. That way that code can be utilized to talk to cosmos in general, rather than just for the packaging commands. Renamed cosmospackage to packagemanager and Cosmos to PackageManager. Created ServiceManager.

### DIFF
--- a/dcos/cosmos.py
+++ b/dcos/cosmos.py
@@ -125,6 +125,9 @@ class Cosmos:
         :return: response returned by calling cosmos at url
         :rtype: requests.Response
         """
+        if not self._endpoint_exists(endpoint):
+            raise DCOSException(
+                'Cosmos called with incorrect endpoint {}'.format(endpoint))
         try:
             headers = headers_preference[0]
             if http_request_type is 'post':
@@ -216,6 +219,17 @@ class Cosmos:
             'Accept': self._get_accept(endpoint, version)
         }
         return remove_nones(merge_dict(simple_header, headers))
+
+    def _endpoint_exists(self, endpoint):
+        """
+
+        :param endpoint: a possible cosmos endpoint
+        :type endpoint: str
+        :return: true if endpoint is a valid cosmos endpoint,
+        false otherwise
+        :rtype: bool
+        """
+        return endpoint in self._request_versions
 
     def _get_accept(self, endpoint, version):
         """
--- a/dcos/cosmos.py
+++ b/dcos/cosmos.py
@@ -25,8 +25,23 @@ class Cosmos(object):
             self.cosmos_url = cosmos_url
 
         def _data(versions, http_method):
+            """
+            Create an object with the data for an endpoint.
+
+            :param versions: the accept versions in order
+            of priority.
+            :type versions: list[str]
+            :param http_method: should be 'post' or 'get'
+            :type http_method: str
+            :return:
+            """
             return {'versions': versions, 'http_method': http_method}
 
+        # This structure holds information about an endpoint. Currently
+        # the information stored is the return type versions, and the
+        # http request method. These two field should be stored in a
+        # dictionary of the form:
+        # {'versions': versions, 'http_method': http_method}.
         self._endpoint_data = {
             'capabilities': _data(['v1'], 'get'),
             'package/add': _data(['v1'], 'post'),
@@ -107,9 +122,7 @@ class Cosmos(object):
         :return: the Response object returned by cosmos
         :rtype: requests.Response
         """
-        if not self._endpoint_exists(endpoint):
-            raise DCOSException(
-                'Cosmos called with unexpected endpoint {}'.format(endpoint))
+        self._ensure_endpoint_exists(endpoint)
         url = self._get_endpoint_url(endpoint)
         request_versions = self._get_request_version_preferences(endpoint)
         headers_preference = list(map(
@@ -181,6 +194,21 @@ class Cosmos(object):
             else:
                 raise e
 
+    def _ensure_endpoint_exists(self, endpoint):
+        """
+        Will throw an error if endpoint is not part of the cosmos'
+        api.
+
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: nothing
+        :rtype: None
+        """
+        if not self._endpoint_exists(endpoint):
+            raise DCOSException(
+                'Programer error: no data for endpoint {}'.format(endpoint))
+
     def _get_endpoint_url(self, endpoint):
         """
         Gets the url for the cosmos endpoint 'endpoint'
@@ -204,11 +232,10 @@ class Cosmos(object):
         :return: list of versions in preference order
         :rtype: list[str]
         """
+        self._ensure_endpoint_exists(endpoint)
         versions = self._endpoint_data.get(endpoint).get('versions')
-
         if not versions:
-            raise DCOSException('Could not get versions')
-
+            raise DCOSException('Programer error: could not get versions')
         return versions
 
     def _get_http_method(self, endpoint):
@@ -220,11 +247,10 @@ class Cosmos(object):
         :return: http method type
         :rtype: str
         """
+        self._ensure_endpoint_exists(endpoint)
         method = self._endpoint_data.get(endpoint).get('http_method')
-
         if method is not 'post' and method is not 'get':
-            raise DCOSException('Could not get http_method')
-
+            raise DCOSException('Programmer error: could not get http_method')
         return method
 
     def _get_header(self, endpoint, version, headers=None):
--- a/dcos/cosmos.py
+++ b/dcos/cosmos.py
@@ -1,49 +1,13 @@
-import collections
-import functools
-
-import six
 from six.moves import urllib
 
 from dcos import config, http, util
-from dcos.errors import (DCOSAuthenticationException,
-                         DCOSAuthorizationException, DCOSBadRequest,
-                         DCOSException, DCOSHTTPException)
+from dcos.errors import DCOSBadRequest, DCOSException
 
 logger = util.get_logger(__name__)
 
 
-def cosmos_error(fn):
-    """Decorator for errors returned from cosmos
-
-    :param fn: function to check for errors from cosmos
-    :type fn: function
-    :rtype: Response
-    :returns: Response
-    """
-
-    @functools.wraps(fn)
-    def check_for_cosmos_error(*args, **kwargs):
-        """Returns response from cosmos or raises exception
-
-        :returns: Response or raises Exception
-        :rtype: valid response
-        """
-        response = fn(*args, **kwargs)
-        content_type = response.headers.get('Content-Type')
-        if content_type is None:
-            raise DCOSHTTPException(response)
-        elif _format_header_type('package/error', 'v1', '') in content_type:
-            logger.debug('Error: {}'.format(response.json()))
-            error_msg = _format_error_message(response.json())
-            raise DCOSException(error_msg)
-        return response
-
-    return check_for_cosmos_error
-
-
 class Cosmos:
-    """A wrapper on cosmos service, abstracts away http request,
-    endpoint headers, and versions"""
+    """A wrapper on cosmos that abstracts away http requests"""
 
     def __init__(self, cosmos_url=None):
         if cosmos_url is None:
@@ -66,7 +30,7 @@ class Cosmos:
             'package/uninstall': 'post'
         }
 
-        self._versions = {
+        self._request_versions = {
             'capabilities': ['v1'],
             'package/add': ['v1'],
             'package/describe': ['v2', 'v1'],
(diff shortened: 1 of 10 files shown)


jsancio:
@larioj Are you missing files? I can't seem to find the implementation for `call_cosmos_endpoint`?

larioj:
@jsancio It should be in cosmos.py

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: the value for the Content-Type header key for endpoint
+        :rtype: str
+        """
+        version = 'v1'
+        if (endpoint, version) in self._special_content_types:
+            return self._special_content_types[(endpoint, version)]
+        return format_cosmos_header_type(endpoint, version, 'request')
```
jsancio: To avoid the `version = v1` you can:

```
next(
  (value for ((current_endpoint, version), value) in _special_content_type.items() if current_endpoint == endpoint),
  None
)
```
Same as in Scala:
```
_special_content_types.find(
  case ((current_endpoint, _), _) => current_endpoint == endpoint
).map(case (_, value) => value)
```

dcos/packagemanager.py
```diff
@@ -308,12 +254,18 @@ def package_add(self, dcos_package):
...
-        headers = self._request_preferences().get(request).pop(0)
         with util.open_file(dcos_package, 'rb') as pkg:
-            headers['Content-MD5'] = util.md5_hash_file(pkg)
+            extra_headers = {'Content-MD5': util.md5_hash_file(pkg)}
         with util.open_file(dcos_package, 'rb') as data:
-            return self._post(request, headers=[headers], data=data)
+            try:
+                return self._post('add', headers=extra_headers, data=data)
+            except DCOSHTTPException as e:
+                if e.status() == 404:
```
tamarrow: `e.response.status_code`

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+                # reattempt with one less item in headers_preference
+                response = self._call_cosmos(url,
+                                             http_request_type,
+                                             headers_preference[1:],
+                                             data,
+                                             json,
+                                             **kwargs)
+            else:
+                raise e
+        return response
```
jsancio: I would just replace `response =` with `return` and remove this line. It would make it tail-recursive and allow Python to remove the stack if it does such optimizations. :)

tamarrow:
@larioj do you want to merge this into master or new-install?

dcos/servicemanager.py
```diff
@@ -0,0 +1,43 @@
...
+        :type app_id: None | str
+        :return: json encoded response of cosmos' service/start endpoint
+        :rtype: dict
+        """
+        endpoint = 'service/start'
+        json = {
+            'packageName': package_name,
+            'packageVersion': package_version,
+            'options': options,
+            'appId': app_id
```
jsancio: Actually, let's remove this (`appId`) from here and Cosmos. Please file an issue against Cosmos and tag it for 1.10.

dcos/packagemanager.py
```diff
@@ -308,12 +254,18 @@ def package_add(self, dcos_package):
         :return: Response to the package add request
         :rtype: Response
         """
-        request = "add"
-        headers = self._request_preferences().get(request).pop(0)
         with util.open_file(dcos_package, 'rb') as pkg:
-            headers['Content-MD5'] = util.md5_hash_file(pkg)
+            extra_headers = {'Content-MD5': util.md5_hash_file(pkg)}
         with util.open_file(dcos_package, 'rb') as data:
```
tamarrow: why are you opening the same file twice?

larioj:
@tamarrow I'd like to get it into new-install

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+            'package/render': ['v1'],
+            'package/repository/add': ['v1'],
+            'package/repository/delete': ['v1'],
+            'package/repository/list': ['v1'],
+            'package/search': ['v1'],
+            'package/uninstall': ['v1'],
+            'service/start': ['v1']
+        }
+
+        self._special_content_types = {
```
tamarrow: lets call these experimental to be consistent with the command

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        :return: the value for the Content-Type header key for endpoint
+        :rtype: str
+        """
+        version = 'v1'
+        if (endpoint, version) in self._special_content_types:
+            return self._special_content_types[(endpoint, version)]
+        return format_cosmos_header_type(endpoint, version, 'request')
+
+
+def format_cosmos_header_type(endpoint, version, suffix):
```
tamarrow: lets rename this to `get_cosmos_header`

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        :param headers: extra keys for the header
+        :type headers: dict[str, str]
+        :return: a header that can be used to query cosmos at endpoint
+        :rtype: dict[str, str]
+        """
+        simple_header = {
+            'Content-Type': self._get_content_type(endpoint),
+            'Accept': self._get_accept(endpoint, version)
+        }
+        return remove_nones(merge_dict(simple_header, headers))
```
tamarrow: when will there be Nones?

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        Gets the http method cosmos expects for the endpoint
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: http method type
+        :rtype: str
+        """
+        method = self._http_method.get(endpoint)
+
+        if method is not 'post' and method is not 'get':
```
tamarrow: if this were to happen, theres a bug here: https://github.com/dcos/dcos-cli/pull/846/files#diff-cbb7ec8379691d235f3e8db785de6a6aR22

You should just call `self._http_method.get(endpoint)` directly when needed

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+            if not _matches_expected_response_header(headers,
+                                                     response.headers):
+                raise DCOSException(
+                    'Server returned incorrect response type, '
+                    'expected {} but got {}'.format(
+                        headers.get('Accept'),
+                        response.headers.get('Content-Type')))
+        except DCOSBadRequest as e:
+            if len(headers_preference) > 1:
+                # reattempt with one less item in headers_preference
```
tamarrow: This shouldn't be necessary unless theres a bug in cosmos with content negotiation

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        # except 404, because then the url is fine, just not cosmos enabled
+        except DCOSHTTPException as e:
+            logger.exception(e)
+            return e.status() != 404
+        except Exception as e:
+            logger.exception(e)
+            return True
+        return response.status_code == 200
+
+    def call_cosmos_endpoint(self,
```
tamarrow: cosmos_request

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        http_request_type = self._get_http_method(endpoint)
+        return self._call_cosmos(
+            url,
+            http_request_type,
+            headers_preference,
+            data,
+            remove_nones(json),
+            **kwargs)
+
+    def _call_cosmos(self,
```
tamarrow: _cosmos_request

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: the url of endpoint
+        :rtype: str
+        """
+        return urllib.parse.urljoin(self.cosmos_url, endpoint)
+
+    def _get_request_version_preferences(self, endpoint):
```
tamarrow: the method name is the same length as the code in this method :) We should just call `self._request_versions.get(endpoin`

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+            'package/render': 'post',
+            'package/repository/add': 'post',
+            'package/repository/delete': 'post',
+            'package/repository/list': 'post',
+            'package/search': 'post',
+            'package/uninstall': 'post',
+            'service/start': 'post'
+        }
+
+        self._request_versions = {
```
tamarrow: we should use a dict that contains all info for an endpoint instead of duplicating the endpoints

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+        :type json: dict
+        :param kwargs: Additional arguments to requests.request
+                       (see py:func:`request`)
+        :type kwargs: dict
+        :return: the Response object returned by cosmos
+        :rtype: requests.Response
+        """
+        if not self._endpoint_exists(endpoint):
+            raise DCOSException(
+                'Cosmos called with incorrect endpoint {}'.format(endpoint))
```
tamarrow: `unexpected endpoint`?

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+    :param b: a dictionary, may be None
+    :type b: None | dict
+    :return: the result of merging a with b
+    :rtype: dict
+    """
+    if a is None and b is None:
+        return {}
+
+    if a is None:
+        return b
```
tamarrow: this is not a shallow copy

dcos/cosmos.py
larioj: I'm not sure, why we don't like 'version = v1'. The reason why I'm putting it like that is that for content-types, the version is always v1.

larioj: This may, of course, change in the future.

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+    :param suffix: The string that will be appended to
+    endpoint type, most commonly 'request' or 'response'
+    :type suffix: str
+    :return: a formatted value for a Content-Type or Accept header key
+    :rtype: str
+    """
+    prefix = endpoint.replace('/', '.')
+    separator = '-'
+    if suffix is None or suffix is "":
+        separator = ''
```
larioj: Thanks

dcos/packagemanager.py
```diff
@@ -39,7 +35,8 @@ def check_for_cosmos_error(*args, **kwargs):
         content_type = response.headers.get('Content-Type')
         if content_type is None:
             raise DCOSHTTPException(response)
-        elif _get_header("error", "v1") in content_type:
+        elif cosmos.format_cosmos_header_type(
+                "package/error", "v1", '') in content_type:
```
larioj: Okay

dcos/servicemanager.py
larioj: Done. DCOS-11815

dcos/packagemanager.py
larioj: I believe the hash is moving the file pointer.

dcos/packagemanager.py
larioj: Is this not the same thing?

dcos/cosmos.py
larioj: This aren't just for the experimental. This is for any content type type that doesn't conform to the usual format.

dcos/cosmos.py
larioj: Okay

dcos/cosmos.py
larioj: Okay

dcos/cosmos.py
larioj: This was part of the code that I factored out. The way that I understand it is we first try the highest version of the accept type, and then decrease, until we either run out of versions or we don't get a DCOSBadRequest.

dcos/cosmos.py
larioj: Fair enough.

dcos/cosmos.py
jsancio: I am okay either way.

dcos/packagemanager.py
jsancio: Yes. the seek pointer is move all the way to the end of file (EOF). You have to seek to the beginning or re-open it like you are doing.

dcos/cosmos.py
jsancio: I am okay with `call_endpoint` or `request`.

dcos/cosmos.py
larioj: Users can overwrite the default headers if they want to. For example, if I want a header that does not have a 'Content-Type'. I can just pass in a {'Content-Type': None}. Since users can do this, I would like to remove all nones.

dcos/cosmos.py
larioj: @jsancio suggested _format_media_type. I like that better.

dcos/cosmos.py
larioj: True...

dcos/cosmos.py
larioj: Okay

dcos/cosmos.py
larioj: Actually I'll keep this since we are now storing the versions in a more complicated data structure.

dcos/cosmos.py
```diff
@@ -0,0 +1,396 @@
...
+        if cosmos_url is None:
+            self.cosmos_url = get_cosmos_url()
+        else:
+            self.cosmos_url = cosmos_url
+
+        def _data(versions, http_method):
+            return {'versions': versions, 'http_method': http_method}
+
+        self._endpoint_data = {
+            'capabilities': _data(['v1'], 'get'),
```
tamarrow: 👍 Lets add a comment explaining the structure this expects.

dcos/cosmos.py
```diff
@@ -0,0 +1,396 @@
...
+    :type cosmos_url: str
+    """
+
+    def __init__(self, cosmos_url=None):
+        if cosmos_url is None:
+            self.cosmos_url = get_cosmos_url()
+        else:
+            self.cosmos_url = cosmos_url
+
+        def _data(versions, http_method):
```
tamarrow: pydoc

dcos/cosmos.py
```diff
@@ -0,0 +1,396 @@
...
+    def _get_http_method(self, endpoint):
+        """
+        Gets the http method cosmos expects for the endpoint
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: http method type
+        :rtype: str
+        """
+        method = self._endpoint_data.get(endpoint).get('http_method')
```
tamarrow: this will throw an exception if self._endpoint_data.get(endpoint) is None

dcos/packagemanager.py
```diff
@@ -658,6 +531,9 @@ def _format_error_message(error):
         error_message = _format_json_schema_mismatch_message(error)
     elif error.get("type") == "MarathonBadResponse":
         error_message = _format_marathon_bad_response_message(error)
+    elif error.get('type') == 'NotImplemented':
+        error_message = 'DC/OS has not been ' \
+                        'configured to support this operation: package add'
```
tamarrow: lets remove `package add` from the message. I assume this error type will be try for other operations as well

dcos/cosmos.py
tamarrow: By user you mean developer of the CLI. If the developer explicitly overrides the default content-type for whatever reason, we shouldn't remove it.

dcos/cosmos.py
tamarrow: i like `request` because its consistent with our http module

dcos/cosmos.py
```diff
@@ -0,0 +1,396 @@
...
+        :type json: dict
+        :param kwargs: Additional arguments to requests.request
+                       (see py:func:`request`)
+        :type kwargs: dict
+        :return: the Response object returned by cosmos
+        :rtype: requests.Response
+        """
+        if not self._endpoint_exists(endpoint):
+            raise DCOSException(
+                'Cosmos called with unexpected endpoint {}'.format(endpoint))
```
larioj: The message is confusing. I'll change it so that it's more apparent that this is a programming error, not a cosmos one.

dcos/cosmos.py
```diff
@@ -0,0 +1,401 @@
...
+    :param request_headers: the headers for a cosmos request
+    :type request_headers: dict[str, str]
+    :param response_headers: the headers for a cosmos response
+    :type response_headers: dict[str, str]
+    :return: true if the Content-Type value of the response header matches the
+    Accept value of the request header, false otherwise
+    :rtype: bool
+    """
+    return (request_headers.get('Accept')
+            in response_headers.get('Content-Type'))
```
jsancio: Okay.

jsancio:
LGTM! :ship: it.

dcos/cosmos.py
tamarrow: we should just remove it.

dcos/packagemanager.py
tamarrow: bump

dcos/cosmos.py
tamarrow: bump

dcos/cosmos.py
```diff
@@ -0,0 +1,396 @@
...
+        Gets the list of versions for endpoint in preference order.
+        The first item is most preferred, and last is least preferred.
+
+        :param endpoint: a cosmos endpoint, of the form 'x/y',
+        for example 'package/repo/add' or 'service/start'
+        :type endpoint: str
+        :return: list of versions in preference order
+        :rtype: list[str]
+        """
+        versions = self._endpoint_data.get(endpoint).get('versions')
```
tamarrow: bump

dcos/cosmos.py
tamarrow: bump

dcos/cosmos.py
larioj: This is a developer error, and we shouldn't be accounting for bad programming.

dcos/cosmos.py
larioj: This is a developer error, and we shouldn't be accounting for bad programming.

tamarrow:
@larioj you have a syntax error: 
`[18:16:49][Step 4/5] dcos/servicemanager.py:1:1: F401 'dcos.util' imported but unused`

PR #849, opened 2016-12-02
'package add' from universe

larioj:


cli/dcoscli/data/help/experimental.txt
```diff
@@ -4,8 +4,14 @@ Description:
 Usage:
     dcos experimental --help
     dcos experimental --info
-    dcos experimental package add <dcos-package>
-    dcos experimental package build [--output-directory=<output-directory>] <build-definition>
+    dcos experimental package add ((--dcos-package=<dcos-package>) |
```
tamarrow: don't need double parens for --dcos-pacakge and --package-name

cli/dcoscli/data/help/experimental.txt
```diff
@@ -14,18 +20,26 @@ Commands:
         Build a package to install to DC/OS or share with Universe.
```
tamarrow: lets specify that we are building locally

dcos/packagemanager.py
```diff
@@ -245,27 +245,42 @@ def remove_repo(self, name):
...
+                        'Content-MD5': util.md5_hash_file(pkg)
+                    }
+                with util.open_file(dcos_package, 'rb') as data:
+                    return self._post('add', headers=extra_headers, data=data)
+            else:
+                json = {'packageName': package_name}
+                if package_version is not None:
+                    json['packageVersion'] = package_version
+                return self._post('add', params=json)
+        except DCOSHTTPException as e:
```
tamarrow: this is too broad of an except. How are we determining this new feature. can we add a cosmos capability?

dcos/packagemanager.py
```diff
@@ -245,27 +245,42 @@ def remove_repo(self, name):
...
-            except DCOSHTTPException as e:
-                if e.status() == 404:
-                    message = 'Your version of DC/OS ' \
-                              'does not support this operation'
-                    raise DCOSException(message)
-                else:
-                    raise e
+        try:
+            if dcos_package:
+                with util.open_file(dcos_package, 'rb') as pkg:
```
tamarrow: lets add a comment explaining what this is doing

cli/dcoscli/data/help/experimental.txt
```diff
@@ -14,18 +20,26 @@ Commands:
...
 
 Options:
+    --dcos-package=<dcos-package>
+        Path to a DC/OS Package.
     -h, --help
         Print usage.
     --info
         Print a short description of this subcommand.
+    --local
+        Add a DC/OS package from the local disk.
```
cruhland: Is this supported?

cli/dcoscli/experimental/main.py
```diff
@@ -71,17 +72,22 @@ def _info():
...
+    :type package_name: None | str
+    :param package_version: the version of a remote DC/OS package
+    :type package_version: None | str
     :return: process status
     :rtype: int
     """
     package_manager = get_package_manager()
-    response = package_manager.package_add(dcos_package)
+    response = package_manager.package_add(
+        dcos_package, package_name, package_version)
```
cruhland: What do you think about doing some processing of the command-line arguments here, before calling methods on `package_manager`? For example, you could ensure that `dcos_package` and `package_name` are not both specified.

The reason for doing that is to decouple the `package_manager` interface from the command-line arguments, allowing the interface to be more direct. In fact, we should probably split `package_manager.package_add()` into two methods that each add from a single source.

cruhland:
@larioj there is an empty `test_experimental.py` file in this PR, can it be deleted if you don't plan to add code to it?

dcos/packagemanager.py
```diff
@@ -245,27 +245,42 @@ def remove_repo(self, name):
...
+                    extra_headers = {
+                        'Content-Type':
+                            'application/vnd.dcos.'
+                            'universe.package+zip;version=v1',
+                        'Content-MD5': util.md5_hash_file(pkg)
```
cruhland: It turns out that the `Content-MD5` header has been dropped from the HTTP standard. Although the S3 API still uses it, we might want to consider the alternatives discussed [here](http://stackoverflow.com/questions/8239873/message-integrity-check-with-http-headers-since-content-md5-was-deprecated).

dcos/packagemanager.py
cruhland: It looks like `util.md5_hash_file()` simply iterates over the file object but doesn't close it. So alternatively we could open the file just once, calculate the hash value, and then call `pkg.seek(0)` to reset the stream to the beginning.

cruhland:
@larioj back to you

dcos/packagemanager.py
larioj: @jsancio ?

larioj: We do handle all of those. They are part of cosmos_error. Except for the 404 for which I give a nicer exception in this context.

larioj: If you want to, I could move the 404 to cosmos_error, but the message would apply to all other endpoint too.

dcos/packagemanager.py
larioj: @cruhland I'll do the second option.

cli/dcoscli/experimental/main.py
larioj: Okay, I'll split them up, and do the checking.

larioj: Actually, I'm wary of checking that they are not both specified, because the library should take care of that. If at any point they are both specified that would be a programmer error, and I've been told by @tamarrow in the past that I should not code for programmer errors.

jsancio: I think @larioj is saying that `docopt` guarantee that both are not specified because of the docopt specification so we don't need to do it, no? If both are specified then it means either a bug in the command line specification or in docopt.

cruhland: Ah yes, that's a good point. Never mind then :)

cli/dcoscli/experimental/main.py
```diff
@@ -71,17 +72,24 @@ def _info():
...
+    :type package_version: None | str
     :return: process status
     :rtype: int
     """
     package_manager = get_package_manager()
-    response = package_manager.package_add(dcos_package)
+    if dcos_package:
+        response = package_manager.package_add_local(dcos_package)
+    else:
+        response = package_manager.package_add_remote(package_name, package_version)
```
cruhland: :+1:

PR #851, opened 2016-12-07
Remove skips for pods tests and a test affected by a Marathon bug

### LINKED ISSUE #783: Remove pods gating once pods have landed in testing/master

cruhland:
Fixes #783.

### DIFF
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -5,7 +5,7 @@ import time
 import six
 
 from dcos import emitting, util
-from dcos.errors import DCOSException
+from dcos.errors import DCOSException, DefaultError
 
 logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
@@ -117,7 +117,7 @@ def _output(curr_header, output_header, header, lines):
         if output_header and header != curr_header:
             emitter.publish('===> {} <==='.format(header))
         if lines == ['']:
-            emitter.publish(DCOSException('(No logs for this task)'))
+            emitter.publish(DefaultError('No logs for this task'))
         for line in lines:
             emitter.publish(line)
     return header
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -88,7 +88,9 @@ def _stream_files(curr_header, fn, mesos_files):
             reachable_files.remove(mesos_file)
             continue
 
-        if lines:
+        if lines == ['']:
+            emitter.publish(DCOSException('Empty file: {}'.format(mesos_file)))
+        elif lines:
             curr_header = _output(curr_header,
                                   len(reachable_files) > 1,
                                   six.text_type(mesos_file),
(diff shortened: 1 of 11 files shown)


cruhland:
@tamarrow can I delete the "CLI with pods" build configuration in TeamCity as well?

tamarrow:
Yes please!

cli/tests/integrations/test_task.py
```diff
@@ -132,11 +130,11 @@ def test_log_pod_task():
...
 
         # pod task log are not executor logs, so normal executor stderr
-        # logs shouldn't be seen and this pod shoudn't have any logging
+        # logs shouldn't be seen and this pod shouldn't have any logging
         # to stderr
-        assert returncode == 1
-        assert stderr == b'No files exist. Exiting.\n'
-        assert stdout == b''
+        assert returncode == 0
+        assert stderr == b''
```
cruhland: A `stderr` file is now being created for each pod task, so the command no longer fails to find it. The file stays empty, so it can still be distinguished from the executor's `stderr`. I'm not sure why `stdout` now has the newline, maybe because the command succeeds instead of failing? As the comment says, it's really the value of `stderr` that's the key assertion of this test.

tamarrow: stdout doesn't print a newline if it succeeds. It's probably doing a newline print of the the stderr file. If the file is empty, we should probably tell the user as much.

cruhland:
@tamarrow back to you!

cli/dcoscli/log.py
```diff
@@ -116,6 +116,8 @@ def _output(curr_header, output_header, header, lines):
     if lines:
         if output_header and header != curr_header:
             emitter.publish('===> {} <==='.format(header))
+        if lines == ['']:
+            emitter.publish(DCOSException('(No logs for this task)'))
```
tamarrow: We can use a `DefaultError` here instead of `DCOSException`. Lets remove the parens.

cruhland:
@tamarrow not sure why the windows test hasn't reported success, here is the latest run: https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/windows-binary/979/

cruhland:
@tamarrow this is really odd. It seems that several of the `dcos task` commands are getting different output with the OS X and Windows binaries than with the Linux binary. Have you seen anything like this before? What would be the best way for me to do manual investigation?

tamarrow:
Lets try to repro. You can ssh to the OSX agent to test mac and create a windows VM (or use our spare windows laptop) for windows. I'll DM you with more details

cruhland:
@tamarrow all green!

cruhland:
@tamarrow :ship: ?

PR #854, opened 2016-12-14
New install

larioj:
This is the feature branch for the new installing workflow. It adds a new subcommand called experimental, and three new commands bellow that. It includes test, although some are commented out as some functionality is not yet available in DCOS.

The three new commands are as follows:
```
dcos experimental package build 
dcos experimental package add
dcos experimental service start
```


tamarrow:
@larioj your rebase is wrong. you've picked up changes from other commits. @cruhland can you help jesus sort this out please.

PR #856, opened 2016-12-15
Updated `dcos task exec` ATTACH_CONTAINER_INPUT semantics.

klueska:
With the new semantics, an ATTACH_CONTAINER_INPUT call will be attempted
once before committing to a persistent connection where the actual input
is streamed to the agent.

This change is required to get around a limitation of the python client
libraries where an HTTP response is not read until after the entire
request is sent. IN our case, we *never* complete the request (until a
timeout), so we don't see the proper error unless we do things as
updated in this commit.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -1053,6 +1053,12 @@ class TaskIO(object):
         self.attach_input_event = threading.Event()
         self.attach_input_event.clear()
 
+        # Set up an event to block printing the output
+        # until an attach input event has successfully
+        # been established.
+        self.print_output_event = threading.Event()
+        self.print_output_event.clear()
+
         # Set up an event to block the main thread
         # from exiting until signaled to do so.
         self.exit_event = threading.Event()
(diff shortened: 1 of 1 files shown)


dcos/mesos.py
```diff
@@ -1291,6 +1311,19 @@ def _input_streamer():
         # `_process_output_stream` function signals us that it's ready.
         self.attach_input_event.wait()
 
+        # Send an intial "Test" message to ensure that we can establish a
+        # connection with the agent at all. If we can't we will thro an
```
tamarrow: s/thro/throw

dcos/mesos.py
```diff
@@ -1251,12 +1263,11 @@ def _attach_container_input(self):
...
-            """Generator function yielding ATTACH_CONTAINER_INPUT
-            messages for streaming. It creates the initial
-            ATTACH_CONTAINER_INPUT message to establish the
-            connection, then yields a message from the input_queue on
-            each subsequent call.
+        def _initial_input_streamer():
+            """Generator function yielding the intial ATTACH_CONTAINER_INPUT
+            message for streaming. We have a separate generator for this so
+            that we can attempt the connection once before committing to a
+            persistent connection where we stream the rest of the input.
```
tamarrow: pydoc

dcos/mesos.py
```diff
@@ -1271,6 +1282,15 @@ def _input_streamer():
 
             yield self.encoder.encode(message)
 
+        def _input_streamer():
+            """Generator function yielding ATTACH_CONTAINER_INPUT
+            messages for streaming. It yields the _intitial_input_streamer()
+            message, followed by messages from the input_queue on each
+            subsequent call.
+            """
```
tamarrow: pydoc

Issue #604, opened 2016-05-10
Default protocol when setting core.dcos_url


arun-gupta:
Setting the `core.dcos_url` as:

```
./bin/dcos config set core.dcos_url couchbase-ElasticL-UJ5K6BOZQX59-570103311.us-west-1.elb.amazonaws.com
```

gives the following error:

```
Please check url 'couchbase-ElasticL-UJ5K6BOZQX59-570103311.us-west-1.elb.amazonaws.com'. Missing http(s)://
```

CLI should take a reasonable default if no protocol is specified.

Please answer the following questions before submitting your issue. Thanks!
### What version of the DCOS CLI are you using (`dcos --version`)?

dcos version 0.4.4

PR #870, opened 2017-01-09
config: default `core.dcos_url` scheme to https

tamarrow:
#604

### DIFF
--- a/cli/tests/integrations/test_config.py
+++ b/cli/tests/integrations/test_config.py
@@ -66,12 +66,15 @@ def test_get_missing_property(env):
     _get_missing_value('missing.property', env)
 
 
-def test_invalid_dcos_url(env):
-    stderr = b'Please check url \'abc.com\'. Missing http(s)://\n'
-    assert_command(['dcos', 'config', 'set', 'core.dcos_url', 'abc.com'],
-                   stderr=stderr,
-                   returncode=1,
+def test_dcos_url_without_scheme(env):
+    old = b'http://dcos.snakeoil.mesosphere.com'
+    new = b"abc.com"
+    out = b"[core.dcos_url]: changed from '%s' to 'https://%s'\n" % (old, new)
+    assert_command(['dcos', 'config', 'set', 'core.dcos_url', new],
+                   returncode=0,
+                   stderr=out,
                    env=env)
+    config_set('core.dcos_url', old, env)
 
 
 def test_get_top_property(env):
--- a/dcos/jsonitem.py
+++ b/dcos/jsonitem.py
@@ -283,8 +283,8 @@ def _parse_url(value):
     if value_regex is None:
         scheme_match = re.match(scheme_pattern, value, re.IGNORECASE)
         if scheme_match is None:
-            msg = 'Please check url {!r}. Missing http(s)://'.format(value)
-            raise DCOSException(msg)
+            logger.debug("Defaulting URL to https scheme")
+            return "https://" + value
         else:
             raise DCOSException(
                 'Unable to parse {!r} as a url'.format(value))
--- a/tests/test_jsonitem.py
+++ b/tests/test_jsonitem.py
@@ -61,16 +61,6 @@ def bad_array(request):
 
 
 @pytest.fixture(params=[
-    'www.test.com',
-    'http:/hi.com',
-    'https//www.hi.com',
-    'http://bad.port:here'
-    ])
-def bad_url(request):
-    return request.param
-
-
-@pytest.fixture(params=[
     ('string', 'this is a string', 'this is a string'),
     ('string', 'null', None),
     ('object', '{"key":"value"}', {'key': 'value'}),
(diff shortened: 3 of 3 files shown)


jimenez:
LGTM

dcos/jsonitem.py
```diff
@@ -283,8 +283,9 @@ def _parse_url(value):
     if value_regex is None:
         scheme_match = re.match(scheme_pattern, value, re.IGNORECASE)
         if scheme_match is None:
-            msg = 'Please check url {!r}. Missing http(s)://'.format(value)
-            raise DCOSException(msg)
+            logger.debug("Defaulting URL to https scheme")
+            value = "https://" + value.lstrip(scheme_pattern)
```
jgehrcke: `value.lstrip(scheme_pattern)` is a funny bug! 

It will remove all characters contained in the character set defined by `scheme_pattern` from the trailing part of the partial URL:
```>>> set('^(?P<scheme>(?:(?:https?)://))')
{'<', 'p', 'h', 'e', ')', '/', 's', '(', '^', 'P', '?', '>', ':', 'c', 'm', 't'}
>>> value = 'powerdcos.com'
>>> scheme_pattern = r'^(?P<scheme>(?:(?:https?)://))'
>>> "https://" + value.lstrip(scheme_pattern)
'https://owerdcos.com'
```

jgehrcke: IMO the correct way to tackle this would be to use `urllib.parse.urlparse()`, to use `netloc` and `scheme` from the parse result tuple, and to re-assemble the URL using `urllib.parse.unparse()`.

That would get rid of all the regex craziness here.

cli/tests/integrations/test_config.py
```diff
@@ -66,12 +66,15 @@ def test_get_missing_property(env):
...
 
-def test_invalid_dcos_url(env):
-    stderr = b'Please check url \'abc.com\'. Missing http(s)://\n'
-    assert_command(['dcos', 'config', 'set', 'core.dcos_url', 'abc.com'],
-                   stderr=stderr,
-                   returncode=1,
+def test_dcos_url_without_scheme(env):
+    old = b'http://dcos.snakeoil.mesosphere.com'
+    new = b"abc.com"
+    out = b"[core.dcos_url]: changed from '%s' to 'https://%s'\n" % (old, new)
```
jgehrcke: Why do we define these as byte string literals?

dcos/jsonitem.py
jgehrcke: In addition, I do not understand why we want to "lstrip" the scheme here at all. Using the terminology used in this code section, we just confirmed that the string "value" does not have a "scheme pattern" in its beginning. So why would we try to strip it?

cli/tests/integrations/test_config.py
tamarrow: python3 fun!
https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit

dcos/jsonitem.py
tamarrow: I was trying to be have scheme_pattern "clean up" and invalid scheme. For example, value = "httpdcos.com" would be corrected to "https://dcos.com"

But you're right I should just add the scheme to the value as is.

cli/tests/integrations/test_config.py
jgehrcke: I asked because I don't like to seeing byte literals in Python code w/o _obvious_ reason :-) We can leave this as is, but I want to elaborate on my point, because I think it's a pretty damn important point in Python application design.

The thumb rule is to use _text_ processing _within_ the application, and to decode and encode _only_ at its boundaries where the IO is done (the communication with the subprocess in this case). That usually makes code way easier to understand, because the application code then usually only contains text objects (`str` in Python 3). If one does not follow that thumb rule, one ends up having a (randomly appearing) mixture of bytes/str objects within the application code, and it's hard to follow the _assumptions_ associated with the corresponding encoding and decoding approaches. The last sentence is really important. To clarify, I want to point out how exactly this code here relies out fancy stuff (which is non-obvious from the code):
* We rely on  `b'...'` literals to magically use the ASCII codec (try `b'A' == b'\x41'`).
* We rely on the subprocess `stdout` (type `bytes`) to be _compatible_ with the ASCII codec (we do a bytes comparison of encoded text, so we require the encoding to be the same in both operands of the comparison). Which _seems_ to be the case here, but I am quite sure it's not true in all cases: _most likely_ the subprocess uses UTF-8 to encode it's text when it emits it to stdout. And ASCII is a subset of UTF-8. That's why the comparison _seems_ to work here.

To make this clean, explicit, and expressive, a good strategy is to _explicitly instruct_ the invoked subprocess (which is the CPython interpreter here) how to encode its data on stdout (if that's Python, this can be done using the environment variable `PYTHONIOENCODING`). Imagine we set this to `UTF-8`. Then we have exact knowledge about how to interpret the data we receive from that subprocess (and can do something like `subprocess_stdout = subprocess_stdout_bytes.decode('utf-8')`). This we could then compare to the `str`(!) object containing `[core.dcos_url]: changed from '%s' to 'https://%s'\n" % (old, new)`.

PR #875, opened 2017-01-12
Added possibility to truncate a table cell

unterstein:
In the scenario having an app definition like with a real long `cmd` section, the command `dcos marathon app list` would be jumbled. Therefore it would be nice to have the possibility to truncate the cmd section.

Before the change:
```
dcos marathon app list
ID      MEM   CPUS  TASKS  HEALTH  DEPLOYMENT  WAITING  CONTAINER  CMD                                                                                                                                                                   
/sleep  128  0.001   1/1    ---       ---      False      mesos    sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000 && sleep 1000  
```

After the change:
```
dcos marathon app list
ID      MEM   CPUS  TASKS  HEALTH  DEPLOYMENT  WAITING  CONTAINER  CMD                                                                                                                                                                   
/sleep  128  0.001   1/1    ---       ---      False      mesos    sleep 1000 && sleep 1000 &&...  
```

### DIFF
--- a/cli/dcoscli/tables.py
+++ b/cli/dcoscli/tables.py
@@ -918,7 +918,7 @@ def truncate_table(fields, objs, limits, **kwargs):
     tb._left_padding_width = 0
     tb._right_padding_width = 2
 
-    def format(obj, key, function):
+    def format_table(obj, key, function):
         """Formats the given object for the given function
 
         :param object: object to format
@@ -936,7 +936,7 @@ def truncate_table(fields, objs, limits, **kwargs):
         return result
 
     for obj in objs:
-        row = [format(obj, key, fields.get(key)) for key in fields.keys()]
+        row = [format_table(obj, key, fields.get(key)) for key in fields.keys()]
         tb.add_row(row)
 
     return tb
--- a/cli/dcoscli/tables.py
+++ b/cli/dcoscli/tables.py
@@ -1,6 +1,9 @@
 import copy
 import datetime
 import posixpath
+
+import textwrap
+
 from collections import OrderedDict
 
 import prettytable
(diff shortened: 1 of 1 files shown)


kensipe:
@unterstein LGTM!   I would add a test though in `cli/tests/unit/test_tables.py` and/or `cli/dcoscli/tables.py`

cli/dcoscli/tables.py
```diff
@@ -112,7 +115,10 @@ def get_deployment(app):
         ("CMD", get_cmd)
```
tamarrow: we can just use `_truncate_desc` on `get_cmd` directly

unterstein: @kensipe and I discussed about introducing a general helper to enable this truncating feature. If this is too much, I just change to use the truncate functionality only for `cmd`. What do you think @tamarrow ?

tamarrow: 👍 can you update the uses of` _truncate_desc` to use your truncate function?

cli/dcoscli/tables.py
```diff
@@ -919,8 +918,45 @@ def table(fields, objs, **kwargs):
     tb._left_padding_width = 0
     tb._right_padding_width = 2
 
+    def format(obj, key, function):
```
tamarrow: lets call this format_table

unterstein: sure 👍

tamarrow:
unrelated failing test fixed in #885

PR #878, opened 2017-01-18
fix python version requirement

timcharper:
The docs say we need 3.4, but then make tells me 3.5.

One of these is wrong.

```
 ~/src/m8e/dcos-cli [master|…1]
14:59 $ make env
bin/env.sh
Cannot find supported python version 3.5. Exiting...
make: *** [env] Error 1
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Development Dependencies
 
 #. git_ must be installed to download the source code for the DC/OS CLI.
 
-#. python_ version 3.4.x must be installed.
+#. python_ version 3.5.x must be installed.
 
 #. virtualenv_ must be installed and on the system path in order to install
    legacy subcommands. New subcommands are packaged as platform specific
@@ -85,7 +85,7 @@ Running Tests
 Setup
 #####
 
-Tox, our test runner, tests against Python 3.4. We have a set of tests in
+Tox, our test runner, tests against Python 3.5. We have a set of tests in
 the :code:`dcos` package (root directory) and in the :code:`dcoscli` package
 (:code:`cli` directory). When running the tests describe below change
 directory to one of those two and follow the instructions.
@@ -114,7 +114,7 @@ to test SSL certs. To run this test suite be sure to add this resolution to your
 Running
 #######
 
-Tox will run unit and integration tests in Python 3.4 using a temporarily
+Tox will run unit and integration tests in Python 3.5 using a temporarily
 created virtualenv.
 
 You can set :code:`DCOS_CONFIG` to a config file that points to a DC/OS
--- a/cli/tests/integrations/test_ssl.py
+++ b/cli/tests/integrations/test_ssl.py
@@ -114,4 +114,5 @@ def _ssl_error_msg():
     return (
         "An SSL error occurred. To configure your SSL settings, please run: "
         "`dcos config set core.ssl_verify <value>`\n"
-        "<value>: Whether to verify SSL certs for HTTPS or path to certs\n")
+        "<value>: Whether to verify SSL certs for HTTPS or path to certs. "
+        "Valid values are True, False, or a path to a CA_BUNDLE.\n")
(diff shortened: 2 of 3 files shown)


timcharper:
Also added commit to make an error message easier to act on

dcos/http.py
```diff
@@ -97,7 +97,7 @@ def _request(method,
     except requests.exceptions.SSLError as e:
         logger.exception("HTTP SSL Error")
         msg = ("An SSL error occurred. To configure your SSL settings, "
-               "please run: `dcos config set core.ssl_verify <value>`")
```
tamarrow: This is not exactly correct. ssl_verify can be True|False or path to CA_BUNDLE. It is the same as `verify` here: http://docs.python-requests.org/en/master/user/advanced/?highlight=ssl#ssl-cert-verification

tamarrow:
@timcharper thanks for the pr! One small comment and then you'll need to update this tests here: https://github.com/dcos/dcos-cli/blob/master/cli/tests/integrations/test_ssl.py#L113

dcos/http.py
```diff
@@ -100,7 +100,8 @@ def _request(method,
                "please run: `dcos config set core.ssl_verify <value>`")
         description = config.get_property_description("core", "ssl_verify")
         if description is not None:
-            msg += "\n<value>: {}".format(description)
+            msg += "\n\ncore.ssl_verify: {}".format(description)
```
tamarrow: Lets keep this line the same. We print out `"please run: `dcos config set core.ssl_verify <value>`` right before this, so it makes sense to explain what "<value>" is.

dcos/http.py
```diff
@@ -100,7 +100,8 @@ def _request(method,
...
+        msg += "\n\npossible values: `True`, `False`, or a Path to CA"
```
tamarrow: Lets lowercase 'Path' and remove one of the newlines. 
CA should be `path to a CA_BUNDLE file`

tamarrow:
@timcharper some small nits and don't forget to update `test_ssl.test_verify_ssl_without_cert_env_var`.

timcharper:
You got it!

I couldn't figure out how to run the test_ssl tests properly; looks like I need to set up a snakeoil domain and server or something. Anyways, I'll check back in a bit to see if the build still fails and if I need to do anything else.

cli/tests/data/dcos.toml
```diff
@@ -1,5 +1,5 @@
 [core]
+timeout = 5
 dcos_url = "http://dcos.snakeoil.mesosphere.com"
```
tamarrow: please remove changes to this file

dcos/http.py
```diff
@@ -97,10 +97,11 @@ def _request(method,
     except requests.exceptions.SSLError as e:
         logger.exception("HTTP SSL Error")
         msg = ("An SSL error occurred. To configure your SSL settings, "
-               "please run: `dcos config set core.ssl_verify <value>`")
+               "please run: `dcos config set core.ssl_verify <value>`, where "
+               "<value> is `True`, `False`, or a path to a CA_BUNDLE file.")
         description = config.get_property_description("core", "ssl_verify")
```
tamarrow: @timcharper if we change the description here: https://github.com/dcos/dcos-cli/blob/3c95b2b80774d9d7b2888d7c88f4cdd42a405fe9/dcos/data/config-schema/core.json#L39 than we don't have to change this file at all :)

tamarrow:
@timcharper anything I can help with?

cli/tests/data/dcos.toml
timcharper: Done

dcos/http.py
timcharper: Done

dcos/http.py
timcharper: Done

PR #880, opened 2017-01-18
dcos/http: raise exception if expired token

darkonie:
- add raise DCOSAuthenticationException in dcos.http.request if token is
  invalid.
- raise DCOSAuthenticationException in
  dcos.packagemanager.PackageManager().has_capability if user is not logged in

### DIFF
--- a/dcos/packagemanager.py
+++ b/dcos/packagemanager.py
@@ -67,6 +67,10 @@ class PackageManager:
         try:
             response = self.cosmos.call_endpoint(
                 'capabilities').json()
+        except DCOSAuthenticationException:
+            raise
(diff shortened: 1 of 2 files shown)


dcos/packagemanager.py
```diff
@@ -67,6 +67,8 @@ def has_capability(self, capability):
         try:
             response = self.cosmos.call_endpoint(
                 'capabilities').json()
+        except DCOSAuthenticationException:
+            raise
```
tamarrow: lets also raise DCOSAuthorizationException here

darkonie: done

PR #888, opened 2017-01-24
`dcos job` without subcommands will now bring usage information

kensipe:
it is part of the docopt magic


tamarrow:
@kensipe nice catch. will merge once tests pass

PR #891, opened 2017-01-27
config: change show config to communicate env var overwrites

jsrodman:
$ DCOS_URL=http://loalhost dcos --log-level=debug config show
core.dcos_acs_token ********
core.dcos_url http://loalhost       # overwritten by environment var DCOS_URL; config file value: http://52.10.13.118/

### DIFF
--- a/cli/dcoscli/config/main.py
+++ b/cli/dcoscli/config/main.py
@@ -110,6 +110,72 @@ def _unset(name):
     return 0
 
 
+def _format_config(file_value, effective_value, name=None, envvar_name=None):
+    """
+    Construct a string to show on a terminal, indicating the value and
+    possibly other useful things such as the setting name and whether
+    it is being controlled by an environment variable.
+
+    >>> _format_config('x', 'x')
+    'x'
+    >>> _format_config('x', 'x', 'setting.name') ->
+    'setting.name x'
+    >>> _format_config('x', 'y', envvar_name='ENVVAR')
+    'y # overwritten by ENVVAR; config file value: x'
+    >>> _format_config('x', 'y', 'setting.name', envvar_name='ENVVAR')
+    'setting.name y   # overwritten by ENVVAR; config file value: x'
+
+    :param file_value: config value present in the toml file
+    :type file_value: str
+
+    :param effective_value: config value either from file or as overwritten
+                            from the environment
+    :type effective_value: str
+    :param name: config key (not used for single value show)
+    :type name: str|None
+    :param envvar_name: name of environment variable that overwote the value
+    :type envvar_name: str|None
+    :returns: formatted string for terminal
+    :rtype: str
+    """
+
+    # when declaring that vars are overwritten by the environment,
+    # line up those messages to this column (unless the var name is long)
+    overwite_msg_display_column = 35
+
+    # When showing all values, don't print the token value;
+    if name == "core.dcos_acs_token":
+        print_value = "*"*8
+    else:
+        print_value = effective_value
+
+    if file_value == effective_value:
+        if name:
+            return '%s %s' % (name, print_value)
+        else:
+            return effective_value
+    else:
+        if not envvar_name:
+            envvar_name = "N/A"  # this should never happen
+        if name:
+            s = '%s %s' % (name, print_value)
+        else:
+            s = effective_value
+
+        left_pad_fmt = '%-{}s'.format(overwite_msg_display_column)  # '%-35s'
+
+        msg_start = left_pad_fmt + ' # overwritten by environment var %s; '
+
+        if print_value != effective_value:
+            # We're obscuring the effective security token
+            # so don't report the file value either
+            msg = msg_start + "config file value differs"
+            return msg % (s, envvar_name)
+
+        msg = msg_start + 'config file value: %s'
+        return msg % (s, envvar_name, file_value)
+
+
 def _show(name):
     """
     :returns: process status
@@ -119,19 +185,35 @@ def _show(name):
     toml_config = config.get_config(True)
 
     if name is not None:
-        value = toml_config.get(name)
-        if value is None:
+        file_value = toml_config.get(name)
+        try:
+            # If the user presented a partial key name, eg 'core' when
+            # we have 'core.xyz'; we will get an exception here
+            effective_value, envvar_name = config.get_config_val_envvar(name)
+        except DCOSException as e:
+            # The expected case of a partial key name has special
+            # handling via this mechanism.
+            if isinstance(file_value, collections.Mapping):
+                exc_msg = config.generate_choice_msg(name, file_value)
+                raise DCOSException(exc_msg)
+            raise  # Unexpected errors, pass right along
+
+        if effective_value is None:
             raise DCOSException("Property {!r} doesn't exist".format(name))
-        elif isinstance(value, collections.Mapping):
-            raise DCOSException(config.generate_choice_msg(name, value))
         else:
-            emitter.publish(value)
+            msg = _format_config(file_value, effective_value,
+                                 envvar_name=envvar_name)
+            emitter.publish(msg)
+
     else:
         # Let's list all of the values
         for key, value in sorted(toml_config.property_items()):
-            if key == "core.dcos_acs_token":
-                value = "*"*8
-            emitter.publish('{} {}'.format(key, value))
+            file_value = toml_config.get(key)
+            effective_value, envvar_name = config.get_config_val_envvar(key)
+
+            msg = _format_config(file_value, effective_value, key,
+                                 envvar_name=envvar_name)
+            emitter.publish(msg)
 
     return 0
 
--- a/dcos/config.py
+++ b/dcos/config.py
@@ -55,9 +55,13 @@ def get_config(mutable=False):
     return load_from_path(path, mutable)
 
 
-def get_config_val(name, config=None):
-    """Returns the config value for the specified key. Looks for corresponding
-    environment variable first, and if it doesn't exist, uses the config value.
+def get_config_val_envvar(name, config=None):
+    """Returns a tuple of the config value for the specified key and
+    the name of any environment variable which overwrote the value
+    from the configuration.  If no environment variable applies, None
+    is returned for the environment variable name.
+    Looks for corresponding environment variable first, and if it
+    doesn't exist, uses the config value.
     - "core" properties get resolved to env variable DCOS_SUBKEY. With the
     exception of subkeys that already start with DCOS, in which case we look
     for SUBKEY first, and "DCOS_SUBKEY" second, and finally the config value.
@@ -68,7 +72,7 @@ def get_config_val(name, config=None):
     :param config: config
     :type config: Toml
     :returns: value of 'name' parameter
-    :rtype: str | None
+    :rtype: (str | None, str | None)
     """
 
     if config is None:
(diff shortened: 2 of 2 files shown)


jsrodman:
Apologies, there are a few awkward bits about this pull request.

1 - I don't know the git login of the fine programmer I wrote this with.  Given a list of my interviewers today, I could determine it.
2 - make test seems to assume virtualenv matches the python environment I was using to write and test; on this mac the default of virtualenv is 2.x and I was using 3.x.  I tested all the functionality manually but still there could be objections like PEP8 things.

jsrodman:
Unrelated; I don't know how you do license  / copyright / attribution, but I would like to formally declare that this (trivial) code can be considered licensed irrevocably, transferrably, and indefinitely to Mesosphere Inc and the DC/OS open source project.

jsrodman:
Aside: I'm not a big fan of this test I broke, mostly because I don't know what it is trying to test.
It did catch an unintended output change, so that's great, but it is not named in a way that suggests it's trying to validate error handling.  Single line of comment would go a long way.

```
def test_get_top_property(env):
    returncode, stdout, stderr = exec_command(
        ['dcos', 'config', 'show', 'core'], env=env)

    assert returncode == 1
    assert stdout == b''
    assert stderr.startswith(
        b"Property 'core' doesn't fully specify a value - "
        b"possible properties are:\n")
```

jsrodman:
I *think* the check failures are unrelated?

CI/teamcity says:
```
[08:16:24]+ make clean env
[08:16:30]  Failed building wheel for MarkupSafe
```
I don't *think* this is a markup enforcement step, but rather an attempt to build the tool.

linux-binary says:
```

[08:07:02]Unable to find image 'linux-binary:1235' locally
[08:07:03]Pulling repository docker.io/library/linux-binary
[08:07:04]docker: Error: image library/linux-binary:1235 not found.
[08:07:04]See 'docker run --help'.
[08:07:04]Process exited with code 125
```
This one seems very definitely not related.

dcos/config.py
```diff
@@ -55,9 +55,13 @@ def get_config(mutable=False):
     return load_from_path(path, mutable)
 
 
-def get_config_val(name, config=None):
-    """Returns the config value for the specified key. Looks for corresponding
-    environment variable first, and if it doesn't exist, uses the config value.
+def get_config_val_and_envvar(name, config=None):
+    """Returns a tuple of the config value for the specified key and
```
tamarrow: config value returned is the overwritten value, correct?

dcos/config.py
```diff
@@ -85,7 +89,27 @@ def get_config_val(name, config=None):
...
+    - everything else gets resolved to DCOS_SECTION_SUBKEY
+
+    :param name: name of paramater
+    :type name: str
+    :param config: config
+    :type config: Toml
+    :returns: value of 'name' parameter
+    :rtype: str | None
+    """
+    val, ignore = get_config_val_and_envvar(name, config=None)
```
tamarrow: lets use `val, _ = ` here

cli/dcoscli/config/main.py
```diff
@@ -109,6 +109,38 @@ def _unset(name):
...
+    :param file_value: config value present in the toml file
+    :type file_value: str
+
+    :param effective_value: config value either from file or as overwritten from
+                            the environment
+    :type effective_value: str
+    :param name: config key (not used for single value show)
+    :type name: str|None
+    :param envvar_name: name of environment variable that overwote the value
+    :type name: str|None
```
tamarrow: s/name/envvar_name

cli/dcoscli/config/main.py
```diff
@@ -109,6 +109,38 @@ def _unset(name):
 
     return 0
 
+def _format_config(file_value, effective_value, name=None, envvar_name=None,
+                   left_pad_fmt=None):
```
tamarrow: lets remove `left_pad_fmt` from a param and create the format pattern in this function. since this function "formats" it makes sense to include it directly

cli/dcoscli/config/main.py
```diff
@@ -109,6 +109,38 @@ def _unset(name):
...
+    :returns: formatted string for terminal
+    :rtype: str
+    """
+    if file_value == effective_value:
+        if name:
+            return '%s %s' % (name, effective_value)
+        else:
+            return effective_value
+    else: # file was overwritten by environment
+        assert envvar_name and left_pad_fmt # envvar_name & left_pad_fmt needed
```
tamarrow: lets check and raise an exception here instead.

tamarrow:
ok to test

@jsrodman thanks for contributing and continuing what we worked on yesterday! I've left a few comments.

 Looks like you have syntax errors. See [here](https://teamcity.mesosphere.io/viewLog.html?buildId=542965&buildTypeId=DcosIo_DcosCli_Ci&tab=buildLog#_focus=1382). If you run tox -e py35-syntax (from the top level directory) it will run the syntax checker we use in the CI. 

The linux CI failed building (unrelated to your change). I'm re-running the CI.

cli/dcoscli/config/main.py
jsrodman: Currently the password case is in the _show function, because it's special logic as regards value and effective_value, which means duplication, or just going back to the inline '%-35s', not sure which is least ugly.  I'll duplicate the vars too if that seems best.

jsrodman: Alternatively, the caller could simply tell the formatter whether it's an override case, which would enable password formatting to be handled in this function too.

dcos/config.py
jsrodman: both get_config_val and get_config_val_and_envvar are returning the same value, which is the effective_value in the overwrite case (the one from the environment variable).  Neither of these return the original value in the case that the environment variable triggers an overwrite.  

(sorry, wasn't sure entirely sure what "the overwritten value" meant.)

jsrodman:
@tamarrow : i tried to raise some questions in the review yesterday, but didn't know what "pending" meant, until now.

As for the syntax checker, I did try, but ran into errors for both py35-syntax and py36-syntax.  Today, though, I think I figured out the right edits to tox.ini, should be correct now.

Putting off a squash for now.

cli/dcoscli/config/main.py
jsrodman: Thanks.

cli/dcoscli/config/main.py
jsrodman: I decided to move all the logic into _format_config; including token obscuring. It seems to be ok.

cli/dcoscli/config/main.py
jsrodman: I went with a ValueError exception.

cli/dcoscli/config/main.py
```diff
@@ -118,20 +150,46 @@ def _show(name):
 
     toml_config = config.get_config(True)
 
+    # when declaring that vars are overwritten by the environment,
+    # line up those messages to this column (unless the var is long)
+    overwite_msg_display_column = 35
+    left_pad_fmt = '%-{}s'.format(overwite_msg_display_column) # '%-35s'
```
jsrodman: No longer relevant.

dcos/config.py
jsrodman: Done.

jsrodman:
This might be done, but having trouble running the test suite locally.  Again believe I manually tested all intended cases.  If tests pass and thumbs up, I'll squash it all.

jsrodman:
Whoops, tox -e doesn't cover the whole project in one run.
Not sure about two complaints it has though, they seem worse than fixing them.

cli/dcoscli/config/main.py
```diff
@@ -119,19 +188,35 @@ def _show(name):
     toml_config = config.get_config(True)
 
     if name is not None:
-        value = toml_config.get(name)
-        if value is None:
+        file_value = toml_config.get(name)
+        try:
+            # If the user presented a partial key name, eg 'core' when
+            # we have 'core.xyz'; we will get an exception here
+            effective_value, envvar_name = config.get_config_val_and_envvar(name)
```
jsrodman: flake8.checker is unhappy about this, but I pretty much think it's just "wrong" in this case.
The goal of the width rule is

1 - no explosion in width, so you can reasonable read code.
2 - Don't go crazy on indentation
3 - Don't make overly complex expressions

On 1, this line is 81 characters.
On 2 we have a if -> try -> val = func(), hardly unreasonable indentation.
On 3, we have x, y = func(val); the only complexity here is the tuple unpack which is only made worse by indirection

The function name is unfortunately long which I could interperet this as fingering, but seemed the clearest way to present the distinction from the extremely similarly intended get_config_val().

Basically I don't see a better rendering of this line that is 2 characters shorter.  Is this a hard requirement in this project?

tamarrow: Yes. How about `get_config_val_info`? You can also use a continuation character ('\')

dcos/config.py
tamarrow: 👍

cli/dcoscli/config/main.py
```diff
@@ -110,6 +110,75 @@ def _unset(name):
...
+
+    if file_value == effective_value:
+        if name:
+            return '%s %s' % (name, print_value)
+        else:
+            return effective_value
+    else:
+        if not envvar_name:
+            exc_msg = "envvar_name must be provided if overwritten is true"
+            raise ValueError(exc_msg)
```
tamarrow: We should use `DCOSException` here since it's the baseline exception for the CLI (for exceptions we make). 

This message is confusing in the point of view for the user. Since this situation is caused by programmer error, we should log what happened `logger.debug(.....)`, and continue (returning `None`)

cli/dcoscli/config/main.py
jsrodman: I certainly can use a line continuation with \, but \ line continuations are a lot worse than long lines.  In some scenarios, open paren continuations are much better than long lines.

The proposed function name change seems OK, and the overlong function is kind of junk so will do.

cli/dcoscli/config/main.py
jsrodman: This seems to be against general python practice.  When you call a function, the expectation is it does what it says or you get an exception.  Returning None on a string-formatter is super weird.  Agreed this is not a reasonable user message, though making it log-and-continue means the programmer error is more likely to ship (which is why I started out with an assert).  If we want to treat this like a valid runtime condition, why not catch it in the caller and decide how to represent it there?

tamarrow: This is a private function, so we have a little more leeway here. I'd prefer to just have tests that cover this case. What about returning "N/A" as envvar if its not given? This is a programmer error vs an API/network/error so would be covered by tests rather than in code. But at least N/A is still providing some feedback to the user.

tamarrow: What do you think?

jsrodman: Using a presumed value seems OK, though i think it makes the expected usage pattern less clear.  I was sort of tempted to change the argument into a tuple, but icky in another way.

N/A is OK by me.

tamarrow: sounds good. Once you update this, you have two small syntax nits (reported from the CI), lets :shipit:

tamarrow:
```
[18:33:21][Step 4/5] dcoscli/config/main.py:195:80: E501 line too long (81 > 79 characters)
[18:33:21][Step 4/5] dcoscli/config/main.py:215:80: E501 line too long (80 > 79 characters)
```

tamarrow: @jsrodman bump :)

jsrodman:
I have a huge pile of failures locally because I don't have a cluster to talk to, but I think I've fixed all the ones related to the changes I made.

PR #893, opened 2017-01-31
http: only request with DC/OS Auth if request is to DC/OS cluster

tamarrow:


jgehrcke:
I think we should match both, the scheme _and_ network location, not just the network location. The goal is that if an authentication token has been received through HTTPS, the CLI only sends it through HTTPS. If that is maintained otherwise (through a mechanism that I don't see here), that's also OK.

Another thing I'd love us to ensure is that there is a coupling between the stored authentication token and the cluster URL. Do we currently remove the `core.dcos_acs_token` config key upon changing the `core.dcos_url` config key? If we don't yet, I think we should! That tightly couples the currently configured auth token to the currently configured cluster and ensures that the CLI does not accidentally send an authentication token obtained from cluster A to cluster B (after updating `core.dcos_url`).

tamarrow:
Thanks for the feedback. I've updated both suggestions 1. testing scheme + netloc 2. removing `core.dcos_acs_token` when we update `core.dcos_url`. 

I've thought through the implications of removing the token when we remove `dcos_url` and it should be fine. The user message when removing `dcos_url` informs the user that `acs_token` is removed as well, which should help the user realize the token is tied to the url. In terms of functionality if you remove the `dcos_url`, the `dcos_acs_token` is not valid and will only cause us to leak the token to the wrong source. 

For removing the `dcos_acs_token` when we set/replace the `dcos_url` is a little more confusing for the user (until they realize that the two are tied together). The only thing that could break is scripts that set the `dcos_acs_token` before the `dcos_url` (and have a previous `dcos_url` set). However this should a small user change for a bigger security win. I've added a message to "login" when updating the `dcos_url` to increase clarity on what's happening.

jgehrcke:
Sounds great, thanks @tamarrow!

dcos/http.py
```diff
@@ -92,7 +95,7 @@ def _request(method,
             url=url,
             timeout=timeout,
             auth=auth,
-            verify=verify,
+            verify=False,
```
jgehrcke: Was that change just for development?

tamarrow: yep...

PR #896, opened 2017-02-06
Updated HTTP headers for `dcos task exec` based on Mesos changes.

klueska:
This PR is dependent on https://github.com/dcos/dcos/pull/1179
It should not be merged until this PR has landed.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -1214,8 +1214,8 @@ class TaskIO(object):
             'stream': True,
             'headers': {
                 'Content-Type': 'application/json',
-                'Accept': 'application/json+recordio',
-                'connection': 'keep-alive'}}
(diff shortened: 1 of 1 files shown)


tamarrow:
@klueska once this is merged, how will old builds respond?

klueska:
They will break with an error of:
```

Error while fetching [
https://35.184.88.0/slave/3f67c585-735c-4c1f-8115-57aeecbea113-S0/api/v1]:
HTTP 406: Not Acceptable
```

tamarrow:
@klueska can we not support both? Or is this solely in 1.9 so no customers will be adversely affected?

klueska:
This is only in 1.9, so no one will be affected.

tamarrow:
👍

PR #897, opened 2017-02-07
Allow read-only permissions on ssh key

rchakra3:
I spent a while this morning trying to figure out why read-only permissions on my ssh key were "too open" :)


tamarrow:
ok to test

rchakra3:
@tamarrow done :)

PR #902, opened 2017-02-10
Add tests for dcos.packagemanager

msabramo:
I noticed that `dcos.packagemanager` had 0% test coverage. So I wrote some tests. If you like, I can write some more.

Before:

```
$ tox -e py35-unit | grep 'packagemanager.*%'
.tox/py35-unit/lib/python3.5/site-packages/dcos/packagemanager.py     252    252     0%
```

After:

```
$ tox -e py35-unit | grep 'packagemanager.*%'
.tox/py35-unit/lib/python3.5/site-packages/dcos/packagemanager.py     252    139    45%
```

Cc: @matthewdfuller

PR #903, opened 2017-02-10
tox.ini: Use --cov-report=term-missing

msabramo:
so we can see line numbers of uncovered lines

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,4 +29,4 @@ commands =
 
 [testenv:py35-unit]
 commands =
-  py.test -p no:cacheprovider -vv {env:CI_FLAGS:} --cov {envsitepackagesdir}/dcos tests{posargs}
+  py.test -p no:cacheprovider -vv {env:CI_FLAGS:} --cov {envsitepackagesdir}/dcos --cov-report=term-missing tests{posargs}


tamarrow:
I'm not convinced of the value of adding this to the CI report. It is buried in an already log build/test log and there is no action based on the outcome of the report. I would also argue that its the responsibility of the reviewer to make sure changes have tests. What do you think?

msabramo:
Yeah I don't think people will look at it in the CI logs. I just found it useful myself while developing locally and trying to increase coverage -- e.g.: #902 

The ideal thing to have I think is to have the CI measure the coverage and fail the build if the new or changed code is uncovered - see http://www.couverture.io/ (disclaimer: I am friends with the founder, @aconrad).

tamarrow:
Yes, something like that would be great. It would need to also detect code tested by integration tests, which is what the majority of our tests are.

aconrad:
If you are looking for a code coverage service, I'm currently looking for testers at www.couverture.io and I'll be more than happy to provide direct support. Here's a sample PR on how to do it (with Travis):

https://github.com/benoitc/gunicorn/pull/1456/files

But it seems that you are using TeamCity, in which case you'd want to simply go through the documentation of www.couverture.io and see how to hook up your repository to it.

aconrad:
@tamarrow I saw you were attempting to upload a coverage report to Couverture.io (thanks!) but you were getting a 404 when uploading to this URL:

```
https://app.couverture.io/coverage/github/tamarrow/dcos/dcos-cli/d41fd05225c84154b1f7215c1ef3c1a8b125bdc2/
```

It's indeed invalid because you are missing the job name at the end of the path which can be anything that describes the task that generated the coverage report. You could set it to simply `python` but since you are using tox, a common practice with to use the `$TOXENV` environment variable as the job name which will result in the following URL in your scenario:

```
https://app.couverture.io/coverage/github/tamarrow/dcos/dcos-cli/d41fd05225c84154b1f7215c1ef3c1a8b125bdc2/py35-unit
```

Checkout the docs which provide a sample curl upload command:

http://www.couverture.io/#docs

aconrad:
@tamarrow I realize the link I provided you with doesn't work, there's your username in it AND the org's name, `dcos`. It's one or the other. This would be more like it:

https://app.couverture.io/coverage/github/dcos/dcos-cli/d41fd05225c84154b1f7215c1ef3c1a8b125bdc2/py35-unit

tamarrow:
Thanks. We're not going to add anything now, but I'll check back when we have some time to look into different test coverage reports.

aconrad:
Sounds good. Remember the link above is meant to be used to upload the coverage report using a POST request. Just ping me if I can be of any help, otherwise the docs on the website should have you covered.

PR #906, opened 2017-02-13
tox.ini: Syntax check files in cli/

msabramo:
Cc: @tamarrow

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,9 +23,8 @@ deps =
   flake8
   flake8-import-order==0.9.2
   pep8-naming
-
 commands =
-  flake8 --verbose {env:CI_FLAGS:} dcos tests setup.py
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -23,9 +23,8 @@ deps =
   flake8
   flake8-import-order==0.9.2
   pep8-naming
-
 commands =
-  flake8 --verbose {env:CI_FLAGS:} dcos tests setup.py
+  flake8 --verbose {env:CI_FLAGS:} cli dcos tests setup.py
```
tamarrow: This is already tested in the other tox file (cli/tox.ini). We have two because we publish two packages when we release (dcos and dcoscli package).

msabramo: Ah okay I think I was/am confused about the two packages.

PR #910, opened 2017-02-17
dcos node ssh --private-ip=<private-ip>

ArmandGrillet:
People working on the soak clusters at Mesosphere receive alerts when
something unexpected is happening, e.g. 'Some containers failed to
launch on an agent on agent:<private_ip>'.

This new command allows us to quickly take a look without
having to check the Mesos UI to find the Mesos ID of an agent. Another
use case is when a developer sees an unhealthy AWS instance that is a
DC/OS private agent without a public IP.

### DIFF
--- a/cli/dcoscli/data/help/node.txt
+++ b/cli/dcoscli/data/help/node.txt
@@ -13,7 +13,7 @@ Usage:
     dcos node list-components [--leader --mesos-id=<mesos-id> --json]
     dcos node log [--follow --lines=N --leader --master --mesos-id=<mesos-id> --slave=<agent-id>]
                   [--component=<component-name> --filter=<filter>...]
-    dcos node ssh (--leader | --master | --mesos-id=<mesos-id> | --slave=<slave-id> | --private-ip=<private-ip>)
+    dcos node ssh (--leader | --master | --mesos-id=<mesos-id> | --private-ip=<private-ip> | --slave=<agent-id>)
                   [--config-file=<path>]
                   [--user=<user>]
                   [--master-proxy]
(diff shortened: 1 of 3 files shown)


tamarrow:
Do you have access to the private-ip directly or are you proxying via the master?

ArmandGrillet:
I am proxying via the master, the command I have tested: `dcos node ssh --proxy-ip=<master-public-ip> --private-ip=<agent-private-ip>`.

tamarrow:
Okay, in that case you can just run `dcos node ssh --master-proxy --private-ip=<private-ip>`.  Lets use <agent-ip> instead of <private-ip> here. Can you add a test please?

ArmandGrillet:
Will do. I also need to fill an issue about `--master-proxy` as it uses the private IP of the master on the AWS DC/OS cluster I use, this is why I use `--proxy-IP` instead.

cli/dcoscli/data/help/node.txt
```diff
@@ -13,7 +13,7 @@ Usage:
                   [--user=<user>]
                   [--master-proxy]
                   [--proxy-ip=<proxy-ip>]
-                  (--leader | --master | --mesos-id=<mesos-id> | --slave=<slave-id>)
+                  (--leader | --master | --mesos-id=<mesos-id> | --slave=<slave-id> | --private-ip=<private-ip>)
```
tamarrow: nit: lets put `private-ip` before `--slave`

tamarrow:
thanks @ArmandGrillet one small nit, a rebase, and then lets merge

ArmandGrillet:
Fixed and rebased, looks ready for squash and merge!

cli/tests/integrations/test_node.py
```diff
@@ -101,6 +101,11 @@ def test_node_ssh_slave():
     _node_ssh(['--mesos-id={}'.format(slave_id), '--master-proxy'])
 
 
+def test_node_ssh_slave_with_private_ip():
```
tamarrow: we need to add skip this test on windows, see https://github.com/dcos/dcos-cli/pull/910/files#diff-67456e9c321bac44727371a2aba614dfR109

ArmandGrillet:
Condition added to the test and rebase done again.

tamarrow:
thanks @ArmandGrillet

PR #912, opened 2017-02-17
use config file to determine logs source

darkonie:
DC/OS metadata endpoint will expose a config where the user is able to specify
what logging strategy to use. We should be using `dcos-log` API only if user
explicitly disables sandbox logging.
https://github.com/dcos/dcos/pull/1259

### DIFF
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -218,11 +218,25 @@ def dcos_log_enabled():
     :return: does cosmos have LOGGING capability.
     :rtype: bool
     """
+
+    # https://github.com/dcos/dcos/blob/master/gen/calc.py#L151
+    return logging_strategy() == 'journald'
+
+
+def logging_strategy():
+    """ function returns logging strategy
+
+    :return: does cosmos have LOGGING capability.
+    :rtype: str
+    """
+    # default strategy is sandbox logging.
+    strategy = 'logrotate'
+
     has_capability = packagemanager.PackageManager(
         get_cosmos_url()).has_capability('LOGGING')
 
     if not has_capability:
-        return False
+        return strategy
 
     base_url = config.get_config_val("core.dcos_url")
     url = urllib.parse.urljoin(base_url, '/dcos-metadata/ui-config.json')
@@ -237,15 +251,14 @@ def dcos_log_enabled():
     except DCOSException:
         emitter.publish('Unable to determine logging mechanism for '
                         'your cluster. Defaulting to files API.')
-        return False
+        return strategy
 
     try:
         strategy = response['uiConfiguration']['plugins']['mesos']['logging-strategy']  # noqa: ignore=F403,E501
     except KeyError:
-        return False
+        pass
 
-    # https://github.com/dcos/dcos/blob/master/gen/calc.py#L151
-    return strategy == 'journald'
+    return strategy
 
 
 def follow_logs(url):
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -6,10 +6,14 @@ import sys
 import time
 
 import six
+from six.moves import urllib
 
-from dcos import emitting, http, packagemanager, sse, util
+from dcos import config, emitting, http, packagemanager, sse, util
 from dcos.cosmos import get_cosmos_url
-from dcos.errors import DCOSException, DefaultError
+from dcos.errors import (DCOSAuthenticationException,
+                         DCOSAuthorizationException,
+                         DCOSException,
+                         DefaultError)
 
 logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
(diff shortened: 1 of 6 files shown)


cli/dcoscli/service/main.py
```diff
@@ -175,22 +172,17 @@ def _log_service(follow, lines, service, file_):
...
-    except (DCOSAuthenticationException,
-            DCOSAuthorizationException):
-        raise
-    except DCOSException as e:
-        emitter.publish(DefaultError(e))
-        emitter.publish(DefaultError('Falling back to files API...'))
 
-    task = _get_service_task(service)
-    return _log_task(task['id'], follow, lines, file_)
+    # if joruanld logging is disabled, read from files API.
```
tamarrow: spelling on journald

cli/dcoscli/task/main.py
```diff
@@ -232,28 +230,27 @@ def _log(follow, completed, lines, task, file_):
...
+
+        log.log_files(mesos_files, follow, lines)
+        return 0
+
+    # otherwise
+    if file_ in ('stdout', 'stderr'):
+        _dcos_log(follow, tasks, lines, file_, completed)
+        return 0
+
+    raise DCOSException('Only stdout/stderr files are supported by dcos-log '
```
tamarrow: when would this case happen?

cli/dcoscli/node/main.py
```diff
@@ -491,32 +489,32 @@ def _log(follow, lines, leader, slave, component, filters):
...
-            raise
-    except DCOSException as e:
-        emitter.publish(DefaultError(e))
-        emitter.publish(DefaultError('Falling back to files API...'))
 
-    if component or filters:
-        raise DCOSException('--component or --filter is not '
-                            'supported by files API')
+    # dcos-log does not support logs from leader and agent.
+    if not (leader or slave) or (leader and slave):
```
tamarrow: we only need to check the `and` case here since we checked the `or` case above

cli/dcoscli/log.py
```diff
@@ -6,10 +6,14 @@
 import time
 
 import six
+from six.moves import urllib
```
tamarrow: we no longer support python2.x so we don't need to use six if you don't want to :)

cli/dcoscli/log.py
```diff
@@ -214,9 +218,34 @@ def dcos_log_enabled():
...
+
+    if not base_url:
+        raise config.missing_config_exception(['core.dcos_url'])
+
+    try:
+        response = http.get(url).json()
+    except (DCOSAuthenticationException, DCOSAuthorizationException):
+        raise
+    except DCOSException:
+        emitter.publish('Unable to determine if dcos-log is enabled')
```
tamarrow: if the LOGGING capability exists, this endpoint should also exist, right?

Lets log the actual error and report What do you think about "Unable to determine logging mechanism for your cluster. Defaulting to X."?

cli/dcoscli/task/main.py
darkonie: this will happen if dcos-log is enabled and user passes something like 
`dcos task log test.ae5b24da-f556-11e6-af80-0e47c9122ff4 myfile`

tamarrow: How about "Invalid file [file_name]. dcos-log only supports stdout/stderr"?

darkonie:
@tamarrow tests passed, comments addressed

cli/dcoscli/node/main.py
```diff
@@ -491,32 +489,32 @@ def _log(follow, lines, leader, slave, component, filters):
...
 
     if lines is None:
         lines = 10
 
     lines = util.parse_int(lines)
 
-    try:
-        _dcos_log(follow, lines, leader, slave, component, filters)
+    # if journald logging is disabled. Read from files API and exit.
+    if not log.dcos_log_enabled():
```
malnick: We want to keep serving `component` logs via the CLI even if the journald logger for mesos is disabled. The only thing the CLI should deprecate when the journald logger for Mesos is disabled is using the journald API to get task logs - for that, it should use the files API (which this PR already provides). CC @spahl

darkonie:
@malnick @tamarrow done

tamarrow:
LGTM. @malnick ?

cli/dcoscli/node/main.py
```diff
@@ -498,7 +498,8 @@ def _log(follow, lines, leader, slave, component, filters):
     lines = util.parse_int(lines)
 
     # if journald logging is disabled. Read from files API and exit.
-    if not log.dcos_log_enabled():
+    # https://github.com/dcos/dcos/blob/master/gen/calc.py#L151
+    if 'journald' not in log.logging_strategy():
```
tamarrow: this is just a str comparison, right? lets just use `!=`

darkonie: no this could be either `journald` or `journald+logrotate` we need to run this if no journald in string found

PR #914, opened 2017-02-23
Updated 'dcos task exec' to print better error message.

klueska:
Running `dcose task exec` against containers launched by the docker
containerizer is currently unsupported. This commits prints a better
error message when attempting to run this command against such a
container.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -1015,8 +1015,21 @@ class TaskIO(object):
         client = DCOSClient()
         master = get_master(client)
 
-        # Get the URL to the agent running the task.
+        # Get the task and make sure its container was launched by the UCR.
+        # Since task's containers are launched by the UCR by default, we want
+        # to allow most tasks to pass through unchecked. The only exception is
+        # when a task has an explicit container specified and it is not of type
+        # "MESOS". Having a type of "MESOS" implies that it was launched by the
+        # UCR -- all other types imply it was not.
         task_obj = master.task(task_id)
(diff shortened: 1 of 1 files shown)


dcos/mesos.py
```diff
@@ -1015,8 +1015,16 @@ def __init__(self, task_id, cmd=None, args=None,
         client = DCOSClient()
         master = get_master(client)
 
-        # Get the URL to the agent running the task.
+        # Get the task and make sure its a container launched by the UCR.
         task_obj = master.task(task_id)
+        if "container" in task_obj.dict():
+            if "type" in task_obj.dict()["container"]:
+                if task_obj.dict()["container"]["type"] != "MESOS":
```
tamarrow: lets just use check `if task_obj.dict()["container"]["type"] != "MESOS":` and catch `ValueError`. It's a little cleaner than the nesting `ifs`

klueska: I had something similar to this at first, but the problem is I need to allow everything through except the one case where `task_obj.dict()["container"]["type"] != "MESOS":`. Meaning, I want to allow it to continue even `if "container" not in task_obj.dict()` and `if "type" in task_obj.dict()["container"]`.

I didn't see a better way of getting these semantics.

tamarrow: 👍 lets add a comment?

klueska: sounds good.

klueska:
bumped.

PR #915, opened 2017-02-24
task: add mesos agent ID to `dcos task` table

tamarrow:
This makes it easy to see a task, and ssh into the agent using either the private ip or the mesos id.

### DIFF
--- a/cli/dcoscli/tables.py
+++ b/cli/dcoscli/tables.py
@@ -40,12 +40,14 @@ def task_table(tasks):
         ("USER", lambda t: t.user()),
         ("STATE", lambda t: t["state"].split("_")[-1][0]),
         ("ID", lambda t: t["id"]),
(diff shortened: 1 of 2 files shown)


cli/dcoscli/tables.py
```diff
@@ -37,6 +37,7 @@ def task_table(tasks):
     fields = OrderedDict([
         ("NAME", lambda t: t["name"]),
         ("HOST", lambda t: t.slave()["hostname"]),
+        ("MESOS ID", lambda t: t["slave_id"]),
         ("USER", lambda t: t.user()),
         ("STATE", lambda t: t["state"].split("_")[-1][0]),
         ("ID", lambda t: t["id"]),
```
jimenez: maybe we should put this at the end, in case people have already done scripts parsing this and are cutting on column number?

tamarrow:
thanks @jimenez! updated

PR #917, opened 2017-02-24
Set exception cause for proper traceback of requests.

jeschkies:
Exceptions would be swollowed in `dcos.rpcclient.http_req`. This changes adds an
exception cause according to PEP-3134. Closes VELOCITY-1289.

### DIFF
--- a/dcos/rpcclient.py
+++ b/dcos/rpcclient.py
@@ -127,8 +127,7 @@ class RpcClient(object):
             return method_fn(url, *args, **kwargs)
         except DCOSHTTPException as e:
             text = _get_response_text(e.response)
-            logger.error('DCOS Error: %s\n%s',
-                         e.response.reason, text)
+            logger.exception('DCOS Error: %s\n%s', e.response.reason, text)
 
             try:
                 json_body = e.response.json()
--- a/dcos/rpcclient.py
+++ b/dcos/rpcclient.py
@@ -145,7 +145,8 @@ class RpcClient(object):
                 request_method=e.response.request.method,
                 request_url=e.response.request.url,
                 json_body=json_body)
-            raise DCOSException(message)
+
+            raise DCOSException(message) from e
 
 
 def _get_response_text(response):
--- a/dcos/rpcclient.py
+++ b/dcos/rpcclient.py
@@ -103,7 +103,7 @@ class RpcClient(object):
         return 'Error: {}'.format(message)
 
     def http_req(self, method_fn, path, *args, **kwargs):
-        """Make an HTTP request, and raise a DCOS-specific exception for
+        """Make an HTTP request, and raise a DC/OS-specific exception for
         HTTP error codes.
 
         :param method_fn: function to call that invokes a specific HTTP method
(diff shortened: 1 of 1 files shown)


jeschkies:
@timcharper could you approve the second commit as well :).

kensipe:
@jeschkies lgtm!

dcos/rpcclient.py
```diff
@@ -127,8 +127,7 @@ def http_req(self, method_fn, path, *args, **kwargs):
             return method_fn(url, *args, **kwargs)
         except DCOSHTTPException as e:
             text = _get_response_text(e.response)
-            logger.error('DCOS Error: %s\n%s',
-                         e.response.reason, text)
+            logger.exception('DCOS Error: %s\n%s', e.response.reason, text)
```
tamarrow: lets change to DC/OS

dcos/rpcclient.py
```diff
@@ -145,7 +144,8 @@ def http_req(self, method_fn, path, *args, **kwargs):
                 request_method=e.response.request.method,
                 request_url=e.response.request.url,
                 json_body=json_body)
-            raise DCOSException(message)
+
+            raise DCOSException(message) from e
```
tamarrow: Why did you add the `from e` here?

jeschkies: Throwing a new exception hides the original error and its stack trace. `from e` adds a `__cause__` and the stack trace. See exception chaining in [PEP-3134](https://www.python.org/dev/peps/pep-3134/).

dcos/rpcclient.py
jeschkies: Fixed.

dcos/rpcclient.py
tamarrow: the CLI should never show a stack trace to the user. The stack trace will be shown already because you switching to use `logger.exception` above.

jeschkies: The stacktrace is required for Shakedown. Right now debugging is super heard. Actually, `RpcClient.http_req` swallows more information since the actual method is passed.

tamarrow: Customer first here, we can't show a stack trace. Why can't you use the debug flag and then see the stack trace there? You are welcome to add more logging wherever else it may be helpful.

jeschkies: Hehe, I think we want the same here. I'm all for hiding stack trace unless debug mode is enabled. I'm not sure how that applies to `raise DCOSException(message) from e` though.

tamarrow: 👍 we should remove `from e` and the stacktrace will show up in debug mode from `logger.exception` above.

jeschkies: Why remove it? How do you show the `DCOSException`? I think the code that handles it could decide how to print the error.

tamarrow: I can't speak for shakedown. But the internals of the CLI (ie stacktrace) are not the right abstraction layer for your debugging. What information do you want from the CLI that you are missing?

jeschkies: Ah, so I don't use the CLI but the dcos Python client that comes with the CLI `dcos.marathon`. It's used in Shakedown and some exception are pretty useless when they swallow the actual exception. The debug line I added help here. However, I don't it's good practice to catch an exception and throw a new one. IMHO the CLI should get as much information as possible and decide how to present what to the user.

So  it pretty much comes down to shakedown using this package as a kind of Python client for DC/OS. It also adds other methods.

tamarrow:
Closing this pr since there has been no resolution or activity. We can continue discussing on the original JIRA and re-open this pr once we've agreed on a way forward.

PR #930, opened 2017-03-11
task: clean up `dcos task`

tamarrow:
Update `dcos task` only shows active tasks, `--completed` flag to only
show completed tasks and `--all` to show all tasks.

### DIFF
--- a/cli/dcoscli/task/main.py
+++ b/cli/dcoscli/task/main.py
@@ -123,13 +123,13 @@ def _cmds():
 
         cmds.Command(
             hierarchy=['task', 'log'],
-            arg_keys=['--follow', '--completed', '--lines', '<task>',
+            arg_keys=['--all', '--follow', '--completed', '--lines', '<task>',
                       '<file>'],
             function=_log),
 
         cmds.Command(
             hierarchy=['task', 'ls'],
-            arg_keys=['<task>', '<path>', '--long', '--completed'],
+            arg_keys=['<task>', '<path>', '--all', '--long', '--completed'],
             function=_ls),
 
         cmds.Command(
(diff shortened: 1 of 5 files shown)


dcos/mesos.py
```diff
@@ -25,6 +25,13 @@
 logger = util.get_logger(__name__)
 
 
+COMPLETED_TASK_STATES = [
+    "TASK_FINISHED", "TASK_KILLED", "TASK_FAILED", "TASK_LOST", "TASK_ERROR",
+    "TASK_GONE", "TASK_GONE_BY_OPERATOR", "TASK_DROPPED", "TASK_UNREACHABLE",
+    "TASK_UNKNOWN"
+]
```
tamarrow: consistency with UI: https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/constants/TaskStates.js

cli/dcoscli/task/main.py
```diff
@@ -29,7 +29,6 @@ def docopt_wrapper(usage, real_usage, **keywords):
     """ A wrapper around the real docopt parser.
     Redirects a failed command to /dev/null and prints the proper
     real_usage message, instead of just the usage string from usage.
-
```
jimenez: maybe we want to keep this?

dcos/mesos.py
```diff
@@ -567,7 +584,9 @@ def _framework_dicts(self, inactive=False, completed=False):
                 yield framework
 
         for framework in self.state()['frameworks']:
-            if inactive or framework['active']:
+            if framework['active'] and active:
```
jimenez: what are the other state possibilities here a comment would be great, also couldn't this two conditions fit in one line?

tamarrow: 'active' is a property in the framework dict, that is either True or False. `active` the parameter in the function here, is also a boolean, which means we want to include frameworks that are in the active state. 

I'm not sure what two conditions you are referring to that 'could fit in one line'.

cli/dcoscli/task/main.py
tamarrow: done

PR #922, opened 2017-03-03
dcos node: Add --field option

msabramo:
This lets the user control what fields are printed in the table and allows printing arbitary data that wasn't possible to show before, except when using `--json`.

Here's an example of referencing a custom attribute called `ethos_role`:

```
$ dcos node --field=ip --field=id --field=attributes.ethos_role
     IP                          ID                    ATTRIBUTES.ETHOS_ROLE
10.10.3.201   74df965e-a46e-46a9-8d2a-639db0af0e25-S3          public
10.10.12.29   74df965e-a46e-46a9-8d2a-639db0af0e25-S2         control
10.10.12.174  74df965e-a46e-46a9-8d2a-639db0af0e25-S0          worker
10.10.12.237  74df965e-a46e-46a9-8d2a-639db0af0e25-S1          proxy
```

By using a `:` in a field name, you can give it a custom heading in the table.

```
$ dcos node --field=ip --field=id --field=ethos_role:attributes.ethos_role
     IP                          ID                    ETHOS_ROLE
10.10.3.201   74df965e-a46e-46a9-8d2a-639db0af0e25-S3    public
10.10.12.29   74df965e-a46e-46a9-8d2a-639db0af0e25-S2   control
10.10.12.174  74df965e-a46e-46a9-8d2a-639db0af0e25-S0    worker
10.10.12.237  74df965e-a46e-46a9-8d2a-639db0af0e25-S1    proxy
```

Note that the table is always sorted by the first field specified and if it's `ip`, it sorts numerically instead of alphabetically.

One more example:

```
$ dcos node --field=disk\ used:used_resources.disk --field=endpoint:pid --field=version
DISK USED           ENDPOINT           VERSION
   0.0     slave(1)@10.10.3.201:5051    1.0.1
  4096.0   slave(1)@10.10.12.174:5051   1.0.1
  4096.0   slave(1)@10.10.12.237:5051   1.0.1
  5096.0   slave(1)@10.10.12.29:5051    1.0.1
```

Cc: @tamarrow, @matthewdfuller, @bossjones

PR #933, opened 2017-03-14
split('\n') => splitlines()

msabramo:
because `splitlines` is smarter about dealing with different line endings.

It will be interesting to see if this test still passes or if it fails on Windows, like the one in #922...

### DIFF
--- a/cli/tests/integrations/test_node.py
+++ b/cli/tests/integrations/test_node.py
@@ -42,7 +42,7 @@ def test_node_table():
 
     assert returncode == 0
     assert stderr == b''
-    assert len(stdout.decode('utf-8').split('\n')) > 2
+    assert len(stdout.decode('utf-8').splitlines()) > 2
 
 
 def test_node_log_empty():


tamarrow:
ok to test

msabramo:
Hmmm, windows build failed. Is it the test that I changed? Because then if it is, I wonder if it's possible that there are only really 2 nodes and that test was passing before because it was wrong?

tamarrow:
It is your test: 
screen shot 2017-03-15 at 8 02 19 am

I'm reluctant to think that the test was incorrectly passing before, the test cluster has 2 nodes. So the table should have 3 rows (header + row for each node). We can try using '\n' as the delimiter for splitlines. I think it's more likely that there is something going on with converting the bytes to `utf-8` and then parsing.

msabramo:
It turned out to be a problem in the Windows environment.

https://github.com/dcos/dcos-cli/pull/922#issuecomment-286912498

PR #934, opened 2017-03-15
Add `dcos config edit` command

msabramo:
which opens temporary file copy of config file with `$EDITOR` and only updates the real config file if validation succeeds.

Happy path, when the edits result in a valid config file:

```
$ dcos config edit
Executing: vim /var/folders/3b/sf0lrb1d69v8gfr8gvzxyy0h0000gn/T/tmp_rfx8308
Updated file: /Users/abramowi/.dcos/dcos.toml
```

Sad path, when user used some unknown parameter:

```
$ dcos config edit
Executing: vim /var/folders/3b/sf0lrb1d69v8gfr8gvzxyy0h0000gn/T/tmpzi8i4xpb
Error: Additional properties are not allowed ('some_unknown_param' was unexpected)
Path: core
Value: {"some_unknown_param": "blah", ...}
File /var/folders/3b/sf0lrb1d69v8gfr8gvzxyy0h0000gn/T/tmpzi8i4xpb failed validation check. Not updating /Users/abramowi/.dcos/dcos.toml.
```

I implemented `dcoscli.util.edit_file` as a (hopefully reusable) util function, because I noticed there are some places in the code where it mentioned that opening with an editor would be a useful future addition.

### DIFF
--- a/cli/dcoscli/config/main.py
+++ b/cli/dcoscli/config/main.py
@@ -3,10 +3,10 @@ import collections
 import docopt
 
 import dcoscli
-from dcos import cmds, config, emitting, http, util
+from dcos import cmds, config, emitting, http, subprocess, util
 from dcos.errors import DCOSException, DefaultError
 from dcoscli.subcommand import default_command_info, default_doc
-from dcoscli.util import decorate_docopt_usage
+from dcoscli.util import decorate_docopt_usage, edit_file
 
 emitter = emitting.FlatEmitter()
 logger = util.get_logger(__name__)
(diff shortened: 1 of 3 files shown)


tamarrow:
we have `dcos config set` for this. We update it in memory, and if validation succeeds we write to the file.

msabramo:
True. For some reason I like opening an editor, but maybe it's just me.

tamarrow:
you can always use your favorite editor directory on the config file

msabramo:
True. Ok, I'll close this.

PR #938, opened 2017-03-16
cluster: move global config to cluster config

tamarrow:
https://jira.mesosphere.com/browse/DCOS-14316

### DIFF
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -1,0 +1,50 @@
+import os
+
+from dcos import config, constants, http, util
+from dcos.errors import DCOSException
+
+logger = util.get_logger(__name__)
+
+
+def create_cluster_config():
+    """Create a cluster specific config file + directory
+    from a global config file. This will move users from global config
+    structure (~/.dcos/dcos.toml) to the cluster specific one
+    (~/.dcos/clusters/CLUSTER_ID/dcos.toml) and set that cluster as
+    the "attached" cluster.
+
+    :rtype: None
+    """
+
+    dcos_url = config.get_config_val("core.dcos_url")
+
+    # if no cluster is set, do not move the cluster yet
+    if dcos_url is None:
+        return
+
+    try:
+        # find cluster id
+        cluster_url = dcos_url.rstrip('/') + '/metadata'
+        res = http.get(cluster_url, timeout=1)
+        cluster_id = res.json().get("CLUSTER_ID")
+
+    # don't move cluster if dcos_url is not valid
+    except DCOSException as e:
+        logger.error(
+            "Error trying to find CLUSTER_ID. {}".format(e))
+        return
+
+    # create cluster id dir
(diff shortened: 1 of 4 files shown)


dcos/config.py
```diff
@@ -12,6 +12,25 @@
...
+    """Returns the path to the deprecated global DCOS config file.
+
+    :returns: path to the DCOS config file
+    :rtype: str
+    """
+
+    return os.environ.get(constants.DCOS_CONFIG_ENV, get_default_config_path())
+
+
+def get_clusters_path():
```
jimenez: Why get_clusters_**path** and not config ? be consistent with get_global_**config**?

dcos/cluster.py
```diff
@@ -0,0 +1,50 @@
...
+    structure (~/.dcos/dcos.toml) to the cluster specific one
+    (~/.dcos/clusters/CLUSTER_ID/dcos.toml) and set that cluster as
+    the "attached" cluster.
+
+    :rtype: None
+    """
+
+    dcos_url = config.get_config_val("core.dcos_url")
+
+    # if no cluster is set, do not move the cluster yet
```
jimenez: does this results in an error or at least info message? like 'config creation skipped: dcos_url unset'

dcos/cluster.py
```diff
@@ -0,0 +1,50 @@
...
+    # create cluster id dir
+    cluster_path = os.path.join(config.get_config_dir_path(),
+                                constants.DCOS_CLUSTERS_SUBDIR,
+                                cluster_id)
+
+    util.ensure_dir_exists(cluster_path)
+
+    # move config file to new location
+    global_config_path = config.get_global_config()
+    util.sh_copy(global_config_path, cluster_path)
```
jimenez: The comment says move but here we're just copying, are we leaving the global config in the old location on purpose?

dcos/config.py
```diff
@@ -12,6 +12,25 @@
...
+    """
+    :returns: path to the directory of cluster configs
```
jimenez: nit: :returns: path to the directory of cluster config files?

dcos/cluster.py
tamarrow: yes, we need to copy until we update a few more things to "cluster" mode. But after #942 we can update to move

dcos/config.py
tamarrow: updated in subsequent commit

dcos/cluster.py
tamarrow: The dcos_url is not unset by this. If the user doesn't have a dcos_url or it's invalid we skip the move to "cluster" configs. No need for an error message because this is a structure move that the user shouldn't have to know about.

dcos/config.py
tamarrow: its a config file + subcommands + potentially the attached file

PR #944, opened 2017-03-17
subcommand: install subcommands in cluster specific directory

tamarrow:
For a given cluster, read in subcommand from that cluster specific
directory + global directory of subcommands. If a subcommand with the
same name exists in both, use the subcommand in the cluster specific
directory.


tamarrow:
https://jira.mesosphere.com/browse/DCOS-14320

PR #945, opened 2017-03-20
minor clean up

darkonie:
- fix typo
- replace KeyError with a broader Exception
- remove stale comments

### DIFF
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -262,7 +262,7 @@ def logging_strategy():
 
     try:
         strategy = response['uiConfiguration']['plugins']['mesos']['logging-strategy']  # noqa: ignore=F403,E501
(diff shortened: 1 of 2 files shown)


cli/dcoscli/log.py
```diff
@@ -262,7 +262,7 @@ def logging_strategy():
 
     try:
         strategy = response['uiConfiguration']['plugins']['mesos']['logging-strategy']  # noqa: ignore=F403,E501
-    except KeyError:
+    except Exception:
```
tamarrow: what other exceptions can we get here?

darkonie:
```python
>>> a={}
>>> a['foo'] = 'bar'
>>> a['foo']['foo2']
  File "<stdin>", line 1, in <module>
TypeError: string indices must be integers
```

PR #950, opened 2017-03-24
fix cli tests

darkonie:


cli/tests/integrations/test_task.py
```diff
@@ -184,7 +184,7 @@ def test_log_two_tasks():
     assert stderr == b''
 
     lines = stdout.decode('utf-8').split('\n')
-    assert len(lines) == 23
+    assert len(lines) == 19
```
adam-mesos: This seems like a very unstable assumption, since we may add/remove lines from executor logging (or change `test-app`) in any given release. Can you instead look for some expected text, like "Subscribed executor" or "Starting task"?

darkonie: i agree with you, this is fragile indeed, however, on the other hand, this assures that we expect exactly 19 lines nothing more nothing less. If you look through the CLI code base, we use similar test all over the places.

tamarrow: While true, that's not a great reason to keep doing it :)

tamarrow:
@darkonie I'm going to merge as is, but if you get a change to update the test to be more resilient that would be appreciated :)

PR #964, opened 2017-04-02
tox.ini: Remove -x flag for unit tests

msabramo:
The `-x` makes py.test abort the test run after the first failure. While this makes the tests fail faster , it is inconvenient if there many tests that are failing, because you will only see 1 test failure and you will have to fix each failure one at a time.

### DIFF
--- a/cli/tox.ini
+++ b/cli/tox.ini
@@ -33,7 +33,7 @@ commands =
 
 [testenv:py35-integration]
 commands =
-  py.test -p no:cacheprovider -vv -x {env:CI_FLAGS:} tests/integrations{posargs}
+  py.test -p no:cacheprovider -vv {env:CI_FLAGS:} tests/integrations{posargs}
 
 [testenv:py35-unit]
 commands =


tamarrow:
This was added purposefully. High level reasoning is that the CI is supposed to be a check rather than a way to find out what tests are broken. And since the tests take so long to run, finding out there is a failure earlier is helpful. Lower level reasoning is that the tests expect a clean cluster, and (sadly) not all the tests clean up after themselves on failure. In practice, what happened in general without this is subsequent failures were due to the first failure, so it wasn't beneficial.

I know you've had trouble running the integration tests, I think time would be best spent getting that to work so you can have faster turnaround. Want to get on slack and get that to work? Or lmk what 
problems you are having and see if we can figure it out?

msabramo:
Yeah that makes sense and I figured there was probably a reason for it. 

I think it makes sense for me to invest a little time in getting the integration tests working for me locally on my machine. I don't think I tried that hard and especially not since discovering dcos-Docker. 

Very much appreciate the offer to help me on Slack. I think I'll try to get it going on my own and if I run into troubles, I might take you up on that offer.

PR #969, opened 2017-04-04
chmod 0600 tests/data/dcos.toml

msabramo:
Otherwise you get errors when running integration tests:

```
> /Users/abramowi/dev/git-repos/dcos-cli/cli/tests/integrations/test_config.py(336)_get_value()
-> assert returncode == 0
(Pdb) stderr
b"Permissions '0o644' for configuration file '/Users/abramowi/dev/git-repos/dcos-cli/cli/tests/data/dcos.toml (pointed to by tests/data/dcos.toml)' are too open.
File must only be accessible by owner. Aborting...
```

Cc: @tamarrow

### DIFF
--- a/cli/tests/integrations/test_config.py
+++ b/cli/tests/integrations/test_config.py
@@ -17,6 +17,7 @@ def env():
         constants.PATH_ENV: os.environ[constants.PATH_ENV],
         constants.DCOS_CONFIG_ENV: os.path.join("tests", "data", "dcos.toml"),
     })
+    os.chmod(r[constants.DCOS_CONFIG_ENV], 0o600)
 
     return r
 


tamarrow:
We chmod in `make test` and not in this file, because we would need to chmod for all test files :)

msabramo:
Oh I can probably move this to a fixture in `conftest.py` then and it would be used no matter what tests are run.

See http://pytest.org/2.2.4/plugins.html#conftest-py-plugins

bamarni:
We are in the process of switching tests instructions / CI setup to use the `dcos cluster setup` command (#1036), so soon enough it won't be necessary anymore to chmod that file.

PR #971, opened 2017-04-06
Add `dcos marathon task kill` command

zen-dog:
- Using `v2/tasks/delete` marathon endpoint
- Possible to kill multiple tasks at once
- Setting `scale=True` will scale down the corresponding app

### DIFF
--- a/cli/dcoscli/marathon/main.py
+++ b/cli/dcoscli/marathon/main.py
@@ -82,7 +82,7 @@ def _cmds():
 
         cmds.Command(
             hierarchy=['marathon', 'task', 'kill'],
-            arg_keys=['<task-ids>', '--scale', '--wipe'],
+            arg_keys=['<task-ids>', '--scale', '--wipe', '--json'],
             function=subcommand.task_kill),
 
         cmds.Command(
@@ -865,7 +865,7 @@ class MarathonSubcommand(object):
         emitter.publish(task)
         return 0
 
-    def task_kill(self, task_ids, scale, wipe):
+    def task_kill(self, task_ids, scale, wipe, json_):
         """Kill one or multiple Marathon tasks
 
         :param task_ids: the id of the task
@@ -874,6 +874,8 @@ class MarathonSubcommand(object):
         :type scale: bool
         :param wipe: whether remove reservations and persistent volumes.
         :type wipe: bool
+        :param json_: output JSON if true
+        :type json_: bool
         :returns: process return code
         :rtype: int
         """
@@ -881,15 +883,29 @@ class MarathonSubcommand(object):
         client = self._create_marathon_client()
         payload = client.kill_and_scale_tasks(task_ids, scale, wipe)
 
-        # No deployment was started and no tasks were killed
+        def print_deployment(deployment, json_):
+            if json_:
+                emitter.publish(deployment)
+            else:
+                emitter.publish('Created deployment: {}'.format(
+                    deployment['deploymentId']))
+
+        def print_killed_tasks(payload, json_):
+            if json_:
+                emitter.publish(payload)
+            else:
+                emitter.publish('Killed tasks: {}'.format(
+                    [task['id'] for task in payload['tasks']]))
+
         if scale:
-            emitter.publish("Started deployment: ")
-            emitter.publish(payload)
+            print_deployment(payload, json_)
         else:
-            emitter.publish('Killed tasks: ')
-            emitter.publish(payload['tasks'])
+            print_killed_tasks(payload, json_)
+
             if len(payload['tasks']) == 0:
-                return 1
+                raise DCOSException(
+                    'Failed to kill tasks. task-ids seems to be unknown')
+
         return 0
 
     def task_show(self, task_id):
--- a/cli/dcoscli/marathon/main.py
+++ b/cli/dcoscli/marathon/main.py
@@ -882,11 +882,14 @@ class MarathonSubcommand(object):
         payload = client.kill_and_scale_tasks(task_ids, scale, wipe)
 
         # No deployment was started and no tasks were killed
-        if not scale and not payload:
-            emitter.publish('No tasks were killed and no deployment was triggered.')
-            return 1
-
-        emitter.publish(payload)
+        if scale:
+            emitter.publish("Started deployment: ")
+            emitter.publish(payload)
+        else:
+            emitter.publish('Killed tasks: ')
+            emitter.publish(payload['tasks'])
+            if len(payload['tasks']) == 0:
+                return 1
         return 0
 
     def task_show(self, task_id):
--- a/cli/dcoscli/marathon/main.py
+++ b/cli/dcoscli/marathon/main.py
@@ -81,6 +81,11 @@ def _cmds():
             function=subcommand.task_stop),
 
         cmds.Command(
+            hierarchy=['marathon', 'task', 'kill'],
+            arg_keys=['<task-ids>', '--scale', '--wipe'],
+            function=subcommand.task_kill),
+
+        cmds.Command(
             hierarchy=['marathon', 'task', 'show'],
             arg_keys=['<task-id>'],
             function=subcommand.task_show),
@@ -858,6 +863,30 @@ class MarathonSubcommand(object):
             raise DCOSException("Task '{}' does not exist".format(task_id))
 
         emitter.publish(task)
+        return 0
+
+    def task_kill(self, task_ids, scale, wipe):
+        """Kill one or multiple Marathon tasks
+
+        :param task_ids: the id of the task
+        :type task_ids: [str]
+        :param scale: Scale the app down after killing the specified tasks
+        :type scale: bool
+        :param wipe: whether remove reservations and persistent volumes.
+        :type wipe: bool
+        :returns: process return code
+        :rtype: int
+        """
+
+        client = self._create_marathon_client()
+        payload = client.kill_and_scale_tasks(task_ids, scale, wipe)
+
+        # No deployment was started and no tasks were killed
+        if not scale and not payload:
+            emitter.publish('No tasks were killed and no deployment was triggered.')
+            return 1
+
+        emitter.publish(payload)
         return 0
 
     def task_show(self, task_id):
--- a/cli/tests/integrations/test_marathon.py
+++ b/cli/tests/integrations/test_marathon.py
@@ -813,7 +813,7 @@ def _stop_task(task_id, wipe=None, expect_success=True):
 
 
 def _kill_task(task_ids, scale=None, wipe=None, expect_success=True):
-    cmd = ['dcos', 'marathon', 'task', 'kill'] + task_ids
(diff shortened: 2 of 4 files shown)


kensipe:
This looks really good... assuming tests pass and the style is met.   The only thing I see missing is: 

1. test for wipe
2. test looking for the deployment id

kensipe:
great docs and +1 for initial tests

kensipe:
I'm also wondering if there should be a table... if there is than we would want --json as well.

zen-dog:
>test looking for the deployment id
That will be `test_kill_and_scale_task` here:
https://github.com/zen-dog/dcos-cli/blob/ad/marathon-task-kill-and-scale/cli/tests/integrations/test_marathon.py#L669
which will assert:
https://github.com/zen-dog/dcos-cli/blob/ad/marathon-task-kill-and-scale/cli/tests/integrations/test_marathon.py#L818

zen-dog:
>I'm also wondering if there should be a table... if there is than we would want --json as well.

I was thinking about using _tasks table_ but there tasks will be displayed with _STATUS=Healthy_ which is strange for tasks that were just killed. I wasn't sure if creating a new table is necessary.

kensipe:
@tamarrow thoughts?

cli/dcoscli/marathon/main.py
```diff
@@ -860,6 +865,32 @@ def task_stop(self, task_id, wipe):
...
+        :param scale: Scale the app down after killing the specified tasks
+        :type scale: bool
+        :param wipe: whether remove reservations and persistent volumes.
+        :type wipe: bool
+        :returns: process return code
+        :rtype: int
+        """
+
+        if not task_ids:
+            raise DCOSException('Please provide at least one task ID')
```
tamarrow: This won't happen because of the docopt definition

dcos/marathon.py
```diff
@@ -387,6 +387,40 @@ def kill_tasks(self, app_id, scale=None, host=None):
...
+            params['scale'] = scale
+        if wipe:
+            params['wipe'] = wipe
+
+        response = self._rpc.http_req(http.post,
+                                      path,
+                                      params=params,
+                                      json={'ids': task_ids})
+
+        if 'tasks' in response.json():
```
tamarrow: Why are we expecting two different response types here?

cli/tests/integrations/test_marathon.py
```diff
@@ -639,6 +639,47 @@ def test_stop_task_wipe():
...
+        task_list = _list_tasks(1, 'zero-instance-app')
+
+
+def test_kill_unknown_task():
+    with _zero_instance_app():
+        start_app('zero-instance-app')
+        watch_all_deployments()
+        task_id = ['unknown-task-id']
+
+        _kill_task(task_id, expect_success=False)
```
tamarrow: What is the expected error?

cli/dcoscli/marathon/main.py
```diff
@@ -860,6 +865,32 @@ def task_stop(self, task_id, wipe):
...
+
+        client = self._create_marathon_client()
+        payload = client.kill_and_scale_tasks(task_ids, scale, wipe)
+
+        emitter.publish(payload)
+        # No deployment was started and no tasks were killed
```
tamarrow: what failures are we expecting that return a status 200? For failures that we expect, we should return human readable strings.

cli/dcoscli/marathon/main.py
zen-dog: Removed

dcos/marathon.py
zen-dog: That's how the `v2/tasks/delete` endpoint works. With `scale=true` it starts a new deployment, otherwise it's just killing tasks.

cli/tests/integrations/test_marathon.py
zen-dog: That killing a non-existing task would return an exit code = 1

cli/dcoscli/marathon/main.py
zen-dog: My idea was: If no tasks were killed (task ID unknown) and no deployment were started (no scaling) then we return an error - basically we issued a command and nothing happened. 
I've added an explanatory message. What do you think @tamarrow ?

dcos/marathon.py
tamarrow: Shouldn't we return the full response in both cases? (ie `response.json()`)

cli/dcoscli/marathon/main.py
tamarrow: If we get to that part of the code, it means marathon returned a status 200. If there was an error, wouldn't we get a different status code? 

If the taskId is unknown, wouldn't that return a different status code and thus failure earlier in the code?

dcos/marathon.py
zen-dog: Full json just wraps the tasks into:
```
{
"tasks":[]
}
```
I can leave it as is but I didn't think it was valuable information to keep.

cli/dcoscli/marathon/main.py
zen-dog: No, we wouldn't. It's a _200_ with an empty task array e.g.:
```
{
"tasks":[]
}
```
This is the way marathon API is right now.

dcos/marathon.py
tamarrow: Looking at what we do for marathon `app kill`, we print `"Started deployment: X` for scale and `Killed tasks [X]` for kill. Would be nice to do the same here. wdyt?

cli/dcoscli/marathon/main.py
tamarrow: Is there a check we can do to see if the task ids are valid beforehand, and if they aren't tell the user? 

And then if we get to this task empty case, just report `Failed to kill tasks`. And idealy, we should add a likely cause or a way for the user to debug.

zen-dog: You could ask from *all* tasks running and compare but that would have a negative performance impact. The endpoint is built the way it would tell you afterwards whether or not the task was killed - I think we should keep it this way.

sschneid:
@mesosphere-ci retest this please

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,40 @@ def task_kill(self, task_ids, scale, wipe):
...
         """
 
         client = self._create_marathon_client()
         payload = client.kill_and_scale_tasks(task_ids, scale, wipe)
 
-        # No deployment was started and no tasks were killed
+        def print_deployment(deployment, json_):
+            if json_:
+                emitter.publish('Created deployment:')
+                emitter.publish(deployment)
```
tamarrow: if json we should just print `deployment` as is, and remove the `Created deployment` line

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,40 @@ def task_kill(self, task_ids, scale, wipe):
...
+            else:
+                emitter.publish('Created deployment: {}'.format(
+                    deployment['deploymentId']))
+
+        def print_killed_tasks(tasks, json_):
+            if json_:
+                emitter.publish('Killed tasks:')
```
tamarrow: similarly here, we should just print `tasks` and not add the 'killed tasks line'

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,40 @@ def task_kill(self, task_ids, scale, wipe):
...
+                emitter.publish(tasks)
+            else:
+                emitter.publish('Killed tasks: {}'.format(
+                    [task['id'] for task in tasks]))
```
tamarrow: 👍

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,40 @@ def task_kill(self, task_ids, scale, wipe):
...
+
         if scale:
-            emitter.publish("Started deployment: ")
-            emitter.publish(payload)
+            print_deployment(payload, json_)
         else:
-            emitter.publish('Killed tasks: ')
-            emitter.publish(payload['tasks'])
-            if len(payload['tasks']) == 0:
+            tasks = payload['tasks']
```
tamarrow: since we're returning json, let's return the response as is (ie remove this line).

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,40 @@ def task_kill(self, task_ids, scale, wipe):
...
+            print_killed_tasks(tasks, json_)
+
+            if len(tasks) == 0:
```
tamarrow: In this case (if it's not json) should we print something like "Failed to kill tasks. Please run X for more information`?

cli/dcoscli/marathon/main.py
zen-dog: Done

cli/dcoscli/marathon/main.py
zen-dog: Done

cli/dcoscli/marathon/main.py
zen-dog: Done

cli/dcoscli/marathon/main.py
zen-dog: Done

cli/dcoscli/marathon/main.py
```diff
@@ -874,22 +874,38 @@ def task_kill(self, task_ids, scale, wipe):
...
-            emitter.publish("Started deployment: ")
-            emitter.publish(payload)
+            print_deployment(payload, json_)
         else:
-            emitter.publish('Killed tasks: ')
-            emitter.publish(payload['tasks'])
+            print_killed_tasks(payload, json_)
+
             if len(payload['tasks']) == 0:
+                print('Failed to kill tasks. task-ids seems to be unknown')
```
tamarrow: Thanks for adding this! the pattern we use for errors is `raise DCOSException(....)` and then you don't need the return 1.

cli/tests/data/dcos.toml
```diff
@@ -1,5 +1,5 @@
 [core]
-dcos_url = "http://dcos.snakeoil.mesosphere.com"
 ssl_verify = "false"
-timeout = 5
 reporting = false
+dcos_url = "http://dcos.snakeoil.mesosphere.com"
+timeout = 5
```
tamarrow: we should undo change to this file

zen-dog: Ouch, that snuck in somehow. Removed

cli/dcoscli/marathon/main.py
zen-dog: We should probably be consistent then e.g.:
https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/marathon/main.py#L709
or 
https://github.com/dcos/dcos-cli/blob/master/cli/dcoscli/marathon/main.py#L679
But this is for another PR.

zen-dog:
@tamarrow , do you think we can merge it?

PR #973, opened 2017-04-07
mesos: update slave terminology to agent

tamarrow:


msabramo:
👍

tamarrow:
@ssk2 that's a pretty large change at this point, so I would punt that to a separate pr

PR #979, opened 2017-04-25
cluster: ux improvements

tamarrow:
* open DC/OS clusters don't have a cluster cert, default to insecure requests
* add --attached param to `cluster list`
* improve error messages
* use current config for http._verify_ssl

### DIFF
--- a/cli/dcoscli/data/help/cluster.txt
+++ b/cli/dcoscli/data/help/cluster.txt
@@ -6,7 +6,7 @@ Usage:
     dcos cluster --info
     dcos cluster --version
     dcos cluster attach <name>
-    dcos cluster list [--json]
+    dcos cluster list [--attached --json]
     dcos cluster remove <name>
     dcos cluster rename <name> <new_name>
     dcos cluster setup <dcos_url>
@@ -17,7 +17,7 @@ Usage:
 
 Commands:
     attach
-        Attach a configured cluster.
+        List only the currently attached cluster.
     list
         List CLI configured clusters.
     rename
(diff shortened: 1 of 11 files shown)


cli/dcoscli/data/help/cluster.txt
```diff
@@ -28,6 +28,8 @@ Commands:
         Setup the CLI to talk to your DC/OS cluster.
 
 Options:
+    --attached
+        List only attached cluster.
```
jgehrcke: "List only the currently attached cluster"?

cli/dcoscli/data/help/cluster.txt
```diff
@@ -37,7 +39,7 @@ Options:
     --insecure
         Allow requests to bypass SSL certificate verification (insecure).
     --no-check
-        Do not check CA certficate downloaded from cluster (insecure).
+        Do not check CA certficate downloaded from cluster (insecure). Enterprise only.
```
jgehrcke: "Applies to Enterprise DC/OS only."?

dcos/cluster.py
```diff
@@ -182,8 +182,13 @@ def get_cluster_cert(dcos_url):
         with urlopen(cert_bundle_url, context=unverified) as f:
             return f.read().decode('utf-8')
     except Exception as e:
-        msg = "Error downloading CA cert from cluster"
-        raise DCOSException("{}:\n{}".format(msg, e))
+        # DC/OS open clusters do not have certs
```
jgehrcke: Maybe "Open source DC/OS does not currently expose its root CA certificate"

dcos/cluster.py
```diff
@@ -182,8 +182,13 @@ def get_cluster_cert(dcos_url):
...
+        if e.code == 404:
```
jgehrcke: if the exception here does not inherit from requests' HTTPError then `e` may not have the `code` property, resulting in an AttributeError. A TCP layer problem might cause this (ConnectionRefusedError etc).

tamarrow: Good catch.

PR #982, opened 2017-04-27
Updated Alabaster license.

jedi2134:


tamarrow:
thanks @jedi2134. It actually doesn't look like we use it at all, so we should probably just remove it.

PR #983, opened 2017-05-01
Easy setup

tamarrow:
Easy setup for the CLI and cluster specific configurations (instead of a global config) to allow cluster switching.


jimenez:
LGTM

PR #984, opened 2017-05-02
http: send auth token for requests matching package.cosmos_url

tamarrow:
https://jira.mesosphere.com/browse/DCOS_OSS-1029

### DIFF
--- a/cli/tests/integrations/test_package.py
+++ b/cli/tests/integrations/test_package.py
@@ -7,7 +7,7 @@ import sys
 import pytest
 import six
 
-from dcos import constants, subcommand
+from dcos import config, constants, subcommand
 
 from .helpers.common import (assert_command, assert_lines, base64_to_dict,
                              delete_zk_node, delete_zk_nodes, exec_command,
@@ -75,7 +75,13 @@ def test_repo_list():
         ).format(**UNIVERSE_TEST_REPOS),
         'utf-8'
     )
+
     assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
+
+    # test again, but override the dcos_url with a cosmos_url config
+    dcos_url = config.get_config_val("core.dcos_url")
+    with update_config('package.cosmos_url', dcos_url):
+        assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
 
 
 def test_repo_add():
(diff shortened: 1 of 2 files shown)


cruhland:
@tamarrow were there any tests that caught this issue?

tamarrow:
No, CCM was having a really bad day.

cruhland:
@tamarrow oh, what I meant was, is there a test that will prevent a regression?

tamarrow:
added one

cli/tests/integrations/test_package.py
```diff
@@ -75,7 +75,11 @@ def test_repo_list():
         ).format(**UNIVERSE_TEST_REPOS),
         'utf-8'
     )
-    assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
+
+    # test overriding cosmos_url config
+    dcos_url = config.get_config_val("core.dcos_url")
+    with update_config('package.cosmos_url', dcos_url, env):
+        assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
```
jsancio: Can we test this as a different test? Let's keep this test the same and add a new test that set `package.cosmos_url`.

tamarrow: If it's an integration test, it will have to test something else as well. But that's why I added a comment.

jsancio: Sure. You can exec `dcos package repo list`. I just want to make sure that that command works without overriding `package.cosmos_url`.

tamarrow: ok, updated

cli/tests/integrations/test_package.py
```diff
@@ -75,8 +75,14 @@ def test_repo_list():
...
         'utf-8'
     )
+
     assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
 
+    # test again, but override the dcos_url with a cosmos_url config
+    dcos_url = config.get_config_val("core.dcos_url")
+    with update_config('package.cosmos_url', dcos_url, env):
+        assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
+
```
jsancio: @tamarrow Sorry for not being clear. Let's create another test. E.g. `test_package_cosmos_url`. Thanks!

tamarrow: this test makes it clear that if the second part of the test fails it's because of the cosmos_url. if there were to be a separate test, it would be unclear if the cosmos_url part didn't work or the output was wrong for another reason and/or the ordering of the tests would matter.

PR #991, opened 2017-05-06
DESCRIPTION.RST: Extend underline

msabramo:
I wonder if this is the reason that the description is not rendering properly on PyPI...?

https://pypi.python.org/pypi/dcos

### DIFF
--- a/DESCRIPTION.rst
+++ b/DESCRIPTION.rst
@@ -1,5 +1,5 @@
 DC/OS Command Line Interface
-===========================
+============================
 The DC/OS Command Line Interface (CLI) is a command line utility that
 provides a user-friendly yet powerful way to manage DC/OS installations.
 


tamarrow:
Apparently pypi doesn't support markdown: http://stackoverflow.com/questions/26737222/pypi-description-markdown-doesnt-work

msabramo:
Yes, that is true that PyPI doesn't support Markdown. Which I know all too well:

- https://github.com/msabramo/setuptools-markdown
- https://github.com/pypa/readme_renderer/pull/3

However, this file is reStructuredText.

```
$ pip install rstcheck
Collecting rstcheck
  Downloading rstcheck-3.1.tar.gz
Collecting docutils>=0.7 (from rstcheck)
  Using cached docutils-0.13.1-py3-none-any.whl
Building wheels for collected packages: rstcheck
  Running setup.py bdist_wheel for rstcheck ... done
  Stored in directory: /Users/abramowi/Library/Caches/pip/wheels/c1/b8/41/fe85fab5053ae89c25cd10e28f97fe68f08b36979d1c774268
Successfully built rstcheck
Installing collected packages: docutils, rstcheck
Successfully installed docutils-0.13.1 rstcheck-3.1

$ rstcheck DESCRIPTION.rst
DESCRIPTION.rst:2: (WARNING/2) Title underline too short.
```

so I think this PR will fix the issue.

PR #996, opened 2017-05-11
Add dcos-binary-install-cli.sh

karlkfi:
dcos-install-cli.sh uses pip, which isn't the recommended method of install any more. So I wrote a script to install the binary instead, using os/platform detection and DC/OS API to determine what version to download and where to download to.

I would have named it dcos-install-cli.sh, but that probably would have broken things.

Benefits of this installer:

1) Install with a single line:

    ```
    curl https://downloads.dcos.io/dcos-cli/bin/install/install-binary-dcos-cli.sh | bash -s <dcos-url>
    ```
2) Same command works on Windows, Mac, and Linux
3) Same command works for all versions of DC/OS
4) Docs for CLI install become substantially simpler and easier to maintain
5) Command in the GUI doesn't need to be updated every release
6) Users don't need to update their CLI install automation every release


karlkfi:
Test timed out... Flakey?

```
[tests.integrations.test_job.test_empty_list] tests/integrations/test_job.py:36 (test_empty_list)
```

This change is unlikely to affect anything being tested.

karlkfi:
I added some version parsing and used `dcos cluster setup` instead of `dcos config set core.dcos_url` for versions >= 0.5.0, but I'm still not sure this is a good idea.

I personally don't want the cli install script to do `dcos auth login` and it sounds like `dcos cluster setup` does that? (tho i'm not sure how to test since master is still marked 1.9-dev).

tamarrow:
I think this script should download the correct CLI for your cluster and not do anything else.

tamarrow:
I don't know what you mean by master is `1.9-dev`. ccm testing/master? The 0.5.x CLIs work for 1.8, 1.9, etc, I'm not updating the download links for older clusters because we need to update the docs as well.

karlkfi:
> I think this script should download the correct CLI for your cluster and not do anything else.

Does that mean you don't think it should do `dcos config set core.dcos_url`? 
Or you don't think it should do `dcos cluster setup`?
Or both?

tamarrow:
neither, no dcos configuration, just download the correct cli

karlkfi:
Even if we already have the cluster url? 
If we downloaded it to be compatible with a specific cluster, shouldn't it target that specific cluster?

karlkfi:
Alright, I did some refactoring and stripped out the config at the end (tho now people have to do cli version detection to decide how to configure, but that's relatively easy and json-free.

karlkfi:
Can we upload this to https://downloads.dcos.io/dcos-cli/install.sh ? 
That would be shorter than https://downloads.dcos.io/dcos-cli/bin/install/install-binary-dcos-cli.sh, which is were it would be if we put it next to `install-dcos-cli.sh`...

tamarrow:
We can't replace the old install script because it will break for all people who have old clusters. so either do an if/else or a new location.

we should also have some tests for this.

karlkfi:
> We can't replace the old install script
The old script is https://downloads.dcos.io/dcos-cli/bin/install/install-dcos-cli.sh
That's why I put "binary" in the name of the new script. But I could also just move it somewhere else and call it `install.sh`.

karlkfi:
Do we have tests for `install-dcos-cli.sh`? I don't mind adding CI tests, but it doesn't look like you're using Jenkinsfiles here, which makes it a bit more complicated for me to set up.

Do we need one for each OS (mac/linux/windows)?

karlkfi:
:( looks like I was wrong... `dcos-cli/bin/install/install-dcos-cli.sh` is being installed to https://downloads.dcos.io/dcos-cli/install.sh already.

karlkfi:
So... https://downloads.dcos.io/dcos-cli/install-binary.sh ?

tamarrow:
ok. we have no tests, but if we add this, we should add tests. we have a CI that runs on each platform, so the test should pass on all of them. Also can you post some example output please?

karlkfi:
```
$ DCOS_VERSION=1.10.0 bin/install/install-binary-dcos-cli.sh http://m1.dcos/
DC/OS URL: http://m1.dcos/
DC/OS Version: 1.10.0
Install Path: /usr/local/bin
Download URL: https://downloads.dcos.io/binaries/cli/darwin/x86-64/dcos-1.10/dcos
Download Path: /var/folders/k1/2wh3ny9s4g5_1q9q4bw9jj2c0000gn/T/dcos-install-cli.yU4p88w2Fv2E/dcos
Downloading...
Installing...
CLI Version: 0.5.1
Executable Path:
/usr/local/bin/dcos
```

Everything's STDERR except the last line which is STDOUT so you can save it to a variable.

karlkfi:
I have no idea if these Jenkinsfiles will work, but I had to commit them to test them...

karlkfi:
So i spent a day learning how to use the new Jenkins Declarative Pipeline DSL and fixing the CCM scripts and now I can't find any way to trigger the pipeline build on GitHub PRs...

You can add the ghprb-plugin config manually to the pipeline in Jenkins, but then when it triggers it doesn't find the ghprb-plugin config in the Jenkinsfile, so it removes it from the build config...

So then I tried to make another build manually and use a MultiJob phase to trigger the previously mentioned pipeline job, but the trigger doesn't seem to work...

https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/scripts-test/
https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/scripts-test-trigger/

At this point, I've wasted too much time on this and need to get back to other things...

karlkfi:
Turns out MultiJob builds can't trigger Pipeline builds: https://issues.jenkins-ci.org/browse/JENKINS-38825 :man_facepalming:

karlkfi:
Alright, I made a new MultiJob Jenkins job: 
https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/test-install-script-0-prs/

Right now it tests the CLI install script on linux and mac targeting a testing/master CCM cluster.

I still need to add windows, but I'm not sure how to execute bash shell scripts on windows in jenkins... Tho I know the azure boxes have win-bash on them... somewhere...

karlkfi:
Rebased to pull in #998 test fix

karlkfi:
Still failing:
```
tests.integrations.test_experimental.test_service_start_happy_path
```

Is that because we're testing against master, @tamarrow?

karlkfi:
To test, run https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/test-install-script-0-prs/ with 
```
ghprbActualCommit : refs/pull/996/head
```

karlkfi:
New tests pass on latest push : https://jenkins.mesosphere.com/service/jenkins/job/DCOS-CLI/job/test-install-script-0-prs/15/

karlkfi:
I've rebased this, removed my custom jenkinsfiles and added linux/mac/windows installer tests.

karlkfi:
Rebased again

karlkfi:
uh... what happened to the CI tests?

karlkfi:
I dont think this is related to my changes. Looks like a flakey bug in DC/OS itself or maybe DC/OS not having automatic retry.

```
STDERR: 
CMD: ['dcos', 'experimental', 'package', 'add', '--dcos-package', 'C:\\Users\\SSHD\\AppData\\Local\\Temp\\tmpxgt5gqkb\\helloworld2-0.1.0-d7847c60d148329d4422f9c7781111d1.dcos']
STDOUT: 
STDERR: Authentication failed. Please run `dcos auth login`
```

karlkfi:
Looks like master CI is broken.

karlkfi:
rebased again

karlkfi:
Only the binary tests ran here...

karlkfi:
run linux integration tests

karlkfi:
run linux integration tests

karlkfi:
woot! linux test worked:
```
[linux-tests] [/workspace] Running shell script
[linux-tests] + rm -rf /root/.dcos
[linux-tests] + grep -q ^.* dcos.snakeoil.mesosphere.com$ /etc/hosts
[linux-tests] + echo 34.214.79.241 dcos.snakeoil.mesosphere.com
[Pipeline] [linux-tests] sh
[linux-tests] [/workspace] Running shell script
[linux-tests] + bash dcos-cli/tests/test-install.sh 34.214.79.241
[linux-tests] DC/OS URL: 34.214.79.241
[linux-tests] DC/OS Version: 1.10.0-beta2
[linux-tests] Install Path: /usr/local/bin
[linux-tests] Download URL: https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.10/dcos
[linux-tests] Download Path: /tmp/dcos-install-cli.6lU7vd01iRIs/dcos
[linux-tests] Downloading...
[linux-tests] Installing...
[linux-tests] CLI Version: 0.5.4
[linux-tests] Executable Path:
[linux-tests] Install Path: /usr/local/bin/dcos
[Pipeline] [linux-tests] dir
[linux-tests] Running in /workspace/dcos-cli/cli
```

karlkfi:
run windows integration tests

karlkfi:
run mac integration tests

karlkfi:
run windows integration tests
(cluster creation failed)

karlkfi:
run windows integration tests

karlkfi:
both mac & windows failed due to CREATE_FAILED

karlkfi:
run mac integration tests

karlkfi:
run windows integration tests

karlkfi:
windows failure:
```
[windows-tests] jenkins@JENKINS-BASE6 C:\windows\workspace>bash dcos-cli/tests/test-install.sh "${DCOS_URL}" 
[windows-tests] set: pipefail: unknown option name
```

mac failure:
```
[mac-tests] tests/integrations/test_marathon_debug.py::test_debug_details FAILED
```

karlkfi:
run mac integration tests

karlkfi:
Windows test is failing because win-bash does not support `set -o pipefail`. win-bash also does not support `$!` and isn't being supported any more... so maybe we shouldn't test with it...

jsancio:
What is our plan with this PR? cc @klueska @bamarni

bamarni:
I didn't know we already had a script for this, with the plan of doing https://jira.mesosphere.com/browse/DCOS_OSS-1756, users won't have to make sure they install the correct cli version for their cluster anymore.

This means dcos-cli would assume the role of installer and is to be installed only once, so I think we shouldn't add on top of it shell install scripts to be used as piping curl into bash. Rather I'd like to document how to download the CLI, verify the checksum, and move it to the PATH.

We could also move towards supporting OS pacakge managers, for example the project is popular enough for having an official homebrew package (https://github.com/Homebrew/homebrew-core/pull/13378), which is not maintained by us at the moment.

karlkfi:
The script does three things:
- Install the right cli version for your cluster
- Install the right cli binary for your environment
- Uses defaults to decide where to install the CLI

This allows the GUI and Docs to list exactly one command to install the CLI:
- The command doesn't change between DC/OS versions (so it can be hard coded in automation)
- The command is the same for mac, windows, and linux (so the GUI and CLI can be simplified UX and easier to maintain)

Even if you have a binary that updates itself to the latest or even to the latest per cluster, that still only solves 1, the user still has to deal with 2, 3, 4, and 5.

The reason this is still pending is because 2 is actually hard to solve, because bash on windows is crippled and requires an even more complicated install process than this script. So we might need two install scripts (mac/linux sh vs windows bat). I dont have experience writing batch scripts. So that's why this PR is hung.

bamarni:
My concern is not that much about the usefulness of this script, it'd indeed allow the same single command to be run on most setups (even though I'm not sure curl and bash are that safe to assume, especially on windows). But it's more about making it the recommended installation way :
- this script requires to have a cluster running in order to install the CLI, while this is currently the case with the GUI recommended method, we would be able to move away from this with DCOS_OSS-1756
- piping curl into bash is very sensible, there are various flags that need to be passed to curl to make it safer to be piped, this is also not optimal in terms of security, reproducibility, especially in the context of automation. 

So I'd be ok with having this as a simple and quick easy way to install the CLI, but not making it the recommended way. This would be similar to https://get.docker.com/ for example.

bamarni:
closing because of the concerns mentioned above, the cli will also not be tied to the dc/os version anymore in 1.12.

PR #1001, opened 2017-06-06
Update cosmos client to use v3 describe response

jsancio:

PR #1004, opened 2017-06-17
Fixed broken integration tests.

klueska:
A number of marathon tests were failing on windows and blocking PRs from being merged. After looking at the tests and fixing them, it's not clear how they were ever working (even on linux/mac). I've updated them to be more stable both on linux/mac and on windows.

After fixing these issues, I noticed a bunch of other integration tests were failing as well. This PR is attempting to address all of these outstanding failing tests and turn them all green again.

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -285,7 +285,7 @@ def _describe(package_name,
             if render:
                 app_output = pkg.marathon_json(user_options)
             else:
-                app_output = pkg.marathon_template()
+                app_output = pkg.marathon_template().decode("UTF-8")
                 if app_output and app_output[-1] == '\n':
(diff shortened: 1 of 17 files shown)


jsancio:
@klueska I updated #1001. There was a small bug in that PR. You may want to update this PR to include those changes.

klueska:
Cool. Thanks Still working through some remaining failures (not just on windows anymore and mostly related to `test_package.py`).

cruhland:
@klueska I'll review once the tests are in a good state. Let me know if those package tests are giving you trouble.

klueska:
The one failing test on `continuous-integration/teamcity` is flaky, and the windows failure this time around is a node-provisioning failure. I will attempt to address both of these in subsequent PRs.

PR #1006, opened 2017-06-22
Migrate CI jobs to Jenkins with a Jenkinsfile.

klueska:
This PR provides a full rewrite of the CLI build system, including the addition of a source controlled Jenkinsfile for automated builds of a single Jenkins pipeline. The rewrite of the build system was designed to make the build scripts themselves more cross platform so we could run (essentially) the same commands across linux/mac/windows instead of writing highly customized build scripts for each platform into the CI. The move towards a Jenkinsfile was to (1) consolidate our CI into a single system (previously it was split across both Jenkins and TeamCity), and (2) put our CI scripting under version control instead of only tracking it the CI system itself.

Currently the Jenkinsfile is designed to accommodate a single job to run both binary builds and integration tests across all platforms. Ideally these would be split into separate Jenkins jobs, but the Multibranch Pipeline Plugin we currently use on Jenkins doesn't make splitting these jobs out like this easy (it hardcodes the location it looks for the Jenkinsfile to be at the root of the repository and you can't change it). In future revisions we should consider splitting the Jenkinsfile apart (and possibly pulling it into a library for other jobs to use). Once that is possible we can have separate Jenkinsfiles for each logical build and have them display separately in the checks field at the bottom of the PR.


Jenkinsfile
```diff
@@ -0,0 +1,347 @@
...
+    String platform
+    int createAttempts
+
+    TestCluster(WorkflowScript script, String platform) {
+        this.script = script
+        this.platform = platform
+        this.createAttempts = 0
+    }
+
+    def _create() {
```
karlkfi: i dont understand the naming patterns here. Why are we duplicating names and using underscores inconsistently?

Makefile
```diff
@@ -1,16 +1,34 @@
...
 clean:
-	bin/clean.sh
+	@bash bin/clean.sh
+
+distclean: clean
+	@bash bin/distclean.sh
 
 env:
-	bin/env.sh
+	@bash bin/env.sh
```
karlkfi: why do these all need to be prefixed with bash? isn't having them all use a hashbang and be executable enough?

bin/docker.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+CURRDIR=$(dirname "${0}")
+source ${CURRDIR}/common.sh
+
+: ${DOCKER_RUN:="docker run \
```
karlkfi: what does this `:=` syntax do in bash variable evaluation?

bin/docker.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
```
karlkfi: I would recommend using `set -o errexit -o nounset -o pipefail` in all bashfiles, not just `-e`.

cli/requirements.txt
```diff
@@ -3,4 +3,5 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
+PyInstaller==3.2.1
```
karlkfi: new dependency?

Jenkinsfile
```diff
@@ -0,0 +1,347 @@
...
+            > ${this.platform}_dcos_url"""
+
+        return script.readFile("${this.platform}_dcos_url")
+    }
+
+    def acsToken() {
+        def dcosUrl = this.dcosUrl()
+
+        script.sh """
+            python -c \
```
karlkfi: what's happening here? can it be in it's own script file and not inline? I'd rather not have inline python in the tests.

Jenkinsfile
```diff
@@ -0,0 +1,347 @@
...
+                }
+                create()
+            }
+        }
+    }
+
+    def dcosUrl() {
+        script.sh """
+            ./dcos-launch describe -i ${this.platform}_cluster_info.json \
+            | python -c \
```
karlkfi: can we extract this inline python?

Jenkinsfile
karlkfi: these functions all need comments to describe them too

klueska: I wanted to have simple wrappers around just the sh commands instead of embedding them in more complicated logic for the "real" variants of those commands. The exception is delete, which didn't need a wrapper, so didn't have an '_' version. The '_' itself was meant to imply that they shouldn't use used outside the class. I could probably use the private keyword for that though.

Makefile
klueska: Not on windows. This is necessary for using the Makefile on windows.

bin/docker.sh
klueska: If it's not set, set it. This way I could override this variable from a script that wraps this one (as I actually do in the `cli/bin/docker.sh` script.

bin/docker.sh
klueska: Makes sense.

cli/requirements.txt
klueska: It's not a "new" dependency so much as it is a consolidation of installing this dependency as part of the standard `make env` step. Previously, the scripts in the CI did:
```
echo "Building binary with pyinstaller"
cd cli
pip install -r requirements.txt
pip install -e .
pip install pyinstaller==3.1.1
pyinstaller binary/binary.spec
```

Which seems completely unnecessary to me. I guess the idea was that if you are only developing and don't need to build the binary, then you don't need this. I think it makes more sense to include it always though.

Jenkinsfile
klueska: Meaning a script file in the repo that we just call out to? I was hoping to avoid having individual script files like this just for CI, but I could be convinced otherwise.

Jenkinsfile
klueska: Ditto.

Makefile
```diff
@@ -1,16 +1,34 @@
-all: env test packages
+all: env packages doc test
+all-docker: env-docker packages-docker doc-docker test-docker
 
 clean:
-	bin/clean.sh
+	@bash bin/clean.sh
+
+distclean: clean
```
adamtheturtle: Not new in this PR but these targets appear to be unrelated to the names of files that they create and as such they should be [`PHONY`](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) targets.

Jenkinsfile
```diff
@@ -0,0 +1,386 @@
...
+                         variable: 'CLI_TEST_SSH_KEY_PATH']]) {
+
+                        withEnv(["DCOS_URL=${dcosUrl}",
+                                 "DCOS_ACS_TOKEN=${acsToken}"]) {
+                            try {
+                                body()
+                            } catch (Exception e) {
+								println "Build failed. The DC/OS cluster at" +
+										" ${dcosUrl} will remain temporarily" +
+										" active so you can debug what went" +
```
jsancio: The formatting looks strange. Do you have tabs?

Jenkinsfile
```diff
@@ -0,0 +1,386 @@
...
+        return script.readFile("${this.platform}_dcos_url")
+    }
+
+    // Retreives the ACS Token of a cluster previously created using `create()`.
+    def getAcsToken() {
+        def dcosUrl = this.getDcosUrl()
+
+		// In the future, consider doing the following with curl / jq instead
+		// of inline python (however, jq is not installed on our windows
+		// machines at the moment).
```
jsancio: Strange formatting. Do you have tabs?

Jenkinsfile
```diff
@@ -0,0 +1,386 @@
...
+// This Jenkinsfile runs a set of parallel builders for the dcos-cli across
+// multiple platforms (linux/mac/windows). One set of builders builds the CLI
+// into a binary on each platform. The other set of builders runs integration
+// tests on each platform. Under the hood, the integration test builders use
+// `dcos_launch` to create the DC/OS clusters for each platform to run their
+// tests against. Unfortunately, `dcos_luanch` only works reliably on Linux,
+// so we use a single linux instance to create all of the clusters and separate
+// linux/mac/windows instances to run the actual tests.
+
+// These are the platforms we are building against.
```
jsancio: Let's use multiline comments for multiline comments.

klueska:
Updated based on Jose's comments.

PR #1007, opened 2017-06-29
Add disclaimer during install

larioj:
https://jira.mesosphere.com/browse/DCOS-16434

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -352,7 +352,7 @@ def _install(package_name, package_version, options_path, app_id, cli,
         emitter.publish(pre_install_notes)
 
     if yes:
-        emitter.publish('Continue installing? [yes/no] yes')
+        emitter.publish('Continue installing? [yes/no] ')
     if not confirm('Continue installing?', yes):
         emitter.publish('Exiting installation.')
         return 0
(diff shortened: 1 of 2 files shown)


cli/dcoscli/package/main.py
```diff
@@ -334,6 +334,19 @@ def _install(package_name, package_version, options_path, app_id, cli,
...
+        link = ('https://mesosphere.com/'
+                'catalog-terms-conditions/#certified-services')
+    else:
+        link = ('https://mesosphere.com/'
+                'catalog-terms-conditions/#community-services')
+    emitter.publish(
+        ('By Deploying, you agree to '
+         'the Terms and Conditions ' + link + '\n')
+    )
+
```
jsancio: We should also always prompt the user and print "Continue installing?".

Line 353 needs to execute irrespective of line 351.

cli/dcoscli/package/main.py
```diff
@@ -334,12 +334,28 @@ def _install(package_name, package_version, options_path, app_id, cli,
...
     pre_install_notes = pkg_json.get('preInstallNotes')
     if app and pre_install_notes:
         emitter.publish(pre_install_notes)
-        if not confirm('Continue installing?', yes):
-            emitter.publish('Exiting installation.')
-            return 0
+
+    if yes:
+        emitter.publish('Continue installing? [yes/no] ')
```
jsancio: I think I wasn't 100% accurate in my previous comment. I forgot about `--yes`. We should preserve the old behavior and not print this if the user passes `--yes`. In other words I think this should just be:
```python
if not confirm('Continue installing?', yes):
  emitter.publish('Exiting installation.')
  return 0
```

PR #1008, opened 2017-06-29
Add content negotiation to packaging cli

larioj:


dcos/cosmos.py
```diff
@@ -180,7 +180,8 @@ def _cosmos_request(self,
                         response.headers.get('Content-Type')))
             return response
         except DCOSBadRequest as e:
-            if len(headers_preference) > 1:
+            error_type = e.response.json().get('type')
+            if error_type == 'not_valid' and len(headers_preference) > 1:
```
jsancio: @larioj I think we should fix this in Cosmos first. I want our error types to be consistent. This is not consistent with the rest of our system.

jsancio: This type of errors are constructed here: https://github.com/dcos/cosmos/blob/master/cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala#L37-L52

larioj: @jsancio How do you propose we make them consistent? Should we just change the names? Should we create CosmosErrors for each case and convert to a CosmosError? Does this not have the potential to break a lot of clients?

larioj: Discussed offline. Will send all accept types at once (content negotiation).

dcos/cosmos.py
```diff
@@ -163,33 +159,20 @@ def _cosmos_request(self,
...
-                                            json,
-                                            **kwargs)
-            else:
-                raise e
+        if http_request_type is 'post':
+            response = http.post(
+                url, data=data, json=json, headers=headers, **kwargs)
+        else:
+            response = http.get(
+                url, data=data, json=json, headers=headers, **kwargs)
```
jsancio: HTTP GET request are not allowed to have a body. You can remove `data=data` and `json=json`.

dcos/cosmos.py
```diff
@@ -227,7 +210,7 @@ def _get_http_method(self, endpoint):
         """
         return self._endpoint_data.get(endpoint).get('http_method')
 
-    def _get_header(self, endpoint, version, headers=None):
+    def _get_header(self, endpoint, versions, headers=None):
```
jsancio: Should we just change this to `_get_header(self, endpoint, headers=None)`? Don't we always want to send all of the version that we support?

larioj:
This is WIP

PR #1009, opened 2017-07-07
Update all `test_ls*` tests to accommodate new `.ssl` folder in sandbox.

klueska:
The newest `dcos` master includes a `.ssl` folder that needs to be accounted for when running all of the tests that look for (and account for) files in a task's sandbox. This PR updates these tests appropriately.

### DIFF
--- a/cli/tests/integrations/test_task.py
+++ b/cli/tests/integrations/test_task.py
@@ -271,9 +271,13 @@ def test_ls():
 
 
 def test_ls_multiple_tasks():
-    ls_line = 'stderr.*stdout'
     returncode, stdout, stderr = exec_command(
         ['dcos', 'task', 'ls', 'test-app'])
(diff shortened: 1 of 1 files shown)


klueska:
There is a new folder in the sandbox called `.ssl` so we need to include this in the regex search since the directory doesn't just include `stderr*` and `stdout*` files anymore. Another "fix" would have been to change the regex to `.*stderr.*stdout.*` to simply ensure that at least `stderr` and `stdout` files were included in the listing (but allow others as well). I figured it was better to explicitly include the `.ssl` folder though. Ideally we wouldn't just us a simple regex here, but actually split the line by spaces and search for the exact files. I'll leave that to a future commit though.

klueska:
This passes in CI with #1006 rebased on top of it.

PR #1015, opened 2017-07-14
adding --all flag to cluster remove

kensipe:


karlkfi:
are there tests that need updating?

kensipe:
I prefer: https://github.com/dcos/dcos-cli/pull/1016
over this... but we could have both.

kensipe:
I will add integration test Monday... there is some recent (in the last several months) which is stopping my local integration tests from passing.  I'm setting chmod +x on the test toml... there must be something else... any help would be appreciated.  it looks like the docs may be out of alignment.
any thoughts @jsancio ?

cli/dcoscli/cluster/main.py
```diff
@@ -123,14 +123,19 @@ def _list(json_, attached):
...
+    :type all: bool
     :rtype: None
     """
-
-    return cluster.remove(name)
+    if all:
+        for c in cluster.get_clusters():
+            cluster.remove(c.get_name())
+    else:
+        return cluster.remove(name)
```
adamtheturtle: No need to return this, unless `cluster.remove` has a return type and then the `:rtype:` documentation here needs changing.

kensipe: the original impl was to return... I left it as I found it... but I see no value in it.  I'll remove.

cli/dcoscli/cluster/main.py
```diff
@@ -123,14 +123,19 @@ def _list(json_, attached):
...
     """
     :param name: name of cluster
     :type name: str
+    :param all: remove all clusters if True
+    :type all: bool
     :rtype: None
     """
-
-    return cluster.remove(name)
+    if all:
```
adamtheturtle: `all` is a Python built-in function and so it should not be used as a variable or parameter name. 

❓Also, I find it weird to have a function take `name` and `all` - maybe `Name` can be made to be an `Optional[str]` and if `all` is `True` and `name` is not `None` then an error is raised.

kensipe:
integration tests coming today

kensipe:
The CI integration tests are NOT setup like the developer test instructions... there is no `DCOS_DIR` setup so these tests fail.   Analyzing the jenkins file and testing... it does not appear that the cluster directory structure (under the DCOS_URL) is ever tested in the CI env tests.    It appears that the `dcos.snakeoil.mesosphere.com` is associated with the DCOS_URL in the /etc/hosts file and that all tests are configured to use the `tests/data/dcos.toml`

kensipe:
@klueska code complete and tests are good.   the test failure is due to some unrelated flaky windows test failure.

kensipe:
rebased.

kensipe:
rebased

kensipe:
rebased

kensipe:
rebased

kensipe:
run linux integration tests

cli/tests/integrations/test_cluster.py
```diff
@@ -29,6 +34,58 @@ def test_list():
...
+    assert dcos_dir is not None
+    cluster_dir = os.path.join(dcos_dir, 'clusters')
+    back_dir = os.path.join(dcos_dir, 'backup')
+    copy_tree(cluster_dir, back_dir)
+
+    yield cluster_dir
+    # return to order
+    remove_tree(cluster_dir)
+    copy_tree(back_dir, cluster_dir)
+    remove_tree(back_dir)
```
bamarni: Even though you seem to clean it up properly, I wouldn't use the actual dcos dir as a fixture, this can leave subtle inconsistencies or become a pain to debug if something goes wrong (one would have to remove the folder, setup the cluster again to start from a clean state).

Could you instead copy the whole dcos dir to a temporary location, make your changes and then set it as `DCOS_DIR`, there is something similar here (except that it doesn't copy the current one) : https://github.com/dcos/dcos-cli/blob/ce504929eadbb60e3befeeb03a65566d535482bb/cli/tests/integrations/test_cluster_migration.py#L18-L26

kensipe: this is better!

kensipe:
@bamarni thanks for the review... much better approach.  this is updated and rebased.

kensipe:
run linux integration tests

bamarni:
@kensipe : the test failure is related to your changes, it is trying to copy the directory referenced in `DCOS_DIR`, however that env variable is not set on our integration tests. The example code snippet I've pasted just makes sure it is reverted back to its original value in case it exists, you should still use `config.get_config_dir_path()` to get the actual config path.

kensipe:
yeah... previous conversations with other engs lead me to believe that DCOS_DIR is always set... fix coming.

bamarni:
run linux integration tests

bamarni:
a rebase should fix the failing tests, we recently moved the test suite to use dcos-launch (#1056) and I think it's trying to SSH as the wrong user, should be `centos` now.

kensipe:
rebased

bamarni:
run linux integration tests

bamarni:
`test_debug_details` failed, there is already a ticket about this flaky test, re-running.

bamarni:
run linux integration tests

kensipe:
cool thx!

bamarni:
Thanks @kensipe

PR #1016, opened 2017-07-14
cluster list and remove updates to be composable

kensipe:
* list clusters ids `dcos cluster list --quiet` or `dcos cluster list -q`
* remove a list of ids
* composable `dcos cluster remove $(dcos cluster list --quiet)`
* better documentation

### DIFF
--- a/cli/dcoscli/data/help/cluster.txt
+++ b/cli/dcoscli/data/help/cluster.txt
@@ -23,7 +23,7 @@ Commands:
     rename
         Rename a cluster name in the CLI.
     remove
-        Remove a list of configured clusters from the CLI.
+        Remove configured cluster(s) from the CLI.
     setup
         Setup the CLI to talk to your DC/OS cluster.
 
--- a/cli/dcoscli/data/help/cluster.txt
+++ b/cli/dcoscli/data/help/cluster.txt
@@ -51,7 +51,7 @@ Options:
     --private-key=<key_path>
         Specify the path to a file that contains the private key.
     -q, --quiet
-        Only display cluster IDs
+        Only display cluster names.
     --username=<username>
         Specify the username for login.
     --version
--- a/cli/dcoscli/data/help/cluster.txt
+++ b/cli/dcoscli/data/help/cluster.txt
@@ -23,7 +23,7 @@ Commands:
     rename
         Rename a cluster name in the CLI.
     remove
-        Remove a configured cluster from the CLI.
+        Remove a list of configured clusters from the CLI.
     setup
         Setup the CLI to talk to your DC/OS cluster.
 
--- a/cli/dcoscli/data/help/cluster.txt
+++ b/cli/dcoscli/data/help/cluster.txt
@@ -6,8 +6,8 @@ Usage:
     dcos cluster --info
     dcos cluster --version
     dcos cluster attach <name>
-    dcos cluster list [--attached --json]
-    dcos cluster remove <name>
+    dcos cluster list [--attached --json --quiet]
+    dcos cluster remove <names>...
     dcos cluster rename <name> <new_name>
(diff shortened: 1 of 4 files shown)


kensipe:
The CI integration tests are NOT setup like the developer test instructions... there is no `DCOS_DIR` setup so these tests fail.   Analyzing the jenkins file and testing... it does not appear that the cluster directory structure (under the DCOS_URL) is ever tested in the CI env tests.    It appears that the `dcos.snakeoil.mesosphere.com` is associated with the DCOS_URL in the /etc/hosts file and that all tests are configured to use the `tests/data/dcos.toml`

kensipe:
@klueska all good here with integration tests

bamarni:
https://github.com/kensipe/dcos-cli/blob/2a6419c78fcd1314d9feae2bb1e7d96f598da53a/cli/dcoscli/data/help/cluster.txt#L26

Could we make it clear here that the remove command now accepts multiple clusters? eg. `Remove configured cluster(s) from the CLI.`

kensipe:
rebased and updated the help text as requested.

cli/tests/integrations/test_cluster.py
```diff
@@ -27,6 +31,67 @@ def test_list():
...
+
+    assert num_of_clusters() == 0
+
+
+def num_of_clusters():
+    _, stdout, _ = exec_command(
+        ['dcos', 'cluster', 'list', '--json'])
+    cluster_list = json.loads(stdout.decode('utf-8'))
+    return len(cluster_list)
+
```
klueska: Isn't this already done elsewhere? I'm sure there are other tests that rely on this. We should think about putting this in a helper library instead of duplicating it all over the place.

dcos/cluster.py
```diff
@@ -215,9 +215,9 @@ def get_clusters():
 
 def get_cluster(name):
     """
-    :param name: name of cluster
+    :param name: name or cluster_id of cluster
     :type name: str
-    :returns: Cluster identified by name
+    :returns: Cluster identified by name or cluster_id
```
klueska: These changes to appear to logically below to this PR. I'm hesitant to squash this into the commit since it is unrelated.

kensipe:
rebased... I will address nits later today

kensipe:
rebased

cli/dcoscli/data/help/cluster.txt
```diff
@@ -50,6 +50,8 @@ Options:
         Specify the authentication provider to use for login.
     --private-key=<key_path>
         Specify the path to a file that contains the private key.
+    -q, --quiet
+        Only display cluster IDs
```
elezar: Here we refer to `IDs`, but it seems as if we refer to them as `names` elsewhere.

kensipe:
rebased and addressed @elezar feedback

kensipe:
run linux integration tests

kensipe:
@bamarni ready to roll!

bamarni:
@kensipe : as we already have #1015 in the pipeline, I'd rather leave this one on hold. It's less user-friendly and we don't have any other use-case for composition including the cluster list right now.

This feature can be great when one is also able to filter using criterias (similarly to `docker ps` used with `docker rm` or `docker stop`). So we might revisit it in the future.

kensipe:
@bamarni @klueska I think this is a mistake... I know there were 2 ways presented to potentially remove all clusters... however this way is the "best" way IMO and adds more... it adds the ability to make things composable.   I would want this even with #1015 added to the code base.

bamarni:
@kensipe : could you please elaborate on why it is better? I'm not closing the door forever here, just saying that I don't see any use-case for this at the moment (as we are going to add the `--all` flag to `dcos cluster remove`).

PR #1019, opened 2017-07-22
Updated Jenkinsfile to test against 'master', 'stable' and 'legacy' clusters.

klueska:


Jenkinsfile
```diff
@@ -283,110 +300,128 @@ def testBuilder(String platform, String nodeId, String workspace = null) {
...
+                    unset DCOS_ACS_TOKEN; \
+                    make test-binary"'''
+            }
         }
-    }
-})
+    })
+}
+
+def testBuilders = [:]
```
jsancio: I am not familiar with this language but I'll assume that `def` without parameters is a value.

Jenkinsfile
```diff
@@ -423,14 +458,26 @@ node('py35') {
...
         [$class: 'UsernamePasswordMultiBinding',
          credentialsId: '323df884-742b-4099-b8b7-d764e5eb9674',
          usernameVariable: 'DCOS_ADMIN_USERNAME',
          passwordVariable: 'DCOS_ADMIN_PASSWORD']]) {
 
-        parallel builders
+        parallel binaryBuilders
+
+        def exceptions = []
+        for (version in DCOS_VERSIONS) {
```
jsancio: Can we reverse the order of execution. I think we should run against master first. Or even better run them all in parallel. Is that possible?

Jenkinsfile
```diff
@@ -14,9 +14,27 @@
...
+      credentialsId: '542aa287-9464-49cb-bdd3-9e4ca457dd87',
+      variable: 'CF_TEMPLATE_URL_1_9_2'],
+     [$class: 'StringBinding',
+      credentialsId: 'fd1fe0ae-113d-4096-87b2-15aa9606bb4e',
+      variable: 'CF_TEMPLATE_URL_MASTER']]) {
+
+    TEMPLATE_URLS['1.8.9'] = "${env.CF_TEMPLATE_URL_1_8_9}"
+    TEMPLATE_URLS['1.9.2'] = "${env.CF_TEMPLATE_URL_1_9_2}"
+    TEMPLATE_URLS['master'] = "${env.CF_TEMPLATE_URL_MASTER}"
+}
```
jsancio: How are you planning to manage all of these magical strings and URLs? Do we need to update these for every minor release? What happens when we do a major release?

klueska: I've now renamed these to 'master', 'stable', and 'legacy'. For now these are pulling from `testing/master`, `stable`, and `testing/1.8.9`. We will need to define a permalink for `legacy` (or come up with a new name for this) in order to make this all automated. In the meantime, I will maintain the `legacy` URL manually.

Jenkinsfile
klueska: Yes, def just declares a variable of any type. Specifically here we are defining a map (it infers it from the value).

Jenkinsfile
klueska: This is now running in the order 'master', 'stable', legacy'.

bamarni:
@klueska : is it still on the pipeline? Does it mean we would have 9 integration builds?

bamarni:
The jenkinsfiles structure has changed but that's still a valid point, I've opened https://jira.mesosphere.com/browse/DCOS_OSS-4035 to track the topic.

PR #1023, opened 2017-08-07
Update help links for dcos marathon add

bamarni:
The first one triggers a 404 and the second one has an unexisting anchor.


klueska:
the test failure here should be fixed if you rebase on master.

PR #1028, opened 2017-08-08
[DCOS_OSS-618] Ignore core.ssl_verify for external requests

bamarni:
Otherwise, when a cluster is configured with a custom CA path, it will
try to use it instead of the system's CA bundle and external requests
will fail.

In such cases, setting it to None will let the requests package enabling
it by default.

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -27,10 +27,40 @@ def _default_is_success(status_code):
 
     return 200 <= status_code < 300
 
+def _is_request_to_dcos(url, toml_config=None):
+    """Checks if a request is for the DC/OS cluster.
 
-def _verify_ssl(verify=None, toml_config=None):
-    """Returns whether to verify ssl
+    :param url: URL of the request
+    :type url: str
+    :param toml_config: cluster config to use
+    :type toml_config: Toml
+    :return: whether the request is for the cluster
+    :rtype: bool
+    """
 
+    if toml_config is None:
+        toml_config = config.get_config()
+
+    dcos_url = urlparse(config.get_config_val("core.dcos_url", toml_config))
+    cosmos_url = urlparse(
+        config.get_config_val("package.cosmos_url", toml_config))
+    parsed_url = urlparse(url)
+
+    # request should match scheme + netloc
+    def _request_match(expected_url, actual_url):
+        return expected_url.scheme == actual_url.scheme and \
+                    expected_url.netloc == actual_url.netloc
+
+    is_request_to_cluster = _request_match(dcos_url, parsed_url) or \
+        _request_match(cosmos_url, parsed_url)
+
+    return is_request_to_cluster
+
+def _verify_ssl(url, verify=None, toml_config=None):
+    """Returns whether to verify ssl for the given url
+
+    :param url: the target URL
+    :type url: str
     :param verify: whether to verify SSL certs or path to cert(s)
     :type verify: bool | str
     :param toml_config: cluster config to use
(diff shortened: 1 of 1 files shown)


dcos/http.py
```diff
@@ -190,6 +210,7 @@ def _request_match(expected_url, actual_url):
     if is_success(response.status_code):
         return response
     elif response.status_code == 401:
+        prompt_login = config.get_config_val("core.prompt_login", toml_config)
```
klueska: This change seems unrelated. It should either be a second commit in this pull request with a detailed commit message describing its purpose, or a separate PR.

dcos/http.py
```diff
@@ -27,6 +27,34 @@ def _default_is_success(status_code):
 
     return 200 <= status_code < 300
 
+def _is_request_to_cluster(url, toml_config=None):
```
klueska: Can we call this `_is_request_to_dcos()` instead?

klueska:
Also, we need to add an integration test before we can call this *done*.

bamarni:
I've just pushed a commit according to your comments.

What kind of integration test are you thinking of? As it involves an HTTPS request to an external URL I don't see an obvious way to do it.

dcos/http.py
```diff
@@ -27,6 +27,34 @@ def _default_is_success(status_code):
...
+    :type toml_config: Toml
+    :return: whether the request is for the cluster
+    :rtype: bool
+    """
+
+    if toml_config is None:
+        toml_config = config.get_config()
+
+    dcos_url = urlparse(config.get_config_val("core.dcos_url", toml_config))
+    cosmos_url = urlparse(
```
klueska: What is `cosmos_url`? And what happens if it's not set in the toml file (as far as I know, it's usually not set).

dcos/http.py
```diff
@@ -27,6 +27,34 @@ def _default_is_success(status_code):
...
+
+    # request should match scheme + netloc
+    def _request_match(expected_url, actual_url):
+        return expected_url.scheme == actual_url.scheme and \
+                    expected_url.netloc == actual_url.netloc
+
+    request_to_cluster = _request_match(dcos_url, parsed_url) or \
+        _request_match(cosmos_url, parsed_url)
+
+    return request_to_cluster
```
klueska: You can probably just return the or statement directly without assigning to an intermediate variable? Otherwise, you should rename this variable to something like `is_request_to_dcos`. Up to you.

dcos/http.py
```diff
@@ -86,7 +114,11 @@ def _request(method,
     if 'headers' not in kwargs:
         kwargs['headers'] = {'Accept': 'application/json'}
 
-    verify = _verify_ssl(verify, toml_config)
+    if _is_request_to_dcos(url, toml_config):
+        verify = _verify_ssl(verify, toml_config)
+    else:
+        # Ignore core.ssl_verify if request is outside the DC/OS cluster (DCOS_OSS-618)
+        verify = None
```
klueska: Instead of doing an if/else here, I would just do:
```
if verify and _is_request_to_dcos(url, toml_config):
    verify = _verify_ssl(verify, toml_config)
```

dcos/http.py
bamarni: I also don't have it set, in that case it is ignored, urlparse will return an empty ParseResult and this won't be matched.

dcos/http.py
bamarni: I'll set it as `is_request_to_dcos` for readability.

dcos/http.py
bamarni: Here `verify` can be a path, and even though I coulnd't see it in the codebase, my concern was if this function gets called with an explicit custom CA path for the cluster and an external https url. In such cases I wanted to force verify to `None`.

PR #1030, opened 2017-08-09
Add a message argument to DCOSAuthenticationException

bamarni:
This also fixes an issue where a string message is passed instead of a Response object.

cf. https://jira.mesosphere.com/browse/DCOS-16144

### DIFF
--- a/dcos/errors.py
+++ b/dcos/errors.py
@@ -49,12 +49,19 @@ class DCOSAuthenticationException(DCOSHTTPException):
 
     :param response: requests Response object
     :type response: Response
+    :param message: An optional message for the exception. This can be useful
+                    to provide the user with more meaningful information.
+    :type message: str
     """
-    def __init__(self, response):
+    def __init__(self, response, message=None):
+        if message is None:
+            message = "Authentication failed. Please run `dcos auth login`"
+
         self.response = response
+        self.message = message
 
     def __str__(self):
-        return "Authentication failed. Please run `dcos auth login`"
+        return self.message
 
 
 class DCOSAuthorizationException(DCOSHTTPException):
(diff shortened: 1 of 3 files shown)


dcos/errors.py
```diff
@@ -49,11 +49,17 @@ class DCOSAuthenticationException(DCOSHTTPException):
 
     :param response: requests Response object
     :type response: Response
+    :param message: An optional message for the exception.
```
jgehrcke: Should we make it more obvious that this message will actually be exposed to the user?

bamarni: I wouldn't assume it at the class definition level, as it depends on how it's used. Maybe I can turn the sentence like this?

> An optional message for the exception. This can be useful to provide the user with more meaningful information.

dcos/errors.py
```diff
@@ -49,11 +49,18 @@ class DCOSAuthenticationException(DCOSHTTPException):
 
     :param response: requests Response object
     :type response: Response
+    :param message: An optional message for the exception. This can be useful
+                    to provide the user with more meaningful information.
+    :type message: str
     """
-    def __init__(self, response):
+    def __init__(self, response, message=None):
         self.response = response
```
klueska: I wish we had a convention for composing messages from exceptions in the DC/OS CLI. I don't like the idea of just blindly passing in a message like this (without letting the exception itself supply some standard prefix for the message or something). It's probably OK for now, but we really need to architect this better in the future.

Also, can we reorganize this as:
```
def __init__(self, response, message=None):
    if message is None:
        message = "Authentication failed. Please run `dcos auth login`"

    self.response = response
    self.message = message
  		  
def __str__(self):
    return self.message
```

Finally, it seems strange to me to require a response to be sent in when we don't actually do anything with it.  I know you didn't add this, but it's weird nonetheless. I'd almost opt for removing it.

klueska:
ok to test

PR #1031, opened 2017-08-10
Add a test against infinite prompt during cluster setup

bamarni:
Prior to 1254a2f2b5210c68d86f4b16c4c6b16ad2e911ae and when running
"dcos cluster setup" command as non-interactive, the process would run
and ask for user input forever.

This adds an integration test which fails if the process doesn't exit by
itself with an error within 30 seconds.

https://jira.mesosphere.com/browse/DCOS-15590

### DIFF
--- a/cli/tests/integrations/helpers/common.py
+++ b/cli/tests/integrations/helpers/common.py
@@ -12,7 +12,7 @@ from six.moves import urllib
 from dcos import config, http
 
 
-def exec_command(cmd, env=None, stdin=None):
+def exec_command(cmd, env=None, stdin=None, timeout=None):
     """Execute CLI command
 
     :param cmd: Program and arguments
@@ -21,6 +21,10 @@ def exec_command(cmd, env=None, stdin=None):
     :type env: dict | None
     :param stdin: File to use for stdin
     :type stdin: file
+    :param timeout: If the process doesn't terminate after this timeout,
+                    it will get killed and a subprocess.TimeoutExpired
+                    exception will be raised. The timeout is in seconds.
+    :type timeout: int
     :returns: A tuple with the returncode, stdout and stderr
     :rtype: (int, bytes, bytes)
     """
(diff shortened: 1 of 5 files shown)


cli/tests/integrations/test_cluster.py
```diff
@@ -43,3 +44,21 @@ def test_rename():
...
+
+def test_setup_noninteractive():
+    """
+    Run "dcos cluster setup" command as non-interactive with a 30 sec timeout.
+    This makes sure the process doesn't prompt for input forever (DCOS-15590).
+    """
+
+    try:
+        returncode, stdout, _ = exec_command(
+            ['dcos', 'cluster', 'setup',
```
klueska: All arguments at the same level should either be on the same line or split onto individual lines, e.g. not
```
returncode, stdout, _ = exec_command(
    ['dcos', 'cluster', 'setup',
        'https://dcos.snakeoil.mesosphere.com'],
    timeout=30, stdin=subprocess.DEVNULL)
```
but rather
```
returncode, stdout, _ = exec_command(
    ['dcos', 'cluster', 'setup', 'https://dcos.snakeoil.mesosphere.com'],
    timeout=30,
    stdin=subprocess.DEVNULL)
```
or
```
returncode, stdout, _ = exec_command(
    ['dcos',
     'cluster',
     'setup',
     'https://dcos.snakeoil.mesosphere.com'],
    timeout=30,
    stdin=subprocess.DEVNULL)
```

It makes it much easier to look at and understand without haveing to spend alot of time reading the code.

cli/tests/integrations/helpers/common.py
```diff
@@ -36,7 +36,7 @@ def exec_command(cmd, env=None, stdin=None):
 
     # This is needed to get rid of '\r' from Windows's lines endings.
     stdout, stderr = [std_stream.replace(b'\r', b'')
-                      for std_stream in process.communicate()]
+                      for std_stream in process.communicate(timeout=timeout)]
```
klueska: Ideally, this would be split out into it's own separate commit. With the test change in its own commit as well.

bamarni:
I've just updated the PR, I was missing a cleanup in the helper and added a link about this.

I just noticed that this test only passes if I use the cloud formation template for an EE cluster. However for an open source or custom cluster, the `SHA256 fingerprint of cluster certificate bundle:` message might not show up, and instead 3 tabs will get opened in the browser and the test will fail... All this noise can be very annoying, is there a way to avoid this? Maybe detect the kind of cluster, or I can make an integration for another command which requires prompt but would fail more silently in such cases.

klueska:
retest this please

klueska:
ok to test

cli/tests/integrations/test_package.py
```diff
@@ -683,7 +697,8 @@ def test_install_no():
                 b'catalog-terms-conditions/#community-services\n'
                 b'A sample pre-installation message\n'
                 b'Continue installing? [yes/no] Exiting installation.\n'
-            )
+            ),
+            returncode=1
```
klueska: Hmm. At first I thought this seemed strange to return a `1` if the user supplied `no` as valid input. The reason being that the command exits cleanly, whether the user inputs `yes` or `no` (the only difference being whether the package was actually installed or not). I would have expected a `1` only if there was some internal error that made the command exit prematurely.

On thinking about it a bit more though, I agree with the semantics you have in place. Use a `0` to indicate that the command successfully carried out what you asked it to; use a `1` to indicate that it didn't.

An example demonstrating these semantics in a standard unix environment:
```
$ ls
foo        bar        baz
$ ls | grep foo
$ echo $?
0
$ ls | grep qux
$ echo $?
1
```

PR #1036, opened 2017-08-14
Revert "config set core.dcos_url" to its old behaviour

bamarni:
It nows only changes the config value and unsets the acs token, this
ensures backward compatibilty is preserved.

This command is still deprecated and users are encouraged to switch to
the "dcos cluster setup" command.

https://jira.mesosphere.com/browse/DCOS-16088

```
Setup flow for new users of the 0.4.x CLI:
$ export DCOS_CONFIG=<path/to/config> (optional)
  (default when not set is ~/.dcos/dcos.toml)
$ dcos config set core.dcos_url <url>
$ dcos config set core.ssl_verify [“true” | “false” | <path/to/ca_cert>]
$ dcos auth login
  (generates  / updates global config at $DCOS_CONFIG where all settings are set)

Setup flow for new users of the 0.5.x CLI:
$ export DCOS_DIR=<path/to/config_dir> (optional)
  (default when not set is ~/.dcos)
$ dcos cluster setup <url>
  (generates / updates per cluster config under $DCOS_DIR/clusters/<cluster_id> where all settings are set)
  (sets newly set up cluster config as the active one)

Automatic update of config when currently using the 0.4.x CLI and updating to new 0.5.x CLI:
$ export DCOS_CONFIG=<path/to/config> (optional)
  (default when not set is ~/.dcos/dcos.toml)
$ dcos <subcommand>

$ … download and install new dcos 0.5.x binary …

$ export DCOS_DIR=<path/to/config_dir> (optional)
  (default when not set is ~/.dcos)
$ dcos <subcommand>
  (triggers conversion from old config structure to new)
  (moves global config from $DCOS_CONFIG to $DCOS_DIR/clusters/<cluster_id>/dcos.toml)
  (sets newly converted cluster config as the active one)
  (DCOS_CONFIG is no longer honored)

Preserving backwards compatibility:
1. Before `dcos cluster setup <url>` is called, follow the old flow semantics
   and allow users to change the config pointed to by $DCOS_CONFIG
   (print warning message saying these calls are deprecated and recommend using `dcos cluster setup`)
2. After `dcos cluster setup` is called (or automatic conversion has occurred),
   update the ‘active’ cluster’s config whenever old flow commands are called
   (print different warning message saying the calls are deprecated and cluster config state may now be corrupted)
```

### DIFF
--- a/ci/Jenkinsfile-linux-integration
+++ b/ci/Jenkinsfile-linux-integration
@@ -129,30 +129,6 @@ class TestCluster implements Serializable {
 
         return script.readFile("${platform}_dcos_url")
     }
-
-    /**
-     * Retrieves the ACS Token of a cluster previously created using `create()`.
-     */
-    def getAcsToken() {
-        def dcosUrl = this.getDcosUrl()
-
-        /* In the future, consider doing the following with curl / jq instead
-           of inline python (however, jq is not installed on our windows
-           machines at the moment). */
-        script.sh """
-            python -c \
-                'import requests; \
-                 requests.packages.urllib3.disable_warnings(); \
-                 js={"uid":"${script.env.DCOS_ADMIN_USERNAME}", \
-                     "password": "${script.env.DCOS_ADMIN_PASSWORD}"}; \
-                 r=requests.post("http://${dcosUrl}/acs/api/v1/auth/login", \
-                                 json=js, \
-                                 verify=False); \
-                 print(r.json()["token"], end="")' \
-            > ${platform}_acs_token"""
-
-        return script.readFile("${platform}_acs_token")
-    }
 }
 
 
@@ -175,7 +151,6 @@ def testBuilder(String platform, String nodeId, String workspace = null) {
 
             try {
                 def dcosUrl = cluster.getDcosUrl()
-                def acsToken = cluster.getAcsToken()
 
                 node(nodeId) {
                     if (!workspace) {
@@ -199,8 +174,7 @@ def testBuilder(String platform, String nodeId, String workspace = null) {
                               credentialsId: '23743034-1ac4-49f7-b2e6-a661aee2d11b',
                               variable: 'CLI_TEST_SSH_KEY_PATH']]) {
 
-                            withEnv(["DCOS_URL=${dcosUrl}",
-                                     "DCOS_ACS_TOKEN=${acsToken}"]) {
+                            withEnv(["DCOS_URL=${dcosUrl}"]) {
                                 try {
                                     body()
                                 } catch(InterruptedException | hudson.AbortException e) {
@@ -256,14 +230,14 @@ builders['linux-tests'] = testBuilder('linux', 'py35', '/workspace')({
         dir('dcos-cli/cli') {
             sh '''
                export PYTHONIOENCODING=utf-8; \
+               make binary; \
                export CLI_TEST_SSH_USER=centos; \
                export CLI_TEST_MASTER_PROXY=true; \
-               export DCOS_CONFIG=tests/data/dcos.toml; \
-               chmod 600 ${DCOS_CONFIG}; \
-               echo dcos_acs_token = \\\"${DCOS_ACS_TOKEN}\\\" >> ${DCOS_CONFIG}; \
-               cat ${DCOS_CONFIG}; \
-               unset DCOS_URL; \
-               unset DCOS_ACS_TOKEN; \
+               dist/dcos cluster setup dcos.snakeoil.mesosphere.com \
+                   --insecure --username=${DCOS_ADMIN_USERNAME} \
+                   --password-env=DCOS_ADMIN_PASSWORD; \
+               dist/dcos config set core.reporting false; \
+               dist/dcos config set core.timeout 5; \
                make test-binary'''
         }
     }
--- a/ci/Jenkinsfile-linux-integration
+++ b/ci/Jenkinsfile-linux-integration
@@ -113,9 +113,9 @@ class TestCluster implements Serializable {
     }
 
     /**
-     * Retrieves the URL of a cluster previously created using `create()`.
+     * Retrieves the IP of a cluster previously created using `create()`.
      */
-    def getDcosUrl() {
+    def getDcosIp() {
         /* In the future, consider doing the following with jq instead of
            inline python (however, jq is not installed on our windows machines
            at the moment). */
@@ -150,7 +150,7 @@ def testBuilder(String platform, String nodeId, String workspace = null) {
             }
 
             try {
-                def dcosUrl = cluster.getDcosUrl()
+                def dcosIp = cluster.getDcosIp()
 
                 node(nodeId) {
                     if (!workspace) {
@@ -174,19 +174,19 @@ def testBuilder(String platform, String nodeId, String workspace = null) {
                               credentialsId: '23743034-1ac4-49f7-b2e6-a661aee2d11b',
                               variable: 'CLI_TEST_SSH_KEY_PATH']]) {
 
-                            withEnv(["DCOS_URL=${dcosUrl}"]) {
+                            withEnv(["DCOS_IP=${dcosIp}"]) {
                                 try {
                                     body()
                                 } catch(InterruptedException | hudson.AbortException e) {
                                     echo(
                                         "Build interrupted. The DC/OS cluster at" +
-                                        " ${dcosUrl} will be destroyed.")
+                                        " ${dcosIp} will be destroyed.")
                                     destroyCluster = true
                                     throw e
                                 } catch (Exception e) {
                                     echo(
                                         "Build failed. The DC/OS cluster at" +
-                                        " ${dcosUrl} will remain temporarily" +
+                                        " ${dcosIp} will remain temporarily" +
                                         " active so you can debug what went" +
                                         " wrong.")
                                     destroyCluster = false
@@ -220,8 +220,8 @@ builders['linux-tests'] = testBuilder('linux', 'py35', '/workspace')({
         sh '''
            rm -rf ~/.dcos; \
            grep -q "^.* dcos.snakeoil.mesosphere.com$" /etc/hosts && \
-           sed -iold "s/^.* dcos.snakeoil.mesosphere.com$/${DCOS_URL} dcos.snakeoil.mesosphere.com/" /etc/hosts || \
-           echo ${DCOS_URL} dcos.snakeoil.mesosphere.com >> /etc/hosts'''
+           sed -iold "s/^.* dcos.snakeoil.mesosphere.com$/${DCOS_IP} dcos.snakeoil.mesosphere.com/" /etc/hosts || \
+           echo ${DCOS_IP} dcos.snakeoil.mesosphere.com >> /etc/hosts'''
 
         dir('dcos-cli') {
             sh 'make test'
--- a/cli/dcoscli/config/main.py
+++ b/cli/dcoscli/config/main.py
@@ -5,7 +5,6 @@ import docopt
 import dcoscli
 from dcos import cmds, config, emitting, http, util
 from dcos.errors import DCOSException, DefaultError
-from dcoscli.cluster.main import setup
 from dcoscli.subcommand import default_command_info, default_doc
 from dcoscli.util import decorate_docopt_usage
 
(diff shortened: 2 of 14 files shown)


Jenkinsfile
```diff
@@ -330,12 +304,10 @@ builders['linux-tests'] = testBuilder('linux', 'py35', '/workspace')({
         dir('dcos-cli/cli') {
             sh '''
                export PYTHONIOENCODING=utf-8; \
-               export DCOS_CONFIG=tests/data/dcos.toml; \
-               chmod 600 ${DCOS_CONFIG}; \
-               echo dcos_acs_token = \\\"${DCOS_ACS_TOKEN}\\\" >> ${DCOS_CONFIG}; \
-               cat ${DCOS_CONFIG}; \
+               dcos cluster setup ${DCOS_URL} --insecure --username=${DCOS_ADMIN_USERNAME} --password-env=DCOS_ADMIN_PASSWORD; \
```
klueska: We can't do this because we don't have the `dcos` command built yet.

Jenkinsfile
```diff
@@ -355,12 +327,10 @@ builders['mac-tests'] = testBuilder('mac', 'mac')({
...
```
klueska: ditto

Jenkinsfile
```diff
@@ -371,18 +341,17 @@ builders['windows-tests'] = testBuilder('windows', 'windows', 'C:\\windows\\work
...
+
+        bat 'echo %DCOS_URL% dcos.snakeoil.mesosphere.com >> C:\\windows\\system32\\drivers\\etc\\hosts'
 
         dir('dcos-cli/cli') {
             bat '''
                 bash -c " \
                 export PYTHONIOENCODING=utf-8; \
-                export DCOS_CONFIG=tests/data/dcos.toml; \
-                cat ${DCOS_CONFIG}; \
+                dcos cluster setup ${DCOS_URL} --insecure --username=${DCOS_ADMIN_USERNAME} --password-env=DCOS_ADMIN_PASSWORD; \
```
klueska: ditto

cli/dcoscli/config/main.py
```diff
@@ -86,28 +85,18 @@ def _set(name, value):
     """
 
     if name == "core.dcos_url":
-        return _cluster_setup(value)
+        notice = (
+            "This config property is being deprecated. "
+            "To setup the CLI to talk to your cluster, please run "
+            "`dcos cluster setup <dcos_url>`.")
+        emitter.publish(DefaultError(notice))
```
klueska: This might be enough, but I'm not 100% sure. Where is `config` pointing if you don't have a cluster set up yet? If you do have a cluster set up (using the `cluster setup` command), will this just change the URL for whichever cluster is the active cluster? 

To verify this works, I'd want to see some extra integration tests that tested the interaction of using the deprecated commands in conjunction with the new ones.

Jenkinsfile
bamarni: I see.. This was just a first WIP attempt to iterate over the CI build to fix it. After switching to the previous behavior there were an integration test failing for `dcos cluster list` (it should return one but there were none).

I haven't yet exactly figured out how to fix this, I have to investigate further.

cli/dcoscli/config/main.py
bamarni: `config` will point to the current active cluster, if there is none it will just error out with `No cluster is attached. Please run dcos cluster attach <cluster-name>`. So it cannot work if there is no cluster setup.

I think the DCOS_CONFIG stuff is also deprecated? (https://github.com/dcos/dcos-cli/pull/943). It seems like setting-up a cluster is the only way to go now, if I cannot run the cluster setup for CI I'd have to find a way to create manually the structure folder (with the "attached" file etc.).

klueska: Based on our offline discussion, here's the semantics I think we want to have / preserve:
```
Setup flow for new users of the 0.4.x CLI:
$ export DCOS_CONFIG=<path/to/config> (optional)
   (default when not set is ~/.dcos/dcos.toml)
$ dcos config set core.dcos_url <url>
$ dcos config set core.ssl_verify [“true” | “false” | <path/to/ca_cert>]
$ dcos auth login
   (generates  / updates global config at $DCOS_CONFIG where all settings are set)

Setup flow for new users of the 0.5.x CLI:
$ export DCOS_DIR=<path/to/config_dir> (optional)
   (default when not set is ~/.dcos)
$ dcos cluster setup <url>
   (generates / updates per cluster config under $DCOS_DIR/clusters/<cluster_id> where all settings are set)
   (sets newly set up cluster config as the active one)

Automatic update of config when currently using the 0.4.x CLI and updating to new 0.5.x CLI:
$ export DCOS_CONFIG=<path/to/config> (optional)
  (default when not set is ~/.dcos/dcos.toml)
$ dcos <subcommand>

$ … download and install new dcos 0.5.x binary …

$ export DCOS_DIR=<path/to/config_dir> (optional)
   (default when not set is ~/.dcos)
$ dcos <subcommand>
   (triggers conversion from old config structure to new)
   (moves global config from $DCOS_CONFIG to $DCOS_DIR/clusters/<cluster_id>/dcos.toml)
   (sets newly converted cluster config as the active one)
   (DCOS_CONFIG is no longer honored)

Preserving backwards compatibility:
1. Before `dcos cluster setup <url>` is called, follow the old flow semantics
    and allow users to change the config pointed to by $DCOS_CONFIG
    (print warning message saying these calls are deprecated and recommend using `dcos cluster setup`)
2 .After `dcos cluster setup` is called (or automatic conversion has occurred),
    update the ‘active’ cluster’s config whenever old flow commands are called
    (print different warning message saying the calls are deprecated and cluster config state may now be corrupted)
```

bamarni:
@klueska : looking at this more in depth, it seems like the migration part is already in place :)

https://github.com/dcos/dcos-cli/blob/63bcddaa205695e09874ad8c24f89058b4f7405e/cli/dcoscli/main.py#L72-L73

I'm currently running different cluster setup commands with 0.4 / 0.5 / auto-migration in order to compare config outputs.

bamarni:
run all integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

ci/Jenkinsfile-linux-integration
```diff
@@ -233,18 +207,16 @@ builders['linux-tests'] = testBuilder('linux', 'py35', '/workspace')({
...
             sh '''
                export PYTHONIOENCODING=utf-8; \
-               export DCOS_CONFIG=tests/data/dcos.toml; \
-               chmod 600 ${DCOS_CONFIG}; \
-               echo dcos_acs_token = \\\"${DCOS_ACS_TOKEN}\\\" >> ${DCOS_CONFIG}; \
-               cat ${DCOS_CONFIG}; \
-               unset DCOS_URL; \
-               unset DCOS_ACS_TOKEN; \
+               make binary; \
+               dist/dcos cluster setup dcos.snakeoil.mesosphere.com --insecure --username=${DCOS_ADMIN_USERNAME} --password-env=DCOS_ADMIN_PASSWORD; \
```
klueska: I'd rather leave this as it was before so that we can test that config migration works properly.

cli/dcoscli/main.py
```diff
@@ -69,19 +69,22 @@ def _main():
...
-        if args['--help']:
-            command = "help"
-        else:
-            return dcos_help()
+        return dcos_help()
+
+    if constants.DCOS_CONFIG_ENV in os.environ:
+        msg = ('{} is deprecated, please consider using '
+               '`dcos cluster setup <dcos_url>`.')
+        err = errors.DefaultError(msg.format(constants.DCOS_CONFIG_ENV))
```
klueska: Does this mean that this warning is going to be printed for every command that gets issued so long as this environment variable is set? That seems like a little much.

bamarni: Yes, should I make this printed only before the migration occurs?

ci/Jenkinsfile-linux-integration
bamarni: on the other hand right now we don't have tests for `dcos cluster setup`, it also simplifies the jenkinsfile.

cli/dcoscli/main.py
klueska: That seems more reasonable to me. I'd also print a longer message stating how this variable will now be ignored going forward, and that DCOS_DIR is the replacement for relocating where the (now) managed configs live. It also seems like we should probably prompt to verify that the user has read this warning and is OK with migrating.

bamarni: I think I still miss this which isn't implemented when `DCOS_CONFIG` is set : 
> After `dcos cluster setup` is called (or automatic conversion has occurred),
   update the ‘active’ cluster’s config whenever old flow commands are called
   (print different warning message saying the calls are deprecated and cluster config state may now be corrupted)

ci/Jenkinsfile-linux-integration
bamarni: Given the number of different scenarios for the migration, I can probably create a dedicated test suite for these, WDYT? It seems hard to add such tests in our integration suite, as it assumes an already-setup cluster, some env variables already set.

bamarni:
run linux integration tests

bamarni:
> It also seems like we should probably prompt to verify that the user has read this warning and is OK with migrating.

How could we do that so it doesn't break automated scripts?

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run all integration tests

bamarni:
run linux integration tests

bamarni:
run mac integration tests

bamarni:
run linux integration tests

bamarni:
run mac integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

dcos/config.py
```diff
@@ -173,6 +173,15 @@ def get_config_val_envvar(name, config=None):
...
         config = get_config()
 
+    # This is a list of configs which cannot be overriden by env variables
+    env_blacklist = [
+        'core.dcos_url',
+        'core.dcos_acs_token',
+        'core.mesos_master_url',
+    ]
+    if name in env_blacklist:
+        return config.get(name), None
```
bamarni: As we're automatically migrating to the clusters dir structure and ignoring `DCOS_CONFIG`, should we also ignore `DCOS_URL` and `DCOS_ACS_TOKEN` env variables? As the `dcos cluster ...` commands should be the way to go to manage different clusters.

I'm not sure about `DCOS_MASTER_MESOS_URL` though.

bamarni:
run all integration tests

ci/Jenkinsfile-linux-integration
bamarni: I've added integration tests, cf. 5a4d013c169e9c1be8ddde8c74834a34d65959f5

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

cli/dcoscli/config/main.py
darkonie: i think this error message is a little bit confusing
for example:
```
dcos config set core.dcos_url http://52.37.221.58
This config property is being deprecated. To setup the CLI to talk to your cluster, please run `dcos cluster setup <dcos_url>`.
[core.dcos_url]: changed from 'https://34.209.195.182' to 'http://52.37.221.58'
[core.dcos_acs_token]: removed
```
It says that this option is being deprecated and suggests to use a new way of setting up a cluster, but it also configures the CLI to talk to cluster. I think we need to rephrase the message. We probably just want to announce that this subcommand is being deprecated and in future user should be using `dcos cluster setup`

darkonie:
also looks like the integration tests are failing

cli/tests/integrations/helpers/common.py
```diff
@@ -261,7 +254,7 @@ def update_config(name, value, env=None):
     finally:
         # return config to previous state
         if result is not None:
-            config_set(name, result, env)
+            config_set(name, "%s" % result, env)
```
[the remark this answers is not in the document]
bamarni: it is to convert the value into a string, using `str(result)` is simpler indeed :+1:

cli/dcoscli/config/main.py
bamarni: I can rephrase it, currently it also implies that the config is deprecated, but it's actually only the way to set it which is different, how about something like this : 

> Setting the `core.dcos_url` config directly is deprecated, please use the `dcos cluster config <dcos_url>` command instead.

bamarni:
run all integration tests

bamarni:
> would you also explain why do we need to remove a bunch of stuff including functions getAcsToken

I've adapted the CI to use the `dcos cluster setup` command instead of setting-up a config file manually. I've also added integration tests for different config files (with or without acs token, with `DCOS_CONFIG` and/or `DCOS_DIR`) to make sure migration happens as expected.

bamarni:
run all integration tests

bamarni:
@klueska @darkonie : do you have any other concern apart from the deprecation warning message? I tried to address the rest in the previous days and added more integration tests for migration.

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run linux integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

PR #1040, opened 2017-08-18
Karl/fix cleanup

karlkfi:


karlkfi:
replacing #1039 to see if it fixed the commit that gets checked out in jenkins...

It was trying to do `git rev-parse origin/PR-1039^{commit}` which was failing because there's no ref named that...

PR #1043, opened 2017-08-21
tests/ssh: use wait() instead of sleep()+poll()

bamarni:
Was there a specific reason to use `poll()` instead of `wait()` here?

`poll()` would return immediately after the sleep period, even though the SSH command didn't finish. Using `wait()` with a timeout blocks for a given period for the command to finish, otherwise raises an exception.

Without this change I'm experiencing some random failures on `test_node_ssh_with_command()`, where return code is `None` instead of the expected `0`.

### DIFF
--- a/cli/tests/integrations/helpers/common.py
+++ b/cli/tests/integrations/helpers/common.py
@@ -295,10 +295,8 @@ def ssh_output(cmd):
     # ssh must run with stdin attached to a tty
     proc, master = popen_tty(cmd)
 
-    # wait for the ssh connection
-    time.sleep(5)
-
-    proc.poll()
+    # wait for the ssh command to finish
+    proc.wait(30)
     returncode = proc.returncode
 
     # kill the whole process group
--- a/cli/tests/integrations/helpers/common.py
+++ b/cli/tests/integrations/helpers/common.py
@@ -282,11 +282,14 @@ def popen_tty(cmd):
     return (proc, master)
 
 
-def ssh_output(cmd):
+def ssh_output(cmd, wait=False):
     """ Runs an SSH command and returns the stdout/stderr/returncode.
 
     :param cmd: command to run
     :type cmd: str
+    :param wait: whether to wait for the SSH command to finish, or kill it
+                 after 5 seconds.
+    :type wait: bool
     :rtype: (str, str, int)
     """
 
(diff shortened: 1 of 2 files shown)


bamarni:
Hum after running the full test suite with this change some other tests are failing because of the timeout... It seems like some SSH commands are login sessions, so they are not expected to terminate by themselves and the process group is killed manually few lines below.

Would that be an option to use a simple command to run (eg. /bin/true) instead of no command?

bamarni:
Okay I've pushed a second commit now. This isn't the most elegant solution though, I'm open to suggestions on this.

cli/tests/integrations/helpers/common.py
```diff
@@ -295,15 +298,19 @@ def ssh_output(cmd):
     # ssh must run with stdin attached to a tty
     proc, master = popen_tty(cmd)
 
-    # wait for the ssh command to finish
-    proc.wait(30)
-    returncode = proc.returncode
+    if wait:
+        # wait for the ssh command to finish
+        proc.wait(30)
```
klueska: Actually, I'm confused, is this commit modifying the same function as the previous commit? maybe we can just squash both changes into one? Why did you have to add the option for `wait` vs. `poll` at all? Under what conditions is `expected_returncode != None` which would trigger you to use `wait` instead of `poll`?

bamarni:
> Under what conditions is expected_returncode != None which would trigger you to use wait instead of poll?

This helper methods is called from `test_node` integration tests, and used for ssh related tests (basically `dcos node ssh <command>`). `returncode != None` means there is an expected return code (the command is expected to finish). Because on the other hand, some test commands are not expected to finish, this happens when this is a login session (there is no `<command>` argument), so the process has to be killed manually.

When there is an expected returncode (which is when there is a `<command>` argument), I saw a problem where the 5 seconds hard sleep wasn't always enough for my command to finish and the test was flaky, I think in these situations at least it is cleaner to have a soft sleep (waiting for the command to finish, or erroring after 30 seconds).

bamarni:
But as I mentioned previously this parameter is not a very elegant solution and there is probably a better approach, eg. some refactoring or a change in the ssh testing method.

For example the `ssh_output()` helper contains nothing specific to ssh, it just runs a command with a pseudo-terminal.

PR #1045, opened 2017-08-24
Clarifications about --all and --app-id

bamarni:
The first commit updates the `--all` option description, I think we mean `applications` instead of `packages` here.

The second one is about making those options mutually exclusive, as it doesn't make sense to use them together. So when using both of them, one would get an `Invalid subcommand usage` error printed out.


bamarni:
run windows integration tests

bamarni:
run all integration tests

PR #1049, opened 2017-08-25
Also migrate deprecated subcommand directories when migrating deprecated cluster configs.

klueska:
The 0.5.x CLI will automatically migrate config files created by the the 0.4.x CLI from:
```
$DCOS_DIR/dcos.toml
to
$DCOS_DIR/clusters/<cluster-id>/dcos.toml
```

However, we were previously not migrating the `subcommands` folder from the old location to the new. The rationale for this was that the 0.5.x CLI still maintains the notion of "globally installed" subcommands, and we could just treat the legacy `subcommands` folder as our new global one.

The problem with this approach is that commands contained in the legacy `subcommands` folder may never have been intended for multiple clusters to use (in the 0.4.x CLI, there was no notion of multi-cluster support). As such, it is unreasonable to assume that the commands contained in this folder should be accessible globally.

A use case in the wild where this actually became a problem is outlined in the following ticket:
https://jira.mesosphere.com/browse/DCOS_OSS-1588

This PR fixes this by separating the concepts of a "global" subcommand folder from a "deprecated" subcommand folder and migrates the deprecated folder into the new config location at the same time it migrates the config itself.

When new subcommands are then added globally, they are added into the new "global" subcommands folder.

As before, both the deprecated `config.yaml` file and the deprecated `subcommand` folder remain untouched in case users ever want to downgrade from the 0.5.x CLI back to the 0.4.x CLI and still access the same cluster they had before the upgrade.

### DIFF
--- a/dcos/constants.py
+++ b/dcos/constants.py
@@ -14,12 +14,16 @@ DCOS_CLUSTER = 'DCOS_CLUSTER'
 DCOS_CLUSTER_ATTACHED_FILE = "attached"
 """Name of the file indicating the current attached cluster"""
 
-DCOS_SUBCOMMAND_ENV_SUBDIR = 'env'
-"""In a package's directory, this is the cli contents subdirectory."""
+DCOS_GLOBAL_SUBCOMMAND_SUBDIR = 'global_subcommands'
+"""Name of the subdirectory that contains all of the global subcommands. This
+is relative to the location of the executable."""
 
 DCOS_SUBCOMMAND_SUBDIR = 'subcommands'
 """Name of the subdirectory that contains all of the subcommands. This is
 relative to the location of the executable."""
+
+DCOS_SUBCOMMAND_ENV_SUBDIR = 'env'
+"""In a package's directory, this is the cli contents subdirectory."""
 
 DCOS_CONFIG_ENV = 'DCOS_CONFIG'
 """Name of the environment variable pointing to the DC/OS config."""
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -23,8 +23,8 @@ def move_to_cluster_config():
     :rtype: None
     """
 
-    global_config = config.get_global_config()
-    dcos_url = config.get_config_val("core.dcos_url", global_config)
+    deprecated_config = config.get_deprecated_config()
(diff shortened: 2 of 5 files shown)


dcos/constants.py
```diff
@@ -14,13 +14,17 @@
 DCOS_CLUSTER_ATTACHED_FILE = "attached"
 """Name of the file indicating the current attached cluster"""
 
-DCOS_SUBCOMMAND_ENV_SUBDIR = 'env'
-"""In a package's directory, this is the cli contents subdirectory."""
+DCOS_GLOBAL_SUBCOMMAND_SUBDIR = 'global_subcommands'
```
bamarni: As the PR description states : `we can just treat the legacy subcommands folder as our new global one.`, why is there this new directory introduced?

dcos/constants.py
```diff
@@ -14,13 +14,17 @@
...
+"""Name of the subdirectory that contains all of the global subcommands. This
+is relative to the location of the executable."""
```
bamarni: `relative to the location of the executable` -> `relative to $DCOS_DIR`?

bamarni:
A few questions about the global subcommand dir otherwise LGTM. Just out of curiosity, do you have any example of such a global subcommand?

dcos/constants.py
bamarni: Ah my bad, this was the incorrect previous usage, that you changed!

klueska:
I'm actually not a huge fan of the global subcommand because when you `setup` a cluster for a newer version of DC/OS, it's possible that the (already installed) global subcommand will fail on that new cluster. It was a requirement from product though, so @tamar implemented it as requested.

I do understand the overall rationale; You want to install a package once (e.g. dcos-enterprise-cli) and you want the subcommands contained in that package to work for all clusters you manage without having to reinstall the package every time.

It would be good to sit down and think through a more robust flow for this so that we can give both a good user experience while at the same time meeting the product requirement. I'm open to suggestions.

bamarni:
Closing because the python code was moved to the core and we don't do the migration on the new CLI (as it doesn't support the legacy 0.4.x CLI / structure).

PR #1056, opened 2017-09-07
Updated Jenkinsfiles to remove reliance on custom CF templates and fix `ssh` tests.

klueska:
Previously, we relied on custom cloud formation templates to launch the DC/OS clusters required by our CI jobs. This commit updates all of our integration test Jenkinsfiles to instead use dcos-launch in "onprem" mode with the proper configuration setup instead of running against these custom templates.

As part of this PR, it became obvious that some extra work was needed to get our CI system working properly for our various SSH tests with this change. The commits that fix those issues are included in this PR.

PR #1057, opened 2017-09-07
Mention CLI_TEST_MASTER_PROXY=1 in the README

bamarni:
When running integration tests from a machine which cannot reach all
nodes through the network, one would get this failure :

    tests/integrations/test_node.py:257: AssertionError
    ------------------------------ Captured stdout call
    -------------------------------
    SSH COMMAND: ssh-agent /bin/bash -c "ssh-add
    /home/bamarni/.ssh/mesosphere_shared 2> /dev/null && dcos node ssh
    --option StrictHostKeyChecking=no --leader"
    SSH STDOUT:
    SSH STDERR: Running `ssh -t -o StrictHostKeyChecking=no core@10.0.6.194`

    If you are running this command from a separate network than DC/OS,
    consider using `--master-proxy` or `--proxy-ip`
    ssh: connect to host 10.0.6.194 port 22: Network is unreachable

This is for example the case when running integration tests locally with
the default cloudformation template.

https://jira.mesosphere.com/browse/DCOS_OSS-1624

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,6 +141,13 @@ environment appropriately.
     $ export CLI_TEST_SSH_KEY_PATH=<path-to-ssh-key>
 
 
+#. (optional) Enable :code:`CLI_TEST_MASTER_PROXY` in case your private agents
+   are not reachable from the machine you're running tests on. This will proxy
+   the SSH connection throught the master node during integration tests.
+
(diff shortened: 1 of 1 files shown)


bamarni:
This was addressed during #1056

PR #1063, opened 2017-09-15
Simplify some tests

bamarni:
- Use pytest.raises instead of try / except / assert
- don't t assert during exception handling (as it can cause an unreadable error trace)
- Use util.tempdir() instead of a custom function


bamarni:
run all integration tests

bamarni:
run linux integration tests

bamarni:
run mac integration tests

PR #1065, opened 2017-09-21
Ken/composable quiet jobs

kensipe:
* order tables in desc order which is most relevant
* job history has last and quiet, providing the ability to get a list of ids only and just the last id
* job history title for ID changed to Task ID (making it clear that is what it is)
* job show runs shows tasks Ids
* job show runs and list now have quiet ID only capabilities

Provides the ability to do the following: 
`dcos task log $(dcos job history test -f -l -q) --completed stderr`

which provides access to the stderr log for the last failure for the job history of `test`.


kensipe:
this is a fix for:  https://jira.mesosphere.com/browse/DCOS-18815

cli/tests/fixtures/metronome.py
```diff
@@ -107,14 +107,14 @@ def job_history_fixture():
...
-        "finishedAt": "2017-03-31T21:05:46.805+0000",
-        "id": "20170331210532uxgVF"
+        "createdAt": "2017-09-21T21:20:22.581+0000",
+        "finishedAt": "2017-09-21T21:20:25.315+0000",
+        "id": "20170921212022CrNjS"
       },
       {
-        "createdAt": "2017-03-31T21:05:30.613+0000",
-        "finishedAt": "2017-03-31T21:05:41.740+0000",
-        "id": "20170331210530QHpRU"
```
bamarni: If these tests were failing we should improve them not to be time-dependent, otherwise please revert, this introduces unnecessary changes to the review.

cli/dcoscli/data/help/job.txt
```diff
@@ -12,13 +12,13 @@ Description:
         dcos job update <job-file>
         dcos job kill <job-id> [<run-id>][--all]
         dcos job run <job-id> [--json]
-        dcos job list [--json]
+        dcos job list [--json --quiet]
```
bamarni: Should `--json` and `--quiet` be mutually exclusive? As this doesn't make sense to use them together.

cli/dcoscli/job/main.py
```diff
@@ -112,12 +112,12 @@ def _cmds():
...
         cmds.Command(
             hierarchy=['job', 'list'],
-            arg_keys=['--json'],
+            arg_keys=['--json', '--quiet'],
             function=_list),
 
         cmds.Command(
             hierarchy=['job', 'history'],
-            arg_keys=['<job-id>', '--json', '--show-failures'],
+            arg_keys=['<job-id>', '--json', '--failures', '--last', '--quiet'],
```
bamarni: changing `--show-failures` to `--failures` might break user scripts, we should avoid BC break unless absolutely necessary.

cli/tests/fixtures/metronome.py
kensipe: yeah... tests and changes were frustrating... always have been.   the alignment of the text is checked, however there are either extra characters or linefeeds on mac that do not allow for easy creation of the test files... I'll take a look again.

cli/dcoscli/job/main.py
kensipe: understand... the original flag wasn't thought out.

cli/dcoscli/data/help/job.txt
kensipe: definitely!

cli/tests/fixtures/metronome.py
kensipe: I reverted to reduce the change... however there are 2 changes that require changes to the test and results:
1. ID -> TASK ID
2. the order is sorted by time 
none of which requires a change to the fixture.

kensipe:
PR concerns addressed except `--show-failures`

let me know if you require that back in there... It will be easy to put back... I just don't like it (and I'm the one that originally put it there).    It is useful and needed information but is better to have in isolation with the `--failures` flag.

Thanks!!

bamarni:
@kensipe : I agree `--failures` is a better name for that flag, I'm just concerned about BC. What would you think about adding something like this [here](https://github.com/dcos/dcos-cli/blob/8feab71b0228f34c028a68e997996fa0d6889081/cli/dcoscli/job/main.py#L48) : 

``` python
@decorate_docopt_usage
def _main(argv):
    for i, arg in enumerate(argv):
        if arg == '--show-failures':
            argv[i] = '--failures'
            warning = ("'--show-failures' is deprecated, "
                "please use '--failures' instead.\n")
            emitter.publish(DefaultError(warning))

    args = docopt.docopt(
        default_doc("job"),
        argv=argv,
        version='dcos-job version {}'.format(dcoscli.version))

    return cmds.execute(_cmds(), args)
```

This will give users some time to update their scripts until we eventually remove support for this flag in the future.

kensipe:
@bamarni makes a lot of sense... embarrassed I didn't add it initially.   thx!

bamarni:
run linux integration tests

PR #1066, opened 2017-09-26
Update test_remove_all which was failing after #1036

bamarni:
While they were passing independently, the combination of #1015 and #1036 broke the CI.

This PR refactors the failing test a bit, mainly to get the current config dir before setting the DCOS_DIR env variable.


bamarni:
run linux integration tests

bamarni:
run linux integration tests

PR #1072, opened 2017-10-04
Ignore outputs while opening a browser window

bamarni:
Otherwise users might see warning messages while they are prompted for
the token.

cf. https://jira.mesosphere.com/browse/DCOS_OSS-1508
cf. https://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/

### DIFF
--- a/dcos/auth.py
+++ b/dcos/auth.py
@@ -5,7 +5,6 @@ import time
 import webbrowser
 
 import jwt
-
 from six.moves import urllib
 from six.moves.urllib.parse import urlparse
 
@@ -67,7 +66,8 @@ def _prompt_user_for_token(url, token_type):
     msg = msg.lstrip().format(url=url, token_type=token_type)
 
     try:
-        webbrowser.open_new_tab(url)
+        with util.silent_output():
+            webbrowser.open_new_tab(url)
     except webbrowser.Error as exc:
         logger.warning(
             'Exception occurred while calling webbrowser.open(%r): %s',
--- a/dcos/util.py
+++ b/dcos/util.py
@@ -1,6 +1,7 @@
 import collections
 import concurrent.futures
 import contextlib
+import ctypes
 import functools
 import hashlib
 import json
@@ -96,6 +97,52 @@ def remove_path_on_error(path):
     except:
         shutil.rmtree(path, ignore_errors=True)
         raise
+
+
+@contextlib.contextmanager
+def silent_output():
+    """A context manager for suppressing stdout / stderr, it sets their file
+    descriptors to os.devnull and then restores them.
+
+    This is helpful for inhibiting output from C code or subprocesses, as
+    replacing sys.stdout and sys.stderr wouldn't be enough.
+    """
+
+    libc = ctypes.CDLL(None)
+    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
+    c_stderr = ctypes.c_void_p.in_dll(libc, 'stderr')
+
+    # Original fds stdout/stderr point to. Usually 1 and 2 on POSIX systems.
+    original_stdout_fd = sys.stdout.fileno()
+    original_stderr_fd = sys.stderr.fileno()
+
+    # Save a copy of original outputs fds
+    saved_stdout_fd = os.dup(original_stdout_fd)
+    saved_stderr_fd = os.dup(original_stderr_fd)
+
+    # A fd for the null device
+    devnull = open(os.devnull, 'w')
+
+    def _redirect_outputs(stdout_to_fd, stderr_to_fd):
+        """Redirect stdout/stderr to the given file descriptors."""
+        # Flush outputs, including C-level buffers
+        libc.fflush(c_stdout)
+        libc.fflush(c_stderr)
+        sys.stderr.flush()
+        sys.stdout.flush()
+
+        # Make original stdout / stderr fds point to the same file as to_fd
+        os.dup2(stdout_to_fd, original_stdout_fd)
+        os.dup2(stderr_to_fd, original_stderr_fd)
+
+    try:
+        # Redirect outputs to os.devnull
+        _redirect_outputs(devnull.fileno(), devnull.fileno())
+        yield
+    finally:
+        # redirect outputs back to the original fds
+        _redirect_outputs(saved_stdout_fd, saved_stderr_fd)
+        devnull.close()
 
 
 def sh_copy(src, dst):


bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run linux integration tests

dcos/auth.py
```diff
@@ -3,7 +3,6 @@
 import textwrap
 import time
 import webbrowser
-
 import jwt
```
jgehrcke: The typical convention as of PEP8 is to have three import blocks. The first imports from the standard library (which is everything up to `time` here), the second imports external and established third-party modules (which is `jwt`, but probably also `webbrowser`), and the third imports project-local modules/packages.

Not sure if that convention is adhered to in the CLI code :-)

dcos/util.py
```diff
@@ -98,6 +100,63 @@ def remove_path_on_error(path):
...
+    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
+    c_stderr = ctypes.c_void_p.in_dll(libc, 'stderr')
+
+    # Original fds stdout/stderr point to. Usually 1 and 2 on POSIX systems.
+    original_stdout_fd = sys.stdout.fileno()
+    original_stderr_fd = sys.stderr.fileno()
+
+    def _redirect_outputs(to_fd):
+        """Redirect stdout/stderr to the given file descriptor."""
+        # Flush the C-level stdout / stderr buffers
```
jgehrcke: "the C-level stdout / stderr buffers" is not well-defined in my mind... what is that supposed to be? Why is `libc` interaction required here?

I feel like I have messed quite a bit with the CPython I/O system on various different platforms and I never needed to import `ctypes` for that.

Why is e.g. `sys.stdout.flush()` not sufficient?

dcos/util.py
```diff
@@ -98,6 +100,63 @@ def remove_path_on_error(path):
         raise
 
 
+@contextlib.contextmanager
+def silent_outputs():
+    """A context manager for suppressing stdout / stderr, it sets their file
+    descriptors to os.devnull and then restores them.
```
jgehrcke: I am kind of afraid about the redirection to /dev/null. I mean, that precisely means that we throw away data. And here we do so rather thoroughly. I think even in the rather rare scenario that the CPython interpreter crashes with a Segmentation fault then even _that_ would be dumped into /dev/null. Mhhhhm... at least I'd appreciate some discussion around that!

Can't we build a context manager that buffers and that emits upon leaving the context?

dcos/auth.py
bamarni: I didn't change this on purpose, reverting.

dcos/util.py
bamarni: I don't think this is needed in the specific case we're trying to solve (output from a subprocess), but this was inspired by https://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/, where they say this is needed in case we're running C code (eg. a python module implemented as a C extension)  :
 
```With this information in hand, it should be easy to understand what stdout actually is for a C program. stdout is a global FILE object kept for us by the C library, and it buffers output to file descriptor number 1. Calls to functions like printf and puts add data into this buffer. fflush forces its flushing to the file descriptor, and so on.```

So even though we flush `sys.stdout`, there might still be some buffered output from C code.

File: dcos/util.py (hunk 3 above)
Comment (bamarni) [reply dcos/util.py]: Even though we buffer this data, in case of a hard failure this wouldn't be printed either I guess. Also in my specific case, it is already prompting what it is going to try : 
```
If your browser didn't open, please go to the following link : (url)
```
Then it successfully opens a tab, but also prints warning messages. I don't think they should be printed back to the user, even after a delay, this can be confusing as things actually worked.

**Comment by bamarni** [comment]
@jgehrcke : thanks for the review, I'd like to investigate further to see if it can be simplified, and possibly adding tests for this, setting as wip for now.

**Comment by bamarni** [comment]
Also for reference, there are some built-in context managers for this : https://docs.python.org/3.4/library/contextlib.html#contextlib.redirect_stdout

And the same for stderr, however they appeared not to work because it's only for output from python and in the same process, that's why I had to come-up with something more complicated.

**Comment by bamarni** [comment]
I've simplified this a bit, by not touching Python's `sys.stdout` and `sys.stderr` but only swapping at file descriptors level. Unfortunately this appeared to be tricky to unit test as python capture system is doing something similar, while trying to test I was getting an error, kind of stepping on each other toes with pytest...

screencast

As for the C flushing, should I remove it? This makes the output blocking more bullet-proof, but as I said I don't think we have this use-case at the moment.

**Comment by bamarni** [comment]
run linux integration tests

**Comment by bamarni** [comment]
run windows integration tests

File: dcos/auth.py [hunk 4]
```diff
@@ -67,7 +66,8 @@ def _prompt_user_for_token(url, token_type):
     msg = msg.lstrip().format(url=url, token_type=token_type)
 
     try:
-        webbrowser.open_new_tab(url)
+        with util.silent_outputs():
```
klueska: s/silent_outputs/silent_output/

dcos/util.py
```diff
@@ -98,6 +99,52 @@ def remove_path_on_error(path):
         raise
 
 
+@contextlib.contextmanager
+def silent_outputs():
```
klueska: s/silent_outputs/silent_output/

dcos/util.py
klueska: @jgehrcke If an exception is thrown, the output of that will be printed as always (because the finally block restores the original stdout/stderr, which is back in place by the time the raised exception gets printed.

In the case of a segmentation fault, I don't see how buffering would help. A hard crash is a hard crash and the program would have no opportunity to restore the context and print the output anyway.

The only way to handle that would be to handle the OS signal from one of these extraordinary events and restore the IO redirection from there. I'm not sure it is worth the effort for such a large corner case.

dcos/util.py
```diff
@@ -98,6 +99,52 @@ def remove_path_on_error(path):
...
+
+    libc = ctypes.CDLL(None)
+    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
+    c_stderr = ctypes.c_void_p.in_dll(libc, 'stderr')
+
+    # Original fds stdout/stderr point to. Usually 1 and 2 on POSIX systems.
+    original_stdout_fd = sys.stdout.fileno()
+    original_stderr_fd = sys.stderr.fileno()
+
+    def _redirect_outputs(stdout_to_fd, stderr_to_fd):
```
klueska: Can we move this to line 137?

dcos/util.py
jgehrcke: > In the case of a segmentation fault, I don't see how buffering would help. A hard crash is a hard crash and the program would have no opportunity to restore the context and print the output anyway.

That is correct. With the following statement:

> I think even in the rather rare scenario that the CPython interpreter crashes with a Segmentation fault then even that would be dumped into /dev/null

I tried to say that me maybe should just bend Python's I/O system within the context, and not completely bend the underlying file descriptors. In that case we'd still see the bytes directly emitted to the standard streams. When I thought about this I was assuming that the annoying warnings in questions are in fact emitted through Python's I/O system. I looked into the issue again and now see that that assumption is probably wrong: what we want to suppress is probably _not_ emitted through Python's I/O system, but by code that directly interacts with the OS through the OS file descriptors.

Is that correct?

That is, did we confirm that "just" manipulating `sys.stdout` and `sys.stderr` is not sufficient for suppressing the messages in question?

Thinking about this on that level I suppose a cleaner way to deal with this entire issue is to spawn a lean an clean child process via Python's `subprocess()` module. That process' sole purpose would be to use the `webbrowser` module. The trick would be to make it _not_ inherit the parent's standard streams, but to instead point its standard streams to pipes connected to the parent. And to then read its stdout and stderr streams asynchronously in the parent from a thread while prompting for the auth token. _After_ the prompt returns, the CLI process holds the complete data emitted by the browser-controlling process.

Just an idea for what I think would be a more solid architecture. I find it rather intrusive to redirect all stdout and stderr to /dev/null. If that _ever_ swallows helpful debugging information we will regret that.

jgehrcke:
I understand if we decide against it, but I have expressed my major concern and a proposal for what I believe is a cleaner solution via https://github.com/dcos/dcos-cli/pull/1072#discussion_r142967074.

dcos/util.py
bamarni: @jgehrcke : do you mean spawning a process that would invoke the python executable with inline code? If so I don't think this is an option, as we're shipping the dcos-cli as a self-contained binary which doesn't require a python executable to be present.

By cleaner do you mean that this would lead to less code? Because I find the current approach to be cleaner in some other sense, as it doesn't need to go through the additional cost of spawning another process in-between.

jgehrcke: > By cleaner do you mean that this would lead to less code?

No, it's all about

> I find it rather intrusive to redirect all stdout and stderr to /dev/null. If that ever swallows helpful debugging information we will regret that.

I was trying to suggest a way to that helps us retaining the data.

klueska:
@jgehrcke I think we are going to go with @bamarni's current approach (for the time being) as the goal was to explicitly suppress any and all output from the `webbrowser.open_new_tab(url)` command, that is what this helper function will be used for. If we (in the future) want to write a different helper which captures input (as opposed to suppressing it), we can reconsider it then. Such a function might be useful if we want to look at the output and only suppress it if it matches a certain pattern (which is not the case here).

bamarni:
run all integration tests

bamarni:
I've updated the PR and triggered tests, this one just misses an approval.

PR #1073, opened 2017-10-06
Refactor Jenkinsfiles

bamarni:
This extracts common code between Jenkinsfiles into a dedicated base file.


bamarni:
run linux integration tests

bamarni:
run linux integration tests

klueska:
I don't think this will work unless you do something like:
https://stackoverflow.com/questions/37800195/how-do-you-load-a-groovy-file-and-execute-it

Also, "Factorize" is not a word in english, so you mean "Refactor"?

bamarni:
> Also, "Factorize" is not a word in english, do you mean "Refactor"?

Yes :smile_cat: 

I've tried the other approach with the last commit, however I think it doesn't work outside from a node block.

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
At this point I'd rather do this during the 1.12 refactoring.

PR #1075, opened 2017-10-09
Update prompt for confirmation when uninstalling CLI only

bamarni:
By default the following message appears :

> WARNING: This action cannot be undone. This will uninstall [dcos-enterprise-cli] and delete all of its persistent data (logs, configurations, database artifacts, everything).

This is not accurate when uninstalling a CLI only package implicitly (without specifying the --cli flag).

https://jira.mesosphere.com/browse/DCOS-19010

### DIFF
--- a/cli/dcoscli/package/main.py
+++ b/cli/dcoscli/package/main.py
@@ -499,27 +499,33 @@ def _uninstall(package_name, remove_all, app_id, cli, app, skip_confirmation):
     :rtype: int
     """
 
+    package_manager = get_package_manager()
+
     # Don't gate CLI uninstalls.
     # Don't gate uninstalls if the user wants to skip confirmation.
-    if not cli and not skip_confirmation and not _confirm_uninstall(
-        package_name,
-        remove_all,
-        app_id
-    ):
-        return 0
+    if not cli and not skip_confirmation:
+        installed = package.installed_packages(
+            package_manager, None, package_name, cli_only=False)
+        installed_pkg = next(iter(installed), None)
+        if not installed_pkg or not installed_pkg.get('name'):
+            msg = "Package '{}' is not installed.".format(package_name)
+            emitter.publish(msg)
+            return 1
 
-    package_manager = get_package_manager()
+        if not _confirm_uninstall(installed_pkg, remove_all, app_id):
+            return 0
+
     package.uninstall(
         package_manager, package_name, remove_all, app_id, cli, app)
 
     return 0
 
 
-def _confirm_uninstall(package_name, remove_all, app_id):
+def _confirm_uninstall(installed_pkg, remove_all, app_id):
     """Interactively confirm the uninstall with the user.
 
-    :param package_name: The package to uninstall
-    :type package_name: str
+    :param installed_pkg: The currently installed package
+    :type installed_pkg: str
     :param remove_all: Whether to remove all instances of the named package
     :type remove_all: boolean
     :param app_id: App ID of the package instance to uninstall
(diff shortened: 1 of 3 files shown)


bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run mac integration tests

bamarni:
run linux integration tests

cli/dcoscli/package/main.py
```diff
@@ -499,16 +499,20 @@ def _uninstall(package_name, remove_all, app_id, cli, app, skip_confirmation):
...
-        remove_all,
-        app_id
-    ):
-        return 0
+    if not cli and not skip_confirmation:
+        installed = package.installed_packages(
+            package_manager, None, package_name, cli_only=False)
+        installed_pkg = next(iter(installed), None)
+
+        # Don't gate implicit CLI uninstalls.
```
klueska: I think we should still gate them, but with a different prompt. Basically informing them that this CLI will be removed for all instances of interacting with an App (i.e. in the case of installing say the Kafka CLI) and forcing them to confirm with a Yes/No.

jsrodman: As an experienced cli user, I don't want a yes/no confirm behavior for uninstalling cli packages, but I may not represent most users.

I think for `dcos package uninstall foo --cli` you would never want this behavior.  However, for `dcos package uninstall foo` I'm less sure (which I guess is this case.)  It's also hard to describe this accurately: "you'll lose this cli package for all services it might relate to, even though it might not be a service-related extension, and also it will only be lost for this cluster and not other attached clusters."

------

I also hope someone smarter than me knows this boolean construction can't engage for `dcos package uninstall foo` that targets cli and an app both.

cli/tests/integrations/test_package.py
```diff
@@ -401,6 +401,25 @@ def test_uninstall_cli_only_when_no_apps_remain():
...
+            b'catalog-terms-conditions/#community-services\n'
+            b'Installing CLI subcommand for package [helloworld] '
+            b'version [0.1.0]\n'
+            b'New command available: dcos ' +
+            _executable_name(b'helloworld') +
+            b'\n'
+        )
+
+        with _package(name='helloworld',
+                      args=['--yes', '--cli'],
```
jsrodman: I'm confused by --cli being present in a test for "implicitly".

bamarni: This flag is for installation command only, the implicitness is about uninstallation (where we don't specify `--cli`).

cli/dcoscli/package/main.py
bamarni: I'll use a different prompt message, to bypass it there still would be the `--yes` option.

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

cli/dcoscli/package/main.py
```diff
@@ -499,27 +499,33 @@ def _uninstall(package_name, remove_all, app_id, cli, app, skip_confirmation):
...
-    ):
-        return 0
+    if not cli and not skip_confirmation:
+        installed = package.installed_packages(
+            package_manager, None, package_name, cli_only=False)
+        installed_pkg = next(iter(installed), None)
+        if not installed_pkg or not installed_pkg.get('name'):
+            msg = "Package '{}' is not installed.".format(package_name)
+            emitter.publish(msg)
+            return 1
```
elezar: Could we push this case down into `_confirm_uninstall()`?

Also, what is the intended behaviour if the package doesn't exist and we call `_uninstall()` with `cli=True` or `skip_confirmation=True`?

bamarni: The reason I've left it outside is because a missing package would result in an exit code error while a non confirmation would just exit the program successfully, so I would have to adapt the function with these new semantics (and also add the `package_name` parameter needed for the "Package not installed" error message), which felt too much to me.

If the confirmation is skipped and the package doesn't exist, there is an error returned by the underlying API (I think it's Cosmos)  that would be printed, eg `Package [foo] is not installed`. This was actually one of the things mentioned in the JIRA issue, we are currently asking confirmation even for package which doesn't exist, before failing saying they don't exist..

bamarni:
@elezar : about tests, I've added one for `_confirm_uninstall` when a CLI-only package is installed and it checks the boolean return value. Or did you mean testing the exact message being prompted?

bamarni:
@klueska  @elezar  : any news about this one?

elezar:
With regards to 

>@elezar : about tests, I've added one for _confirm_uninstall when a CLI-only package is installed and it checks the boolean return value. Or did you mean testing the exact message being prompted?

I was thinking of the prompt (or whether the user is prompted) since this seems to be what caused this PR to be created. If I am not correct in my understanding, then we can merge this as is.

PR #1076, opened 2017-10-12
Don't use custom CA if HTTPS works out of the box

bamarni:
If it works there is no need to download and use the custom CA. This usually happens when we're dealing with a load-balancer or when a custom CA has already been added to the system.

With the Python library we're using (requests), specifying a path to a custom CA would ignore the default CA bundle in the system, so setting-up a cluster through a load balancer endpoint with a well-known CA (eg. Digicert) was previously failing without this change.

https://jira.mesosphere.com/browse/DCOS-19040

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -1,6 +1,7 @@
 import os
 
 import docopt
+import requests
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
@@ -229,13 +230,14 @@ def setup(dcos_url,
             config.set_val("core.ssl_verify", "false")
         elif ca_certs:
             config.set_val("core.ssl_verify", ca_certs)
-        else:
+        elif _needs_cluster_cert(dcos_url):
             cert = cluster.get_cluster_cert(dcos_url)
-            # if we don't have a cert don't try to verify one
-            if cert is False:
-                config.set_val("core.ssl_verify", "false")
-            else:
+            if cert:
                 stored_cert = _store_cluster_cert(cert, no_check)
+            else:
+                config.set_val("core.ssl_verify", "false")
+        else:
+            config.set_val("core.ssl_verify", "false")
 
         try:
             login(dcos_url,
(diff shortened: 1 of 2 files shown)


bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -230,12 +231,26 @@ def setup(dcos_url,
...
+                    requests.get(dcos_url)
+                    download_custom_ca = False
+                    config.set_val("core.ssl_verify", "true")
+                except requests.exceptions.SSLError:
+                    pass
+
+            if download_custom_ca:
+                cert = cluster.get_cluster_cert(dcos_url)
+                # if we don't have a cert don't try to verify one
+                if cert is False:
```
mnaboka: reverse the if statement
```python
if cert:
  stored_cert = _store_cluster_cert(cert, no_check)
else:
  config.set_val("core.ssl_verify", "false")
```

cli/dcoscli/cluster/main.py
```diff
@@ -230,12 +231,26 @@ def setup(dcos_url,
...
+            # If it works there is no need to download the custom CA.
+            # This usually happens when we're dealing with a load-balancer
+            # or when a custom CA has already been added to the system.
+            if dcos_url.startswith('https://'):
+                try:
+                    requests.get(dcos_url)
+                    download_custom_ca = False
+                    config.set_val("core.ssl_verify", "true")
+                except requests.exceptions.SSLError:
+                    pass
```
mnaboka: maybe wrap this up as a separate function?

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
`test_debug_details` failed again (should be fixed with #1078), retrying...

bamarni:
@mnaboka : I've addressed your comments and added unit tests.

PR #1078, opened 2017-10-13
 Fix Marathon Debug Test

kensipe:
fixing marathon_debug tests.  
reduce integration test time and tests retry assuming eventually true expectations.  
fixed regex to not be dependent on cluster sizes.  
added retrying lib

### DIFF
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -3,5 +3,6 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
+retrying
 PyInstaller==3.2.1
 -e .. # Install the DC/OS package
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -3,6 +3,5 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
-retrying
 PyInstaller==3.2.1
 -e .. # Install the DC/OS package
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -4,4 +4,5 @@ wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
 PyInstaller==3.2.1
+retrying==1.3.3
 -e .. # Install the DC/OS package
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ setup(
         'six>=1.9, <2.0',
         'toml>=0.9, <1.0',
         'sseclient==0.0.14',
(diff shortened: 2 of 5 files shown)


kensipe:
run linux integration tests

cli/requirements.txt
```diff
@@ -3,5 +3,6 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
+retrying
```
bamarni: Could you either specify the exact version or a range so that it sticks to the current major version? This would make sure that in the future we won't jump unexpectedly to the next major version and get breakages.

setup.py
```diff
@@ -69,6 +69,7 @@
         'six>=1.9, <2.0',
         'toml>=0.9, <1.0',
         'sseclient==0.0.14',
+        'retrying',
```
bamarni: We have two `setup.py` and `requirements.txt` pair of files (one for each package, dcos and dcoscli), this one is for dcos, where that dependency is not used. I think you want to revert this line.

kensipe: thank you sir!  will do!

kensipe:
@bamarni thanks!   feedback resolved!

kensipe:
run linux integration tests

setup.py
```diff
@@ -69,6 +69,7 @@
         'six>=1.9, <2.0',
         'toml>=0.9, <1.0',
         'sseclient==0.0.14',
+        'retrying==1.3.3',
```
bamarni: I meant the opposite. Currently this project defines two packages : [dcos](https://pypi.python.org/pypi/dcos) (under `./dcos`) and [dcoscli](https://pypi.python.org/pypi/dcoscli) (under `./cli`), however that dependency is introduced for the `dcoscli` package, so the change in `cli/requirements.txt` was correct and this one should be reverted.

The reason it also works from here is because `dcoscli` requires `dcos` which would bring that dependency along.

kensipe:
run linux integration tests

kensipe:
run linux integration tests

kensipe:
@bamarni it's go time!

cli/requirements.txt
```diff
@@ -4,4 +4,5 @@ wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
 PyInstaller==3.2.1
+retrying==1.3.3
```
bamarni: For the record, this requirements.txt file seems ignored so we had to add the dependency in setup.py, I'll create a ticket about this.

bamarni: see https://jira.mesosphere.com/browse/DCOS_OSS-1762?filter=-1

PR #1081, opened 2017-10-17
Cluster search with fewer chars

kensipe:
* making attaching clusters much easier (as it should be!)

typically a cluster name is something like `ken-lm1juwj` and the cluster-id is `1c0c189e-650c-4caa-a759-4e4d395e4d17`.   These are challenging to type leading mostly to copy and paste when you want to attach to a different cluster.   It is common when using tools like git or docker to use a shorthand for the sha.  Usually requiring the minimum of characters to uniquely match the sha.   This change is designed a user to type `dcos cluster attach ken-l` assuming there is only one, it will attach.

### DIFF
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -221,8 +221,16 @@ def get_cluster(name):
     :rtype: Cluster
     """
 
-    return next((c for c in get_clusters()
-                 if c.get_cluster_id() == name or c.get_name() == name), None)
+    clusters = [c for c in get_clusters()
+                if c.get_cluster_id().startswith(name)
+                or c.get_name().startswith(name)]
+
+    if len(clusters) == 0:
+        return None
+    elif len(clusters) == 1:
+        return clusters[0]
+    else:
+        raise DCOSException("Multiple clusters of [{}] exist".format(name))
 
 
 def remove(name):


kensipe:
run linux integration tests

bamarni:
> making attaching clusters much easier (as it should be!)

Could you please add more information into the description? It is not clear to me what problem you're trying to address or what improvement you suggest.

kensipe:
@bamarni updated desc

bamarni:
This seems to break some tests, could you also add tests for this? There is a new case with the exception being thrown for eg.

bamarni:
Also, I don't have a strong opinion on this, but do we want to support this on both the name and the id? Trying with the `docker rm` command, it only allows this for ids.

kensipe:
yeah... cluster names and ids are strange for the CLI IMO.    If you have a preference or opinion... I'm open.   I leveraged what was there.   I resolve the integration test soonish.

bamarni:
@kensipe : Sorry for the delay, here I think cluster id only would be good.

bamarni:
Thanks for the great idea @kensipe! As this is hanging for a while, I've opened #1148 and added tests for this.

PR #1082, opened 2017-10-18
Don't flush C buffers in the silent_output contextmanager

bamarni:
Symbol names for outputs might differ from one OS to the other. While this works on Linux this was failing on a Macbook Pro, even though it passed on our Mac CI somehow.

As we don't really need this (it was made to make the silencing more robust), we can just remove it.

https://jira.mesosphere.com/browse/DCOS_OSS-1760

### DIFF
--- a/dcos/util.py
+++ b/dcos/util.py
@@ -1,7 +1,6 @@
 import collections
 import concurrent.futures
 import contextlib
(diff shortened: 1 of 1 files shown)


bamarni:
run all integration tests

PR #1084, opened 2017-10-19
Remove the hardcoded 1 second timeout for some HTTP calls

bamarni:
This might be too short in case of network latency or if for any reason
the cluster is not very responsive but still works fine. In such cases
commands like "cluster setup" or "cluster --version" might fail.

There is already a 5 seconds timeout by default that we can rely on.

https://jira.mesosphere.com/browse/DCOS_OSS-1763

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -321,5 +321,10 @@ def _needs_cluster_cert(dcos_url):
             requests.get(dcos_url, timeout=http.DEFAULT_TIMEOUT)
         except requests.exceptions.SSLError:
             return True
+        except Exception as e:
+            logger.warning(
+                'Unexpected exception occurred while calling %s: %s',
+                dcos_url, e)
+            return True
 
     return False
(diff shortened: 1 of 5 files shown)


bamarni:
run all integration tests

bamarni:
run all integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -321,5 +321,10 @@ def _needs_cluster_cert(dcos_url):
             requests.get(dcos_url, timeout=http.DEFAULT_TIMEOUT)
         except requests.exceptions.SSLError:
             return True
+        except Exception as e:
+            logger.warning(
+                'Unexpected exception occurred while calling %s: %s',
+                dcos_url, e)
+            return True
```
elezar: It seems a bit strange to have `True` be the default error case, but maybe I'm missing something here.

bamarni: `_needs_cluster_cert` makes some sort of pre-setup request to check if https works out of the box, but I had miss the case where there is an even harder failure (eg. route not found or similar), in such a situation it is currently failing with a stack trace, with this change it would fail properly later on as shown by `test_setup_unreachable_url`.

PR #1085, opened 2017-10-20
Remove pydoc folder

bamarni:
This was a skeleton for the documentation of the packages we publish, but as we don't want to publish them anymore this can be removed.


bamarni:
run linux integration tests

PR #1091, opened 2017-11-01
Allow Python 3.6 in build scripts

bamarni:
This allows to run make commands with Python 3.6 by exporting PYTHON=python3.6 environment variable for example.

CI and actual releases would still rely on Python 3.5, however that'll let us play with it locally and start considering to bump the version in the future.

### DIFF
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -11,11 +11,13 @@ if [ ! -d "${BUILDDIR}/${VENV}" ]; then
       exit 1
     fi
 
+    # If no PYTHON variable is set, we assume python3.5 on Linux and MacOS (cf. common.sh).
+    # However, more recent versions of Python 3 are also supported.
     PYTHON_MAJOR=$(${PYTHON} -c 'import sys; print(sys.version_info[0])')
     PYTHON_MINOR=$(${PYTHON} -c 'import sys; print(sys.version_info[1])')
 
-    if [ "${PYTHON_MAJOR}" != "3" ] || [ "${PYTHON_MINOR}" != "5" ]; then
-        echo "Cannot find supported python version 3.5. Exiting..."
+    if [ "${PYTHON_MAJOR}" != "3" ] || [ "${PYTHON_MINOR}" \< "5" ]; then
+        echo "Cannot find supported python version +3.5. Exiting..."
         exit 1
     fi
     if [ "$(uname)" = "Windows_NT" ]; then
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -3,6 +3,6 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
-PyInstaller==3.2.1
+PyInstaller==3.3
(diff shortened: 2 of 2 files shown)


bin/env.sh
```diff
@@ -14,7 +14,7 @@ if [ ! -d "${BUILDDIR}/${VENV}" ]; then
     PYTHON_MAJOR=$(${PYTHON} -c 'import sys; print(sys.version_info[0])')
     PYTHON_MINOR=$(${PYTHON} -c 'import sys; print(sys.version_info[1])')
 
-    if [ "${PYTHON_MAJOR}" != "3" ] || [ "${PYTHON_MINOR}" != "5" ]; then
+    if [ "${PYTHON_MAJOR}" != "3" ] || [[ "${PYTHON_MINOR}" < "5" ]]; then
```
klueska: I wouldn't just make this change directly and commit it as is. This would allow anyone who's default python installation is 3.6 to unwittingly pass this check. We should only allow python 3.6 in conjunction with some sort of DCOS_CLI_EXPERIMENTAL variable being set.

bamarni: Good point, this syntax was also failing on winbash apparently. I'll look into it.

bamarni:
@klueska : I've tried to make this check a pure python multiline script but the heredoc didn't work on windows (https://stackoverflow.com/questions/17467441/how-to-embed-python-code-in-batch-script).

The `>=` operator also didn't seem to work, as I didn't find a really elegant solution I went ahead with duplicating the block, let me know if you have another suggestion.

bamarni:
Also, on linux / mac we currently assume `python3.5` if `PYTHON` is not explicitly passed (https://github.com/dcos/dcos-cli/blob/master/bin/common.sh#L24), does this exist on windows too? That'd remove the need of an experimental variable.

bamarni:
run linux integration tests

bamarni:
Ci failed at dcos-launch : 
```
[linux-tests]              'Docker Server Version: 1.10.3\r',
[linux-tests]              'Docker version 1.10 not supported. Please uninstall Docker and '
[linux-tests]              'try again.\r',
```

I'll run it again but it seems like some requirements changed.

bamarni:
run linux integration tests

klueska:
OK, I think I'm OK with this. We should add a comment in `env.sh` though that we assume python3.5 by default if no PYTHON variable is set (and point them to `bin/common.sh`).

cli/requirements.txt
```diff
@@ -3,6 +3,6 @@ tox>=2.2, <3.0
 wheel>=0.24.0, <1.0
 pytest>=2.9.1
 teamcity-messages>=1.20
-PyInstaller==3.2.1
+PyInstaller==3.3
```
bamarni: I had to bump this as otherwise binaries couldn't be generated with Python 3.6 (failing with an error).

bamarni:
run all integration tests

bamarni:
I don't think this will be needed, during the transition period the new cli will be in a subfolder with its own build scripts and makefile.

bamarni:
run all integration tests

PR #1092, opened 2017-11-02
include region/zone in task subcommand with --json option

mnaboka:
(DCOS-19379)[https://jira.mesosphere.com/browse/DCOS-19379]

with fault domain enabled
screen shot 2017-11-02 at 11 57 32 am

if fault domains are disabled, no additional fields will be added.

### DIFF
--- a/dcos/mesos.py
+++ b/dcos/mesos.py
@@ -664,6 +664,17 @@ class Slave(object):
                  for framework in self._framework_dicts()]
         return itertools.chain(*iters)
 
+    def fault_domain(self):
+        """ Fault domain for a given task e.g.
+
+        :returns: fault domain structure from state.json
+        :rtype: dict | None
+        """
+        if 'domain' not in self._short_state:
+            return None
+
+        return {'domain': self._short_state['domain']}
+
     def __getitem__(self, name):
         """Support the slave[attr] syntax
 
@@ -748,6 +759,13 @@ class Task(object):
         self._task = task
         self._master = master
 
+        if 'slave_id' not in self._task or not master:
+            return
+
+        fd = self.fault_domain()
+        if fd:
+            self._task.update(fd)
+
     def dict(self):
         """
         :returns: dictionary representation of this Task
@@ -806,6 +824,14 @@ class Task(object):
         """
 
         return self.executor()['directory']
+
+    def fault_domain(self):
+        """ Fault domain for a given task
+
+        :returns: fault domain structure
+        :rtype: dict | None
+        """
+        return self.slave().fault_domain()
 
     def __getitem__(self, name):
         """Support the task[attr] syntax
--- a/cli/tests/unit/test_task.py
+++ b/cli/tests/unit/test_task.py
@@ -194,3 +194,26 @@ def test_dcos_task_metrics_agent_missing_slave(mocked_get_config_val,
     # Should a task not have a slave ID, expect an error
     with pytest.raises(DCOSException):
         _metrics(True, 'task_id', False)
+
+
+def test_task_fault_domain():
+    fd = {
+        'domain': {
+            'fault_domain': {
+                'region': {
+                    'name': 'us-west-2'
+                },
+                'zone': {
+                    'name': 'us-west-2a'
+                }
+            }
+        }
+    }
+
+    state = {'prop1': 'value1'}
+    state.update(fd)
+    slave = mesos.Slave(state, None, None)
+    assert slave.fault_domain() == fd
+
+    slave = mesos.Slave({}, None, None)
+    assert not slave.fault_domain()


mnaboka:
@klueska we can do either way, remove it completely or set to None.
As for levels of nesting, so basically i am making it like this to be consistent with mesos. Mesos state json has fault domain exposed like this, also `dcos node --json` exposes fault domains in the same format. Assuming client might want to implement parking function once and reuse it.

screen shot 2017-11-02 at 12 27 38 pm

screen shot 2017-11-02 at 12 28 13 pm

klueska:
In both cases we should mimic what mesos state does. If it has multiple nesting levels we should too. If it sets it to null, we should too, etc.

mnaboka:
@klueska that makes perfect sense. I will remove fault domains at all if it is disabled.

mnaboka:
@klueska updated

mnaboka:
run linux integration tests

mnaboka:
run linux integration tests

dcos/mesos.py
```diff
@@ -754,7 +766,15 @@ def dict(self):
         :rtype: dict
         """
 
-        return self._task
+        try:
+            fd = self.fault_domain()
+        except Exception:
+            return self._task
+        if not fd:
+            return self._task
```
klueska: Why make a copy instead of updating the underlying dictionary? And pull the info directly from `self.slave().fault_domain()` (with the `fault_domain()` function for this class just returning `self["domain"]`.

mnaboka: my idea was to no mutate the original `self._task` object. However when the user calls `.dict()` a new copy is being made with additional `domain` field injected. assuming the original `self._task` object might be used by other functions etc. Let me know if you think this does not make sense and we better do it in a different way.

mnaboka:
run linux integration tests

dcos/mesos.py
```diff
@@ -754,7 +766,15 @@ def dict(self):
         :rtype: dict
         """
 
-        return self._task
+        try:
+            fd = self.fault_domain()
+        except Exception:
+            return self._task
```
bamarni: What exception could be returned here, is it a `KeyError`? I'd rather catch the exact exception inside `fault_domain()` and return `None` in such cases.

dcos/mesos.py
klueska: I think the copy should happen in the `__init__` function and then we are free to modify it ourselves however we please from within this class.

dcos/mesos.py
klueska: We shouldn't put try/except blocks around simple KeyError exceptions anyway. We should explicitly check for the existence of the key and return e.g. `None` if appropriate.

dcos/mesos.py
bamarni: I couldn't find places where Task objects are instanciated other than the Framework class in this file, which is also used by the Master class.

I think it's safe to mutate the dict by adding this new key for fault domains.

bamarni: I'm fine with @klueska suggestion too.

mnaboka:
run linux integration tests

mnaboka:
run linux integration tests

mnaboka:
run linux integration tests

dcos/mesos.py
```diff
@@ -748,6 +759,12 @@ def __init__(self, task, master):
         self._task = task
         self._master = master
 
+        if 'slave_id' not in self._task or not master:
+            return
+
+        fd = self.fault_domain()
+        self._task.update(fd)
```
bamarni: CI is currently failing, I think you need to add a check for `not None` before this line : 
```
[linux-tests] CMD: ['dcos', 'service', 'log', 'cassandra']
[linux-tests] STDOUT: 
[linux-tests] STDERR: Traceback (most recent call last):
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/subcommand.py", line 101, in run_and_capture
[linux-tests]     exit_code = m.main([self._command] + self._args)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/service/main.py", line 19, in main
[linux-tests]     return _main(argv)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/util.py", line 24, in wrapper
[linux-tests]     result = func(*args, **kwargs)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/service/main.py", line 32, in _main
[linux-tests]     return cmds.execute(_cmds(), args)
[linux-tests]   File "/workspace/dcos-cli/dcos/cmds.py", line 43, in execute
[linux-tests]     return function(*params)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/service/main.py", line 152, in _log
[linux-tests]     return _log_service(follow, lines, service, file_)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/service/main.py", line 178, in _log_service
[linux-tests]     return _log_task(task['id'], follow, lines, file_)
[linux-tests]   File "/workspace/dcos-cli/cli/.tox/py35-integration/lib/python3.5/site-packages/dcoscli/service/main.py", line 204, in _log_task
[linux-tests]     task = mesos.get_master(dcos_client).task(task_id)
[linux-tests]   File "/workspace/dcos-cli/dcos/mesos.py", line 399, in task
[linux-tests]     tasks = self.tasks(fltr, completed)
[linux-tests]   File "/workspace/dcos-cli/dcos/mesos.py", line 471, in tasks
[linux-tests]     task = self._framework_obj(framework).task(task['id'])
[linux-tests]   File "/workspace/dcos-cli/dcos/mesos.py", line 716, in task
[linux-tests]     return self._task_obj(task)
[linux-tests]   File "/workspace/dcos-cli/dcos/mesos.py", line 730, in _task_obj
[linux-tests]     self._tasks[task['id']] = Task(task, self._master)
[linux-tests]   File "/workspace/dcos-cli/dcos/mesos.py", line 766, in __init__
[linux-tests]     self._task.update(fd)
[linux-tests] TypeError: 'NoneType' object is not iterable
```

mnaboka:
run linux integration tests

mnaboka:
run linux integration tests

mnaboka:
@bamarni tests passed

bamarni:
Thanks @mnaboka

PR #1096, opened 2017-11-07
Show executor metrics

philipnrmn:
This PR fixes [DCOS-19009](https://jira.mesosphere.com/browse/DCOS-19009)

Running `dcos task metrics details <task-id>` does the following:

1. requests container metrics from `.../metrics/v0/containers/<container-id>`
1. requests app metrics from `.../metrics/v0/containers/<container-id>/app`
1. joins the two and prints them out in a table

In some cases, container metrics are not present and the dcos-metrics API returns a 204 empty. We were assuming in those cases that there were also no app metrics. Unfortunately, that is not correct in cases of executors, as in DC/OS SDK frameworks like cassandra and kafka. 

With this PR, we catch the empty case and poll app metrics as well.

### DIFF
--- a/cli/dcoscli/metrics.py
+++ b/cli/dcoscli/metrics.py
@@ -263,8 +263,18 @@ def print_task_metrics(url, app_url, summary, json_):
     :rtype: int
     """
 
-    datapoints = _fetch_metrics_datapoints(url) + _fetch_metrics_datapoints(
-        app_url)
+    container_datapoints = []
+    app_datapoints = []
+
+    # In the case of an executor, app data may exist when
+    # container data does not.
+    try:
+        container_datapoints = _fetch_metrics_datapoints(url)
+    except EmptyMetricsException:
+        pass
+
+    app_datapoints = _fetch_metrics_datapoints(app_url)
+    datapoints = container_datapoints + app_datapoints
 
     if summary:
         if json_:
--- a/cli/dcoscli/metrics.py
+++ b/cli/dcoscli/metrics.py
@@ -9,6 +9,11 @@ logger = util.get_logger(__name__)
 emitter = emitting.FlatEmitter()
 
 
+class EmptyMetricsException(DCOSException):
+    def __init__(self):
+        super().__init__('No metrics found')
+
+
 def _gib(n):
     return n * pow(2, -30)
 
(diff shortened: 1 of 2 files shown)


cli/dcoscli/metrics.py
```diff
@@ -9,6 +9,11 @@
 emitter = emitting.FlatEmitter()
 
 
+class EmptyMetricsException(DCOSException):
+    def __init__(self):
+        DCOSException.__init__(self, 'No metrics found')
```
bamarni: As we don't have to support Python 2, could you switch this to `super().__init__( 'No metrics found')`?

cli/dcoscli/metrics.py
```diff
@@ -258,8 +263,18 @@ def print_task_metrics(url, app_url, summary, json_):
...
+    app_datapoints = []
+
+    # In the case of an executor, app data may exist when
+    # container data does not.
+    try:
+        container_datapoints = _fetch_metrics_datapoints(url)
+    except EmptyMetricsException:
+        pass
+
+    app_datapoints = _fetch_metrics_datapoints(app_url)
```
bamarni: Is it possible that we have the other way around, container data exists but not app data?

cli/dcoscli/metrics.py
philipnrmn: 👍

cli/dcoscli/metrics.py
philipnrmn: No, there will always be app data because we always pass two app values from the mesos module:

https://github.com/dcos/dcos-metrics/blob/3c96b31b746170981211329d35c96da3eda5843b/mesos_module/container_reader_impl.cpp#L153-L158

bamarni:
run linux integration tests

bamarni:
@philipnrmn : there are a few code style errors : 
```
[linux-tests] tests/unit/test_task.py:182:42: E128 continuation line under-indented for visual indent
[linux-tests] tests/unit/test_task.py:183:42: E128 continuation line under-indented for visual indent
[linux-tests] tests/unit/test_task.py:202:1: E302 expected 2 blank lines, found 1
```

You can check these locally by running `tox -e py35-syntax` in the `cli` directory (after running `make env` and `source env/bin/activate`).

I just noticed btw that CI continues and runs tests even though it's going to fail eventually because of syntax errors... I'll look into it.

bamarni:
run linux integration tests

Issue #595, opened 2016-04-28
Make DC/OS CLI Multi Cluster aware


lloesche:
Often times I have more than one DC/OS cluster. I might have separate staging and production systems. It would be nice if instead of having to configure several dcos cli's I could have a single one with multiple DC/OS backends configured and then switch between them like e.g.
`dcos marathon app list --cluster=prod`
`dcos marathon app list --cluster=stage`

Together with another function that I'm going to create a separate issue (https://github.com/dcos/dcos-cli/issues/596) for I could then do things like `dcos marathon backup --cluster=prod | dcos marathon restore --cluster=stage`

tamarrow:
Yes, this is a commonly requested feature. As a workaround for now, you can have multiple config files, each for a different cluster, and use the `DCOS_CONFIG` env variable to override what cluster you want to use. Something like `DCOS_CONFIG=~/.dcos/dcos-dev.toml marathon app list`

arun-gupta:
This would be very useful as multiple clusters would be a common scenario. The current way to configure `dcos` CLI is a three step process:

- `dcos config set core.dcos_url <new-dcos-url>`
- OAuth token
- `dcos auth login`

It needs to be simplified.

prein:
I'm using the workaround as suggested by @tamarrow 
However I'm hitting strange issue when running dcos cli, in docker, from gitlab ci runner (I assume it cleans up after each run).
There are 3 toml configs. I set DCOS_CONFIG to the one I want to use. Then I print 'dcos config show'.  It shows the correct url, but then if as very next command I run any dcos cli command requiring auth, it fails for the reason that it's using another cluster url, from other toml file. I'm so confused to see that happening. 

I run it with strace and it confirmed that it's opening only the correct toml file. No other toml file is touched. And no, I don't have default dcos.toml. Then the wrong url comes out of the blue. So confusing. I must be doing some simple mistake perhaps.

tamarrow:
@prein do you have `DCOS_URL` set? It overrides the url being used.

prein:
Oh, I do have DCOS_URL set indeed, to the wrong one! Dunno how come, yet. Thank you for great hint @tamarrow

msabramo:
One simple thing to add would be to check for a `.dcos` directory in the current directory and prefer that over `~/.dcos`. Then you could have directories for different clusters. OTOH, folks may not always want to map a directory to a cluster. 

[...]

PR #1098, opened 2017-11-10
Lookup for DCOS_CLUSTER to determine if a cluster is attached

bamarni:
We support the DCOS_CLUSTER env variable since #994, we should take it into account when checking if a cluster is attached.

This is useful for automated scripts where one would want to attach to a cluster without affecting potential concurrent users (because of the shared attached file).

cf. https://github.com/dcos/dcos-cli/issues/595


bamarni:
run linux integration tests

PR #1099, opened 2017-11-10
Disable HTTP timeouts for dcos task requests

bamarni:
As this command relies on a long-running HTTP connection using chuncked transfer encoding, it is common to have no data to read from the server for an undefined period of time.

This happens for example when opening a shell session and not typing anything.

https://jira.mesosphere.com/browse/DCOS_OSS-1827

This problem was introduced in #1052.


bamarni:
run linux integration tests

klueska:
run windows integration tests

klueska:
run mac integration tests

PR #1101, opened 2017-11-15
Add a command to mark an agent as gone

bamarni:
This adds a `dcos node decommission` command which takes as argument an agent id. It will then mark it as gone using the Mesos operator API.

Currently the integration test is skipped because it requires the opposite of that action in order to put the cluster in a proper state again after that test.

cf. https://jira.mesosphere.com/browse/DCOS-18981


bamarni:
run all integration tests

bamarni:
run linux integration tests

bamarni:
The consensus on DCOS-18981 was to stick to the `decommission` naming. So this is ready for review.

PR #1102, opened 2017-11-17
Remove verbose failure for the deprecated --app-id

bamarni:
For the `package install` command, the --app-id flag used to trigger a warning, it then failed with an explicit error. This PR removes it completely.


bamarni:
run linux integration tests

PR #1103, opened 2017-11-17
Updated the "HTTP SSL Error" message.

joerg84:


joerg84:
Previously one got the following error: 

```Joergs-MacBook-Pro-247:1 joergschad$ dcos cluster setup https://joerg-tf654d-pub-mas-elb-1131462485.us-west-2.elb.amazonaws.com
SHA256 fingerprint of cluster certificate bundle:
5A:82:2F:82:09:17:AA:39:68:A8:84:E9:5B:56:1D:48:CC:15:BB:A8:52:9A:1A:EA:2A:63:D3:1D:4A:29:91:14 [yes/no] yes
An SSL error occurred. To configure your SSL settings, please run: `dcos config set core.ssl_verify <value>`
<value>: Whether to verify SSL certs for HTTPS or path to certs. Valid values are True, False, or a path to a CA_BUNDLE.```

Unfortunately, one can only run `dcos config set core.ssl_verify` after being attached to a cluster, so displaying that error message at that point is not helpful....

bamarni:
@joerg84 : could you provide more information about the kind of cluster you're connecting to? Is it a cluster with a load balancer in front? This might as well be already fixed with 4e3076a47928eeba56c351bf71f950dd20d3687e.

joerg84:
@bamarni It is created by our default terraform templates

joerg84:
In any case whenever I get to see that error message, the original message won't be helpful or?

bamarni:
Thanks for reporting @joerg84 , as seen in person yesterday this happened because the setup URL had an amazon domain which was not present in the certificate : 

`CertificateError("hostname 'joerg-tfe145-pub-mas-elb-2087857032.us-west-2.elb.amazonaws.com' doesn't match either of 'localhost', 'master.mesos', 'leader.mesos', '10.0.0.110', '127.0.0.1', '52.11.193.172', '10.0.0.110', '127.0.0.1', '52.11.193.172'",)`

The error message in such cases should be made more explicit, see https://jira.mesosphere.com/browse/DCOS_OSS-1817.

bamarni:
sorry, I meant https://jira.mesosphere.com/browse/DCOS_OSS-1881

PR #1106, opened 2017-11-20
 Accept an URL argument to `dcos auth list-providers`

bamarni:
The `dcos cluster setup` command accepts a `--provider` argument, however it is currently not possible to see all available providers for a given cluster, as the command for this assumes an already configured cluster.

This adds an optional `dcos_url` argument to `dcos auth list-providers`.

https://jira.mesosphere.com/browse/DCOS_OSS-1817

### DIFF
--- a/dcos/auth.py
+++ b/dcos/auth.py
@@ -328,19 +328,32 @@ def header_challenge_auth(dcos_url):
         raise DCOSAuthenticationException(response)
 
 
-def get_providers():
+def get_providers(dcos_url=None):
     """
     Returns dict of providers configured on cluster
 
+    :param dcos_url: url to cluster
+    :type dcos_url: str
     :returns: configured providers
     :rtype: {}
     """
 
-    dcos_url = config.get_config_val("core.dcos_url").rstrip('/')
+    if dcos_url:
+        # If an URL is specified, set verify to False
+        # as the CA cert for the cluster is likely missing.
+        verify = False
+    else:
+        dcos_url = config.get_config_val("core.dcos_url")
+        if not dcos_url:
+            msg = "No cluster is configured, please provide a dcos_url."
+            raise DCOSException(msg)
+        verify = None
+
     endpoint = '/acs/api/v1/auth/providers'
     url = urllib.parse.urljoin(dcos_url, endpoint)
+
     try:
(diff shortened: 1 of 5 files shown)


bamarni:
run linux integration tests

bamarni:
run linux integration tests

dcos/auth.py
```diff
@@ -328,19 +328,32 @@ def header_challenge_auth(dcos_url):
...
+    if not dcos_url:
+        dcos_url = config.get_config_val("core.dcos_url")
+        if not dcos_url:
+            msg = "No cluster is configured, please provide a dcos_url."
+            raise DCOSException(msg)
+        verify = None
+    else:
+        # If an URL is specified, set verify to False
+        # as the CA cert for the cluster is likely missing.
+        verify = False
```
mnaboka: nit:
```python
if dcos_url:
  # If an URL is specified, set verify to False
  # as the CA cert for the cluster is likely missing.
  verify = False
else:
  dcos_url = config.get_config_val("core.dcos_url")
  if not dcos_url:
    raise DCOSException('No cluster is configured, please provide a dcos_url.')
  verify = None
```

bamarni:
> i think it is a little bit weird that verify can be either None or False and never True. Can it be set to True?

Leaving it to `None` would let the http module decide based on the config what to choose, see https://github.com/bamarni/dcos-cli/blob/17027cf9b888a33f3c77883b3d0cf19236d21123/dcos/http.py#L74-L89

dcos/auth.py
bamarni: I've reverted the condition as suggested.

bamarni:
run linux integration tests

karlkfi:
Can you change the output text when you don't provide a provider and the cluster is configured with one to tell you to use this new command to look them up?

bamarni:
@karlkfi : This is a recurring problem with the dcos cluster command (see [here](https://github.com/dcos/dcos-cli/pull/1103) for a recent example where it suggests to run `dcos config` on an error during setup). The issue is that the cluster setup command calls directly other commands such as login and attach, which will fail with their own outputs. I'll be fixing this problem during the refactoring for DC/OS 1.12.

PR #1108, opened 2017-11-23
Trigger tests

bamarni:
This is not meant to be merged, just to be able to trigger tests against DC/OS 1.10.2 once it's released.


klueska:
There was no need to do this, as the PR that got merged previously already tested against the RC that ended up becoming the 1.10.2 release.

PR #1109, opened 2017-11-24
Linked clusters

bamarni:
This introduces a link and unlink subcommand which allow to link clusters together.

This is a new feature in DC/OS EE 1.11.

### DIFF
--- a/cli/dcoscli/auth/main.py
+++ b/cli/dcoscli/auth/main.py
@@ -178,9 +178,6 @@ def _trigger_client_method(
         auth.oidc_implicit_flow_auth(dcos_url)
     elif client_method == "dcos-usercredential-post-receive-authtoken" or \
             client_method == "dcos-credential-post-receive-authtoken":
-        if not username or not password:
-            msg = "Please specify username and password for provider [{}]"
-            raise DCOSException(msg.format(provider))
         auth.dcos_uid_password_auth(dcos_url, username, password)
     elif client_method == "dcos-servicecredential-post-receive-authtoken":
         if not username or not key_path:
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -1,4 +1,9 @@
+import json
 import os
+import socket
+import urllib
+
+from urllib.parse import urlparse
 
 import docopt
 import requests
@@ -9,13 +14,13 @@ from cryptography.hazmat.primitives import hashes
 
 import dcoscli
 
-from dcos import cluster, cmds, config, emitting, http, util
+from dcos import auth, cluster, cmds, config, emitting, http, jsonitem, util
 from dcos.errors import (DCOSAuthenticationException, DCOSException,
                          DefaultError)
 from dcoscli.auth.main import login
 from dcoscli.subcommand import default_command_info, default_doc
 from dcoscli.tables import clusters_table
-from dcoscli.util import confirm, decorate_docopt_usage
+from dcoscli.util import confirm, decorate_docopt_usage, prompt_with_choices
 
 emitter = emitting.FlatEmitter()
 logger = util.get_logger(__name__)
@@ -31,8 +36,20 @@ def main(argv):
 
 @decorate_docopt_usage
 def _main(argv):
+    doc_name = 'cluster'
+
+    # If a cluster is attached and the cluster linker API is reachable,
+    # use cluster_ee.txt.
+    dcos_url = config.get_config_val('core.dcos_url')
+    if dcos_url:
+        try:
+            cluster.get_cluster_links(dcos_url)
+            doc_name = 'cluster_ee'
+        except Exception:
+            pass
+
     args = docopt.docopt(
-        default_doc("cluster"),
+        default_doc(doc_name),
         argv=argv,
         version='dcos-cluster version {}'.format(dcoscli.version))
 
@@ -75,6 +92,16 @@ def _cmds():
             hierarchy=['cluster', 'rename'],
             arg_keys=['<name>', '<new_name>'],
             function=_rename),
+
+        cmds.Command(
+            hierarchy=['cluster', 'link'],
+            arg_keys=['<dcos_url>', '--provider'],
+            function=_link),
+
+        cmds.Command(
+            hierarchy=['cluster', 'unlink'],
+            arg_keys=['<name>'],
+            function=_unlink),
 
         cmds.Command(
             hierarchy=['cluster'],
@@ -151,10 +178,13 @@ def _attach(name):
     """
 
     c = cluster.get_cluster(name)
-    if c is not None:
-        return cluster.set_attached(c.get_cluster_path())
-    else:
+    if not c:
         raise DCOSException("Cluster [{}] does not exist".format(name))
+
+    if isinstance(c, cluster.LinkedCluster):
+        return setup(c.get_url(), provider=c.get_provider().get('id'))
+
+    return cluster.set_attached(c.get_cluster_path())
 
 
 def _rename(name, new_name):
@@ -175,6 +205,102 @@ def _rename(name, new_name):
         raise DCOSException(msg.format(new_name))
     else:
         config.set_val("cluster.name", new_name, c.get_config_path())
+
+
+def _link(dcos_url, provider_id):
+    """
+    Link a DC/OS cluster to the current one.
+
+    :param dcos_url: master ip of the cluster to link to
+    :type dcos_url: str
+    :param provider_id: login provider ID for the linked cluster
+    :type provider_id: str
+    :returns: process status
+    :rtype: int
+    """
+
+    current_cluster = cluster.get_attached_cluster()
+    if not current_cluster:
+        raise DCOSException('No cluster is attached, cannot link.')
+    current_cluster_url = current_cluster.get_url()
+
+    # Accept the same formats as the `setup` command
+    # eg. "my-cluster.example.com" -> "https://my-cluster.example.com"
+    dcos_url = jsonitem._parse_url(dcos_url)
+
+    try:
+        linked_cluster_ip = socket.gethostbyname(urlparse(dcos_url).netloc)
+    except OSError as error:
+        raise DCOSException("Unable to retrieve IP for '{dcos_url}': {error}"
+                            .format(dcos_url=dcos_url, error=error))
+
+    # Check if the linked cluster is already configured (based on its IP)
+    for configured_cluster in cluster.get_clusters():
+        configured_cluster_host = \
+            urlparse(configured_cluster.get_url()).netloc
+        configured_cluster_ip = socket.gethostbyname(configured_cluster_host)
+
+        if linked_cluster_ip == configured_cluster_ip:
+            linked_cluster_id = configured_cluster.get_cluster_id()
+            linked_cluster_name = configured_cluster.get_name()
+            break
+    else:
+        msg = ("The cluster you are linking to must be set up locally before\n"
+               "running the `cluster link` command. To set it up now, run:\n"
+               "    $ dcos cluster setup {}".format(dcos_url))
+        raise DCOSException(msg)
+
+    providers = auth.get_providers(dcos_url)
+
+    if provider_id:
+        if provider_id not in providers:
+            raise DCOSException(
+                "Incorrect provider ID '{}'.".format(provider_id))
+        provider_type = providers[provider_id]['authentication-type']
+    else:
+        (provider_id, provider_type) = _prompt_for_login_provider(providers)
+
+    message = {
+        'id': linked_cluster_id,
+        'name': linked_cluster_name,
+        'url': dcos_url,
+        'login_provider': {
+            'id': provider_id,
+            'type': provider_type}}
+
+    headers = {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json'}
+
+    http.post(
+        urllib.parse.urljoin(current_cluster_url, '/cluster/v1/links'),
+        data=json.dumps(message),
+        headers=headers)
+
+    return 0
+
+
+def _unlink(name):
+    """
+    Unlink a DC/OS cluster.
+
+    :param name:  name of the cluster
+    :type name: str
+    :returns: process status
+    :rtype: int
+    """
+
+    c = cluster.get_cluster(name)
+    if not c:
+        raise DCOSException('Unknown cluster {}.'.format(name))
+
+    dcos_url = config.get_config_val('core.dcos_url')
+    endpoint = urllib.parse.urljoin(
+            dcos_url, '/cluster/v1/links/' + c.get_cluster_id())
+
+    http.delete(endpoint)
+
+    return 0
 
 
 def setup(dcos_url,
@@ -332,3 +458,45 @@ def _needs_cluster_cert(dcos_url):
             return True
 
     return False
+
+
+def _prompt_for_login_provider(providers):
+    """Prompt the user with login providers.
+
+    :param providers: Login providers from "/acs/api/v1/auth/providers"
+    :type providers: dict
+    :returns: (provider_id, provider_type)
+    :rtype: (string, string)
+    """
+
+    choices = []
+    descriptions = []
+    for idx in providers.keys():
+        provider_type = providers[idx].get('authentication-type')
+        if provider_type not in [
+                auth.AUTH_TYPE_DCOS_UID_PASSWORD,
+                auth.AUTH_TYPE_DCOS_UID_PASSWORD_LDAP,
+                auth.AUTH_TYPE_SAML_SP_INITIATED,
+                auth.AUTH_TYPE_OIDC_AUTHORIZATION_CODE_FLOW]:
+            continue
+
+        choices.append(idx)
+        description = auth.auth_type_description(providers[idx])
+        descriptions.append(description)
+
+    if not choices:
+        raise DCOSException(
+            "There is no supported login provider on the linked cluster.")
+    elif len(choices) == 1:
+        # There is only one supported login provider, don't prompt.
+        provider_id = choices[0]
+        return (provider_id, providers[provider_id]['authentication-type'])
+
+    # Prompt user to choose a provider ID.
+    msg = ("Choose the login method and provider to enable switching to this "
+           "linked cluster:")
+    provider_id = prompt_with_choices(choices, descriptions, msg)
+    if not provider_id:
+        raise DCOSException("Incorrect login provider.")
+
+    return (provider_id, providers[provider_id]['authentication-type'])
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -106,8 +106,12 @@ def _list(json_, attached):
     :rtype: None
     """
 
-    clusters = [c.dict() for c in cluster.get_clusters()
-                if not attached or c.is_attached()]
+    if attached:
+        clusters = [c.dict() for c in cluster.get_clusters()
+                    if c.is_attached()]
+    else:
+        clusters = [c.dict() for c in cluster.get_clusters(True)]
+
     if json_:
         emitter.publish(clusters)
     elif len(clusters) == 0:
--- a/cli/dcoscli/util.py
+++ b/cli/dcoscli/util.py
@@ -4,7 +4,7 @@ from functools import wraps
 
 import docopt
 
-from dcos import emitting
+from dcos import emitting, errors
 
 emitter = emitting.FlatEmitter()
 
@@ -84,6 +84,48 @@ def confirm_text(prompt, confirmation_text):
             emitter.publish(msg)
             count += 1
     return False
+
+
+def prompt_with_choices(choices, descriptions=[], msg='Please choose:'):
+    """
+    Allow to pick an item from a list of choices. For example :
+
+        Please choose a login provider for your linked cluster:
+        1) dcos-uid-password
+        2) saml-sp-initiated
+        (1-2): [...]
+
+    :param choices: choices for the list
+    :type choices: list of str
+    :param descriptions: descriptions for the choices
+    :type descriptions: list of str
+    :param msg: the message to display before the choices
+    :type msg: str
+    :returns: the chosen item
+    :rtype: str
+    """
+
+    if not descriptions:
+        descriptions = choices
+
+    if len(choices) != len(descriptions):
+        msg = 'Count mismatch between choices and descriptions.'
+        raise errors.DCOSException(msg)
+
+    emitter.publish(msg)
+    for i, desc in enumerate(descriptions):
+        emitter.publish('{}) {}'.format(i+1, desc))
+
+    # Write to stdout without new line.
+    msg = '({}-{}): '.format(1, len(choices))
+    emitter.publish(msg, end='')
+
+    try:
+        chosen_id = int(_read_response())
+    except ValueError:
+        return None
+
+    return choices[chosen_id-1] if chosen_id <= len(choices) else None
 
 
 def _read_response():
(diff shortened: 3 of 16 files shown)


bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run linux integration tests

bamarni:
run all integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -31,8 +37,35 @@ def main(argv):
 
 @decorate_docopt_usage
 def _main(argv):
+    doc = default_doc("cluster")
```
klueska: Can you just create an alternate text file (one that includes the {link, unlink} commands) and conditionally pull in the correct? Seems cleaner than patching the contents like this. Especially since this is a one-off hack just for this release.

cli/dcoscli/util.py
```diff
@@ -86,6 +86,44 @@ def confirm_text(prompt, confirmation_text):
     return False
 
 
+def choicelist(choices, descriptions=[], msg='Please choose: '):
+    """
+    Allow to pick an item from a choicelist. For example :
```
klueska: What is a choicelist?

bamarni:
run linux integration tests

cli/dcoscli/auth/main.py
```diff
@@ -178,9 +178,6 @@ def _trigger_client_method(
         auth.oidc_implicit_flow_auth(dcos_url)
     elif client_method == "dcos-usercredential-post-receive-authtoken" or \
             client_method == "dcos-credential-post-receive-authtoken":
-        if not username or not password:
```
klueska: Can we make this:
```
if not (username and password):
...
```
It matches the text below more closely and is a bit easier to follow.

dcos/auth.py
```diff
@@ -155,7 +155,7 @@ def _get_dcostoken_by_dcos_uid_password_auth(
     username = username or url.username
     password = password or url.password
 
-    if password is None:
+    if not username or not password:
```
klueska: Likewise here.

cli/dcoscli/util.py
klueska: I'v just never heard the word choicelist before. I'm fine using it if it's a common term, I've just never heard it before.

cli/dcoscli/util.py
```diff
@@ -86,6 +86,46 @@ def confirm_text(prompt, confirmation_text):
...
+        Please choose a login provider for your linked cluster:
+        1) dcos-uid-password
+        2) saml-sp-initiated
+        (1-2): [...]
+
+    :param choices: choices for the list
+    :type choices: list of str
+    :param descriptions: descriptions for the choices
+    :type descriptions: list of str
+    :returns: the chosen item
```
klueska: You don't have `msg` in the list of parameters here.

cli/dcoscli/util.py
```diff
@@ -86,6 +86,46 @@ def confirm_text(prompt, confirmation_text):
...
+    if len(choices) != len(descriptions):
+        msg = 'Count mismatch between choices and descriptions.'
+        raise errors.DCOSException(msg)
+
+    emitter.publish(msg)
+    for i, desc in enumerate(descriptions):
+        emitter.publish('{}) {}'.format(i+1, desc))
+
+    # Write to stdout directly not to have a new line.
+    sys.stdout.write('({}-{}): '.format(1, len(choices)))
```
klueska: What are the implications of this as opposed to using `emitter`?

cli/dcoscli/util.py
```diff
@@ -86,6 +86,46 @@ def confirm_text(prompt, confirmation_text):
...
+    sys.stdout.flush()
+
+    try:
+        chosen_id = int(_read_response())
+    except ValueError:
+        return None
+
+    return choices[chosen_id-1] if len(choices) >= chosen_id else None
```
klueska: Can you reverse the logic here so that it's `chosen_id <= len(choices)`. That is the standard ordering of inequalities for things like this and it makes it clearer at a glance what is going on.

cli/dcoscli/cluster/main.py
```diff
@@ -31,8 +36,20 @@ def main(argv):
...
 @decorate_docopt_usage
 def _main(argv):
+    doc_name = 'cluster'
+
+    # If a cluster is attached and the cluster linker API is reachable,
+    # use cluster_ee.txt.
+    dcos_url = config.get_config_val('core.dcos_url')
+    if dcos_url:
+        try:
+            cluster.get_cluster_links(dcos_url)
```
klueska: This makes things really slow anytime I want to get the help menu (or run any other) `cluster` subcommand. Is there no other way of determining this information? We should probably be leveraging the cosmos CAPABILITIES API to discover this instead.

E.g.: https://github.com/dcos/cosmos/pull/702/files

cli/dcoscli/cluster/main.py
```diff
@@ -106,8 +133,11 @@ def _list(json_, attached):
     :rtype: None
     """
 
-    clusters = [c.dict() for c in cluster.get_clusters()
+    # Query for linked clusters only when listing all clusters.
+    linked = not attached
+    clusters = [c.dict() for c in cluster.get_clusters(linked)
                 if not attached or c.is_attached()]
```
klueska: The logic above here is confusing. Can you change it to make it more clear about what it si doing. I'm still not sure after reading it a few times. Otherwise, can you make the comment more clear.

cli/dcoscli/cluster/main.py
```diff
@@ -173,6 +206,98 @@ def _rename(name, new_name):
...
+    :returns: process status
+    :rtype: int
+    """
+
+    current_cluster = cluster.get_attached_cluster()
+    if not current_cluster:
+        raise DCOSException('No cluster is attached, cannot link.')
+    current_cluster_url = current_cluster.get_url()
+
+    # Accept the same formats than the `setup` command
```
klueska: s/than/as/

cli/dcoscli/cluster/main.py
```diff
@@ -173,6 +206,98 @@ def _rename(name, new_name):
...
+            raise DCOSException(
+                "Incorrect provider ID '{}'.".format(provider_id))
+        provider_type = providers[provider_id]['authentication-type']
+    else:
+        (provider_id, provider_type) = _prompt_for_login_provider(providers)
+
+    linked_cluster_ip = socket.gethostbyname(urlparse(dcos_url).netloc)
+
+    # Check if the linked cluster is already configured (based on its IP)
+    for configured_cluster in cluster.get_clusters():
```
klueska: I think we should do this first. I.e. before the line for:
```
providers = auth.get_providers(dcos_url)
```
on line 230.

cli/dcoscli/cluster/main.py
bamarni: I'm not too familiar with this API, though I don't get why Cosmos should be responsible for providing this information. The cluster linker is not a cosmos package or service.

For the performance side, we should probably store this kind of information in the config file as it's not meant to change during cluster lifetime. Also this is only requesting when getting the `dcos cluster` help menu, not the main one.

cli/dcoscli/cluster/main.py
bamarni: I also found it unclear at the beginning, that's why I made a variable before and added a comment.

This is the `list` command, the behavior here being that `dcos cluster list --attached` should not query for linked cluster but only return the current one. That's why I added a boolean argument to `get_clusters()`. I'll clarify the comment

cli/dcoscli/cluster/main.py
klueska: Yeah, it's weird that it's part of the cosmos, but that's where it currently lives at the moment (even though it has absolutely nothing to do with cosmos). There has been talk of moving this information somewhere else and building a better schema for it, but it hasn't happened yet.

In the meantime, we should add a CAPABILITY for CLUSTER_LINKING in there and use that here instead of trying to poll the service.

cli/dcoscli/util.py
bamarni: Not sure what you mean by standard, command-line tools such as `dialog` or `whiptail` use the term `menu` : 
> As  its  name  suggests, a menu box is a dialog box that can be used to present a list of choices in the form of a menu for the user to choose.

They also have the terms `list of choices` in the associated description, I've went for `choicelist` as it's more explicit than `menu` and clearly defines what that helper do. Let me know if you prefer `menu`.

cli/dcoscli/util.py
bamarni: The problem with emitter is that it puts a new line at the end, which would make it look like this : 
```
Please choose a login provider for your linked cluster:
1) dcos-uid-password
2) saml-sp-initiated
(1-2):
[waiting for input here...]
```
And I wanted the cursor to be right new to `(1-2)`.

klueska: Sure. I know that emitter prints the extra line. I'm more curious about what *else* emitter does (other than print) that we are missing out on by not using it here.

cli/dcoscli/cluster/main.py
bamarni: That capability would only be for DC/OS EE, I do not see such distinction in the list on the PR you've linked. If I add it there I assume it would also show-up for DC/OS open?

cli/dcoscli/cluster/main.py
klueska: When I read the line:
```
cluster.get_clusters(linked)
```
that implies to me that I am getting the set of clusters that are linked (and no others)

Or if we changed it to:
```
cluster.get_clusters(not attached)
```
that implies to me that we are getting the list of non-attached clusters.

which then seems weird to combine with the condition for `if not attached or c.is_attached()` below it.

The whole thing is just confusing to read from left to right and should be made clearer for someone without having to look at the semantics of what the argument to `cluster.get_clusters(...)` is.

cli/dcoscli/util.py
klueska: What about `prompt_with_choices()` since we actually block this call with a prompt.

cli/dcoscli/util.py
bamarni: The emitter just calls `print()`, that's how that function behaves, though I could add some arguments and rewrite these 2 lines as : 

    print('({}-{}): '.format(1, len(choices)), end='', flush=True)

bamarni: `end` defaulting to `\n`

bamarni: for reference https://stackoverflow.com/questions/493386/how-to-print-without-newline-or-space

klueska: I'd rather extend the `emitter.publish()` command to take an argument to not print a newline somehow. I don't like having multiple ways of printing to the terminal.

cli/dcoscli/cluster/main.py
bamarni: Here I could make the variable more explicit (eg. `with_links`, `attached_only`) and clarify the comment. Would that help?

Otherwise I'll switch to a standard loop instead of a list comprehension.

cli/dcoscli/util.py
bamarni: So do you suggest that I add an `end` and `flush` arguments to the emitter abstract class?

The flat emitter is just an implementation of it, and the print handler is just the default handler for that flat emitter. I'm not sure what was the intent for having the emitter abstraction in the first place, but here I feel like we might push some implementation-specific concepts into the abstraction.

cli/dcoscli/cluster/main.py
klueska: Yeah, that seems like a big limitation here. I just don't like that if connecting to an OSS cluster, we are stuck *always* waiting for the 5 (or more) second timeout for any `cluster` command we issue.

If this is the only way at the moment to reliably get the information about linked cluster support, I'd rather do something smarter so that it doesn't take so long on OSS clusters. Do we have a way of reliably detecting OSS vs. EE at the moment? If so, the logic should be something like:
```
if (EE) and (has linked cluster API):
   ...
```

cli/dcoscli/util.py
bamarni: lgtm, I'll update the function name

cli/dcoscli/cluster/main.py
klueska: Ignoring the implementation for the minute, let me make sure I understand what this is trying to do. In (really) long form, would it be something like:
```
if attached:
    clusters = [c.dict() for c in cluster.get_clusters() if c.is_attached()]
else:
    clusters = clusters = [c.dict() for c in cluster.get_clusters(True)]
```
If so, this feels much clearer to me than trying to wrap that all into a single statement (and is actually faster since there are less comparisons per iteration in the comprehension loop).

cli/dcoscli/cluster/main.py
bamarni: Not sure, I saw this issue getting mentioned by @branden on slack (https://jira.mesosphere.com/browse/DCOS_OSS-1160).
Which says : `There's no easy way to determine if a cluster is OSS or EE before logging in.`

Does it mean that there is an easy way once logged in?

cli/dcoscli/util.py
klueska: As I said, I just don't like having multiple ways of writing out to the console. At some point someone added the emitter class to abstract `print()` away, and it feels weird to sprinkle code around the code base that bypasses that. I'd rather extend the abstraction to do the right thing.

cli/dcoscli/cluster/main.py
bamarni: I think no matter how we check it (it's going to be an HTTP request probably), the performance concern would be addressed by storing some information in the config file, not to have to look-up each time.

klueska: I disagree. If the linker API is actually there, it should return immediately withotu waiting for the timeout. This amount of time is negligible. It's only when it's not there and we have to suffer through the timeout that it's a problem.

bamarni: >  I just don't like that if connecting to an OSS cluster, we are stuck always waiting for the 5 (or more) second timeout 

This wouldn't hit the request timeout, as the cluster would return early with a 404.

bamarni: I was confused, I thought you were worried about the delay of making an HTTP request each time, that's why I suggested using the config file to cache the info.

klueska: OK. Makes sense. It's probably fine as is (for now). When we refactor the CLI for 1.12, this sorts of weird edge cases should go away, hopefully.

bamarni:
run linux integration tests

cli/dcoscli/cluster/main.py
klueska: I'm not convinced that this isn't waiting for the timeout for clusters without the linking service:
```
$ time dcos cluster list
                 NAME                                CLUSTER ID                  STATUS    VERSION            URL            
  klueska-05c7d4a0-ee-testing-master    25c036d0-3de4-4b62-94fb-230d6697dc5a   AVAILABLE   1.11-dev  https://35.197.28.151   
  klueska-b093da7d-ee-testing-master    38f071ed-983f-4bc1-bbfc-a6243e0b7eb9  UNAVAILABLE    N/A     https://35.203.152.39   
hybrid-demo-041a3598-ee-testing-master  4efa4abc-90f3-4d7c-b26f-13587a62216a  UNAVAILABLE    N/A      https://35.197.88.36   
     gpu-test-3a3f136a-ee-stable        5000c578-57d4-49d9-8164-10bb27c8a957  UNAVAILABLE    N/A     https://35.185.215.70   
           klueska-l6ds661              62175879-db9e-4187-908b-da4ca25201e1  UNAVAILABLE    N/A      https://52.25.121.75   
     gpu-test-06744c9a-ee-stable        a8866d32-bded-41c9-8dc6-c6a3d52115b3  UNAVAILABLE    N/A     https://35.197.70.255   
      tf-demo-2e2d7cb3-ee-stable        cec885af-748e-4345-b593-a7eca7781c55  UNAVAILABLE    N/A     https://35.197.28.151   
hybrid-demo-d85b8f67-ee-testing-master  e24793ad-1218-4961-90c1-81a2b6dde276  UNAVAILABLE    N/A     https://35.203.167.157  
 eb76b5d5-f32c-426c-b2d7-633e850509c6   eb76b5d5-f32c-426c-b2d7-633e850509c6  UNAVAILABLE    N/A     https://35.197.117.222  

real	1m3.872s
user	0m0.508s
sys	0m0.097s
```

cli/dcoscli/cluster/main.py
```diff
@@ -173,6 +207,98 @@ def _rename(name, new_name):
...
+        configured_cluster_ip = socket.gethostbyname(configured_cluster_host)
+
+        if linked_cluster_ip == configured_cluster_ip:
+            linked_cluster_id = configured_cluster.get_cluster_id()
+            linked_cluster_name = configured_cluster.get_name()
+            break
+    else:
+        msg = ("The cluster you are linking to must be set up locally before "
+               "running the `cluster link` command.\n\nTo set it up now, run:"
+               " dcos cluster setup {}".format(dcos_url))
```
klueska: This looks awkward with the line break:
```
$ dcos cluster link 35.199.177.232
The cluster you are linking to must be set up locally before running the `cluster link` command.

To set it up now, run: dcos cluster setup https://35.199.177.232
```

I'd rather see it look like:
```
$ dcos cluster link 35.199.177.232
The cluster you are linking to must be set up locally before running the
`cluster link` command.  To set it up now, run:
    $ dcos cluster setup https://35.199.177.232
```

dcos/cluster.py
```diff
@@ -202,24 +229,49 @@ def get_cluster_cert(dcos_url):
         raise DCOSException(error_msg)
 
 
-def get_clusters():
+def get_clusters(linked=False):
```
klueska: I would call this argument: `include_linked`

klueska:
The expected workflow looks good now when manually testing:
```
$ dcos cluster setup 35.199.177.232
SHA256 fingerprint of cluster certificate bundle:
7C:AB:9C:08:9A:1E:A4:74:6D:36:A5:DC:B1:A9:1C:3B:05:65:EC:78:CF:3D:44:16:18:50:5E:87:86:1C:0D:70 [yes/no] yes
35.199.177.232's username: bootstrapuser
bootstrapuser@35.199.177.232's password: 

$ dcos cluster setup 35.197.28.151
SHA256 fingerprint of cluster certificate bundle:
C1:95:04:0B:7C:38:7B:88:0F:BE:41:2A:95:40:78:2C:FB:C6:B6:E0:7A:3D:44:F2:5D:E1:17:02:6A:E8:E5:36 [yes/no] yes
35.197.28.151's username: bootstrapuser
bootstrapuser@35.197.28.151's password: 

$ dcos cluster list
                NAME                              CLUSTER ID                 STATUS   VERSION            URL            
 klueska-765ee9dd-ee-testing-master  149d65a0-13b0-4921-99aa-de7a4894ff41  AVAILABLE  1.11-dev  https://35.199.177.232  
klueska-05c7d4a0-ee-testing-master*  25c036d0-3de4-4b62-94fb-230d6697dc5a  AVAILABLE  1.11-dev  https://35.197.28.151   

$ dcos cluster link https://35.199.177.232

$ dcos cluster attach klueska-765ee9dd-ee-testing-master

$ dcos cluster link https://35.197.28.151
```

However, if I do something stupid, it fails in a nasty way:
```
$ dcos cluster link klueska-05c7d4a0-ee-testing-master
  File "/Users/klueska/projects/dcos/dcos-cli/cli/dcoscli/subcommand.py", line 101, in run_and_capture
    exit_code = m.main([self._command] + self._args)
  File "/Users/klueska/projects/dcos/dcos-cli/cli/dcoscli/cluster/main.py", line 31, in main
    return _main(argv)
  File "/Users/klueska/projects/dcos/dcos-cli/cli/dcoscli/util.py", line 24, in wrapper
    result = func(*args, **kwargs)
  File "/Users/klueska/projects/dcos/dcos-cli/cli/dcoscli/cluster/main.py", line 58, in _main
    return cmds.execute(_cmds(), args)
  File "/Users/klueska/projects/dcos/dcos-cli/dcos/cmds.py", line 43, in execute
    return function(*params)
  File "/Users/klueska/projects/dcos/dcos-cli/cli/dcoscli/cluster/main.py", line 231, in _link
    linked_cluster_ip = socket.gethostbyname(urlparse(dcos_url).netloc)
socket.gaierror: [Errno 8] nodename nor servname provided, or not known
```

klueska:
This should fix that error above:
```
diff --git a/cli/dcoscli/cluster/main.py b/cli/dcoscli/cluster/main.py
index a1903db6..d7001d7e 100644
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -228,7 +228,11 @@ def _link(dcos_url, provider_id):
     # eg. "my-cluster.example.com" -> "https://my-cluster.example.com"
     dcos_url = jsonitem._parse_url(dcos_url)
 
-    linked_cluster_ip = socket.gethostbyname(urlparse(dcos_url).netloc)
+    try:
+        linked_cluster_ip = socket.gethostbyname(urlparse(dcos_url).netloc)
+    except socket.gaierror as error:
+        raise DCOSException("Unable to retrieve IP for '{dcos_url}': {error}"
+                            .format(dcos_url=dcos_url, error=error))
 
     # Check if the linked cluster is already configured (based on its IP)
     for configured_cluster in cluster.get_clusters():
```

dcos/cluster.py
```diff
@@ -223,24 +229,49 @@ def get_cluster_cert(dcos_url):
...
     """
-    :returns: list of configured Clusters
+    Get configured or linked clusters.
+
+    :param linked: whether to look for linked clusters
+    :type linked: bool
```
klueska: This will need to change too then.

cli/dcoscli/cluster/main.py
klueska: I take it back. This is slow because I have so many unreachable clusters. SO it (by definitiion) has to wait for the timeout...

cli/dcoscli/cluster/main.py
klueska: This should do it:
```
        msg = ("The cluster you are linking to must be set up locally before\n"
               "running the `cluster link` command. To set it up now, run:\n"
               "    $ dcos cluster setup {}".format(dcos_url))
```

cli/dcoscli/cluster/main.py
collinvandyck: would be nice if it were done concurrently

dcos/auth.py
```diff
@@ -14,6 +14,13 @@
 
 logger = util.get_logger(__name__)
 
+AUTH_TYPE_DCOS_UID_PASSWORD = "dcos-uid-password"
+AUTH_TYPE_DCOS_UID_SERVICEKEY = "dcos-uid-servicekey"
+AUTH_TYPE_DCOS_UID_PASSWORD_LDAP = "dcos-uid-password-ldap"
+AUTH_TYPE_SAML_SP_INITIATED = "saml-sp-initiated"
+AUTH_TYPE_OIDC_AUTHORIZATION_CODE_FLOW = "oidc-authorization-code-flow"
+AUTH_TYPE_IODC_IMPLICIT_FLOW = "oidc-implicit-flow"
```
vinodkone: typo? s/IODC/OIDC/ ?

cli/dcoscli/cluster/main.py
bamarni: Yes, this is slow when a lot of clusters are unavailable.

Doing requests concurrently would be a nice improvement, otherwise at the bare minimum we could at least print each item one-by-one instead of putting all results in memory.

dcos/auth.py
bamarni: good catch, fixed

bamarni:
run linux integration tests

klueska:
run mac integration tests

klueska:
run windows integration tests

PR #1116, opened 2017-11-28
Bugfix: Metronome was NOT using the schema defined configuration

kensipe:
metronome is not used externally.  
it is `job.url` according to the schema.

### DIFF
--- a/dcos/metronome.py
+++ b/dcos/metronome.py
@@ -47,7 +47,7 @@ def _get_metronome_url(toml_config=None):
     if toml_config is None:
         toml_config = config.get_config()
 
-    metronome_url = config.get_config_val('metronome.url', toml_config)
+    metronome_url = config.get_config_val('job.url', toml_config)
     if metronome_url is None:
         # dcos must be capable to use dcos_url
         _check_capability()


kensipe:
@bamarni this was a bug I discovered

kensipe:
run linux-integration-tests

kensipe:
run linux integration tests

bamarni:
@kensipe : indeed that config doesn't exist under `metronome` but is under `job`.

Is metronome usually installed on the DC/OS cluster and this is an option to set when it's on an external cluster instead? And when it lives on an external cluster, won't it need some sort of authorization credentials?

kensipe:
@bamarni yeah...  at this point we added the ability it was just misconfigured in the CLI.   It is also consistent with marathon (dcos config set marathon.url).   For marathon it is useful to set to a marathon on marathon (mom) but I also use it to test locally in dev (which helps to make sure our stuff is tested by us).   In the case of metronome, we currently do not allow a metronome on marathon... however that may change with multi-tenancy (not sure).     
And as you point out,  testing locally does work but it requires me to have a live cluster to do so.  I was hoping at some point to remove that or make auth a little more lazy (only enforced when needed).

kensipe:
@bamarni looks green.   thanks for your help!

alenkacz:
@kensipe can you maybe help me understand this? I looked here https://dcos.io/docs/1.10/cli/command-reference/dcos-config/dcos-config-set/ but I did not find job.url there. What schema are you referring to? how do I actually verify what is the correct option for this?

bamarni:
@alenkacz : I believe it's this one (https://github.com/dcos/dcos-cli/blob/master/dcos/data/config-schema/job.json), from the CLI perspective this change seems correct, that `metronome.url` config option doesn't exist. So if you both are fine with this then LGTM too.

PR #1118, opened 2017-11-30
Add missing case to cosmos error handling

larioj:
Fixes https://jira.mesosphere.com/browse/COPS-2067

### DIFF
--- a/dcos/packagemanager.py
+++ b/dcos/packagemanager.py
@@ -41,7 +41,7 @@ def cosmos_error(fn):
             logger.debug("Error: {}".format(response.json()))
             error_msg = _format_error_message(response.json())
             raise DCOSException(error_msg)
-        elif int(response.status_code / 100) != 2:
+        elif response.status_code not in range(200, 300):
             raise DCOSHTTPException(response)
         return response
 
--- a/dcos/packagemanager.py
+++ b/dcos/packagemanager.py
@@ -41,7 +41,7 @@ def cosmos_error(fn):
             logger.debug("Error: {}".format(response.json()))
             error_msg = _format_error_message(response.json())
             raise DCOSException(error_msg)
-        elif (response.status_code / 100) != 2:
+        elif int(response.status_code / 100) != 2:
             raise DCOSHTTPException(response)
         return response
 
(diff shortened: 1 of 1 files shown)


larioj:
@jsancio @takirala Quick review

dcos/packagemanager.py
```diff
@@ -41,6 +41,8 @@ def check_for_cosmos_error(*args, **kwargs):
             logger.debug("Error: {}".format(response.json()))
             error_msg = _format_error_message(response.json())
             raise DCOSException(error_msg)
+        elif (response.status_code / 100) != 2:
+            raise DCOSHTTPException(response)
```
jsancio: How does this look to the user? I.e. can I see an example of this error?

dcos/packagemanager.py
```diff
@@ -41,6 +41,8 @@ def check_for_cosmos_error(*args, **kwargs):
             logger.debug("Error: {}".format(response.json()))
             error_msg = _format_error_message(response.json())
             raise DCOSException(error_msg)
+        elif (response.status_code / 100) != 2:
```
bamarni: I don't think this does what you meant to do, since Python 3 a division between 2 ints returns a float (cf. https://www.python.org/dev/peps/pep-0238/).

dcos/packagemanager.py
larioj: It looks like this

```
Error while fetching [http://jlariosmu-elasticl-15r2m2ribeead-755510712.us-west-2.elb.amazonaws.com/package/describe]: HTTP 504: "Gateway Time-out".
```

dcos/packagemanager.py
larioj: Thanks! I casted the result of the division to an int. I don't this is the most idiomatic way to do it.

jsancio: Probably `range` is the idiomatic way. E.g. `reponse.status_code not in range(200, 300)`

bamarni: I'd suggest `if 200 <= response.status_code < 300:` which I find easier to read.

larioj:
@jsancio @bamarni I've made the changes requested. The 'linux-integration-tests' check has been pending for multiple days.

bamarni:
@larioj : tests need to be triggered manually, however they'd be failing right now because of a test which started to fail on the package integration suite (https://mesosphere.slack.com/archives/C02H2798B/p1512426126000076).

I'm working on fixing this then I'll move this PR forward.

bamarni:
run linux integration tests

PR #1119, opened 2017-12-04
Fix CI

bamarni:
* Update dcos-launch download URL
* Wait for DNS records to be propagated in package tests
* Use https when adding 1.7 universe in tests
* Use helloworld instead of marathon in a test for `pacakge describe`


bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
run all integration tests

cli/dcoscli/test/package.py
```diff
@@ -48,30 +52,25 @@ def package(package_name, deploy=False, args=[]):
 
 def setup_universe_server():
     # add both Unvierse servers with static packages
-    add_app('tests/data/universe-v3-stub.json', True)
-    add_app('tests/data/helloworld-v3-stub.json', True)
+    add_app('tests/data/universe-v3-stub.json', False)
+    add_app('tests/data/helloworld-v3-stub.json', False)
```
klueska: What does False do here instead of True?

cli/dcoscli/test/package.py
```diff
@@ -48,30 +52,25 @@ def package(package_name, deploy=False, args=[]):
...
 
     assert_command(
         ['dcos', 'package', 'repo', 'remove', 'Universe'])
 
+    watch_all_deployments()
+
     # Add the two test repos to Cosmos
     for name, url in UNIVERSE_TEST_REPOS.items():
-        assert_command(['dcos', 'package', 'repo', 'add', name, url])
+        # wait for DNS records for the universe app to be propagated
```
klueska: This looks like a completely different strategy from before for waiting for the universe to come online. What's better about this way (or different at least)?

cli/tests/integrations/test_package.py
```diff
@@ -84,7 +84,7 @@ def test_repo_list():
         assert_command(['dcos', 'package', 'repo', 'list'], stdout=repo_list)
 
 
-def test_repo_add():
+def test_repo_add_and_remove():
```
klueska: Why do you combine these now?

cli/dcoscli/test/package.py
bamarni: it proceeds without watching for deployments, this is not needed as there is a `watch_all_deployments()` below.

cli/dcoscli/test/package.py
bamarni: `pacakge repo add` now fails if the universe service is unreachable, see see https://mesosphere.slack.com/archives/C02H2798B/p1512426126000076

cli/tests/integrations/test_package.py
bamarni: they were order-dependent, cf. the commit message.

bamarni:
run all integration tests

bamarni:
@takirala @jsancio : I'd appreciate your confirmation on this (cf. commit messages for detailed descriptions), seems like a new version of cosmos has been released lately into dcos and a few tests began to fail.

bamarni:
run all integration tests

PR #1123, opened 2017-12-11
Cluster linker improvements

bamarni:
- Sort login providers during the `cluster link` prompt
- A LinkedCluster object might be a configured cluster
- Add --linked flag to `dcos cluster list`
- Don't error out in case no cluster is listed
- Don't rely on the http module to get cluster version

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -158,10 +158,7 @@ def _list(json_, attached, linked=False):
         if attached:
             msg = ("No cluster is attached. "
                    "Please run `dcos cluster attach <cluster-name>")
-        else:
-            msg = ("No clusters are currently configured. "
-                   "To configure one, run `dcos cluster setup <dcos_url>`")
-        raise DCOSException(msg)
+            raise DCOSException(msg)
     else:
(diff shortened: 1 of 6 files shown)


bamarni:
run linux integration tests

bamarni:
run linux integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -55,15 +55,21 @@ def _main(argv):
...
     """
+    :param doc_name: the docopt help file in use
+    :type doc_name: str
     :returns: all the supported commands
     :rtype: list of dcos.cmds.Command
     """
 
+    list_arg_keys = ['--json', '--attached']
+    if doc_name == 'cluster_ee':
+        list_arg_keys.append('--linked')
```
bamarni: So here we'd have an option which is only for EE, this is a hack to have it not failing on Open. Should I add the `--linked` flag there too instead?

bamarni:
run linux integration tests

cli/dcoscli/cluster/main.py
klueska: This is fine for now.

PR #1125, opened 2017-12-12
Make concurrent requests when listing clusters

bamarni:
When converting a Cluster into a dict, an HTTP request is being made in order to get the cluster version.

This makes use of Python's `concurrent.futures` module to trigger requests asynchronously.

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -3,6 +3,7 @@ import os
 import socket
 import urllib
 
+from concurrent import futures
 from urllib.parse import urlparse
 
 import docopt
@@ -147,10 +148,18 @@ def _list(json_, attached, linked=False):
     if attached:
         clusters = [c.dict() for c in cluster.get_clusters()
                     if c.is_attached()]
-    elif linked:
-        clusters = [c.dict() for c in cluster.get_linked_clusters()]
     else:
-        clusters = [c.dict() for c in cluster.get_clusters(True)]
+        if linked:
+            clusters = cluster.get_linked_clusters()
+        else:
+            clusters = cluster.get_clusters(True)
+
+        if clusters:
+            # Query for cluster versions concurrently.
+            nb_workers = len(clusters)
+            with futures.ThreadPoolExecutor(max_workers=nb_workers) as pool:
+                reqs = [pool.submit(c.dict) for c in clusters]
+                clusters = [r.result() for r in futures.as_completed(reqs)]
 
     if json_:
         emitter.publish(clusters)


bamarni:
run all integration tests

bamarni:
run all integration tests

bamarni:
As suggested by @collinvandyck in #1109

cli/dcoscli/cluster/main.py
```diff
@@ -147,10 +148,22 @@ def _list(json_, attached, linked=False):
...
     else:
-        clusters = [c.dict() for c in cluster.get_clusters(True)]
+        if linked:
+            clusters = cluster.get_linked_clusters()
+        else:
+            clusters = cluster.get_clusters(True)
+
+        nb_clusters = len(clusters)
+
+        if nb_clusters > 1:
```
collinvandyck: It would probably be cleaner in the code to perform the requests concurrently no matter the number of clusters

cli/dcoscli/cluster/main.py
```diff
@@ -147,10 +148,22 @@ def _list(json_, attached, linked=False):
...
+            # If there are multiple clusters,
+            # query for their versions concurrently.
+            with futures.ThreadPoolExecutor(max_workers=nb_clusters) as pool:
+                reqs = [pool.submit(c.dict) for c in clusters]
+                clusters = [r.result() for r in futures.as_completed(reqs)]
```
collinvandyck: If we're using `as_completed`, does that mean that the ordering of the results can change every time we fetch them?  If so, I think we probably want to maintain the order, somehow.

cli/dcoscli/cluster/main.py
bamarni: updated

cli/dcoscli/cluster/main.py
bamarni: The order in the table will always be the same (https://github.com/dcos/dcos-cli/blob/beb37ffbbfaf7118ed0e36e0a14a9bec39edf034/cli/dcoscli/tables.py#L920), as for the JSON it might vary. Why do you think that's an issue?

collinvandyck: Ah ok. I didn't understand that it would be sorted after the fetching was complete.

bamarni: Not when passing the `--json` flag though, let me know if that's a concern and why.

collinvandyck: it's not a concern

bamarni:
@klueska : I cannot think of other places from the top of my head. I've added test time duration reports to the CI builds lately, that might help us regarding this matter.

As for the abstraction, there are different ways to do concurrent things in Python (http://masnun.rocks/2016/10/06/async-python-the-different-forms-of-concurrency/) so we'd have to come-up with the proper one. I also feel like it's pretty straightforward to use the stdlib directly, in the current case it was basically just a with statement with an additional list comprehension.

PR #1126, opened 2017-12-13
HTTP timeout improvements

bamarni:
This leverages the tuple timeout of the requests package in order to distinguish between the connection timeout and the response read timeout.

When a timeout is passed as a numeric value, it refers to the read_timeout.

When no explicit timeout is passed, the `core.timeout` value in the config file is now used, or if not present, defaults to 180. This also refers to the read_timeout.

The connection timeout defaults to 5. In order to overwrite it, one must pass a tuple of the form `(connection_timeout, read_timeout)`.

https://jira.mesosphere.com/browse/DCOS_OSS-1956
https://jira.mesosphere.com/browse/DCOS_OSS-1957

### DIFF
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -106,7 +106,9 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     try:
         # find cluster id
         cluster_url = dcos_url.rstrip('/') + '/metadata'
-        res = http.get(cluster_url, timeout=1)
+        # This is an informational request,
+        # a 5 seconds read timeout is enough.
+        res = http.get(cluster_url, timeout=5)
         cluster_id = res.json().get("CLUSTER_ID")
 
     except DCOSException as e:
@@ -138,7 +140,11 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     cluster_name = cluster_id
     try:
         url = dcos_url.rstrip('/') + '/mesos/state-summary'
-        name_query = http.get(url, toml_config=cluster.get_config())
+        # This is an informational request,
+        # a 5 seconds read timeout is enough.
+        name_query = http.get(url,
+                              toml_config=cluster.get_config(),
+                              timeout=5)
         cluster_name = name_query.json().get("cluster")
 
     except DCOSException:
@@ -378,6 +384,8 @@ class Cluster():
             self.get_url(), "dcos-metadata/dcos-version.json")
 
         try:
+            # This is an informational request,
+            # a 5 seconds read timeout is enough.
             resp = requests.request(
                 'GET',
                 url=endpoint,
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -389,7 +389,7 @@ class Cluster():
             resp = requests.request(
                 'GET',
                 url=endpoint,
-                timeout=http.DEFAULT_TIMEOUT,
+                timeout=5,
                 verify=False)
         except Exception as e:
             return VERSION_UNKNOWN
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -238,7 +238,7 @@ def request(method,
     dcos_url = urlparse(config.get_config_val("core.dcos_url", toml_config))
 
     # only request with DC/OS Auth if request is to DC/OS cluster
-    if auth_token and _is_request_to_dcos(url):
+    if auth_token and _is_request_to_dcos(url, toml_config=toml_config):
         auth = DCOSAcsAuth(auth_token)
     else:
         auth = None
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -13,7 +13,18 @@ from dcos.errors import (DCOSAuthenticationException,
 
 logger = util.get_logger(__name__)
 
-DEFAULT_TIMEOUT = 5
+DEFAULT_READ_TIMEOUT = 180
+"""The default timeout for reading an HTTP response, it can be overriden
+through the `core.timeout` config. This is not a time limit on the entire
+response download; rather, an exception is raised if the server has not issued
+a response for timeout seconds (more precisely, if no bytes have been received
+on the underlying socket for timeout seconds)."""
+
+DEFAULT_CONNECT_TIMEOUT = 5
+"""The default timeout for establishing an HTTP connection."""
+
+DEFAULT_TIMEOUT = (DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT)
+"""The default timeout tuple for connection and read."""
 
 
 def _default_is_success(status_code):
@@ -93,7 +104,7 @@ def _verify_ssl(url, verify=None, toml_config=None):
 def _request(method,
              url,
              is_success=_default_is_success,
-             timeout=DEFAULT_TIMEOUT,
+             timeout=True,
              auth=None,
              verify=None,
              toml_config=None,
@@ -106,8 +117,15 @@ def _request(method,
     :type url: str
     :param is_success: Defines successful status codes for the request
     :type is_success: Function from int to bool
-    :param timeout: request timeout
-    :type timeout: int
+    :param timeout: How many seconds to wait for the server to send data
+                    before aborting and raising an exception. A timeout is
+                    either a float, a (connect timeout, read timeout) tuple
+                    or `None`. `None` means unlimited timeout. If no timeout
+                    is passed it defaults to the DEFAULT_TIMEOUT tuple, where
+                    the connect timeout can optionally be overridden by the
+                    `core.timeout` config.
+    :type timeout: int | float | None | bool |
+                   (int | float | None, int | float | None)
     :param auth: authentication
     :type auth: AuthBase
     :param verify: whether to verify SSL certs or path to cert(s)
@@ -119,6 +137,15 @@ def _request(method,
     :type kwargs: dict
     :rtype: Response
     """
+
+    if timeout is True:
+        if toml_config is None:
+            toml_config = config.get_config()
+
+        timeout = config.get_config_val("core.timeout", toml_config)
+        timeout = (DEFAULT_CONNECT_TIMEOUT, timeout or DEFAULT_READ_TIMEOUT)
+    elif type(timeout) in (float, int):
+        timeout = (DEFAULT_CONNECT_TIMEOUT, timeout)
 
     if 'headers' not in kwargs:
         kwargs['headers'] = {'Accept': 'application/json'}
@@ -171,7 +198,7 @@ def _request(method,
 def request(method,
             url,
             is_success=_default_is_success,
-            timeout=DEFAULT_TIMEOUT,
+            timeout=True,
             verify=None,
             toml_config=None,
             **kwargs):
@@ -184,8 +211,15 @@ def request(method,
     :type url: str
     :param is_success: Defines successful status codes for the request
     :type is_success: Function from int to bool
-    :param timeout: request timeout
-    :type timeout: int
+    :param timeout: How many seconds to wait for the server to send data
+                    before aborting and raising an exception. A timeout is
+                    either a float, a (connect timeout, read timeout) tuple
+                    or `None`. `None` means unlimited timeout. If no timeout
+                    is passed it defaults to the DEFAULT_TIMEOUT tuple, where
+                    the connect timeout can optionally be overridden by the
+                    `core.timeout` config.
+    :type timeout: int | float | None | bool |
+                   (int | float | None, int | float | None)
     :param verify: whether to verify SSL certs or path to cert(s)
     :type verify: bool | str
     :param toml_config: cluster config to use
--- a/cli/tests/integrations/test_package.py
+++ b/cli/tests/integrations/test_package.py
@@ -8,7 +8,7 @@ import sys
 import pytest
 import six
 
-from dcos import config, constants, errors, subcommand, util
+from dcos import config, constants, errors, http, subcommand, util
 
 from dcoscli.test.common import (assert_command, assert_lines, base64_to_dict,
                                  delete_zk_node,
@@ -1095,8 +1095,9 @@ def _package(name,
     command = ['dcos', 'package', 'install', name] + args
 
     installed = False
+    timeout = http.DEFAULT_READ_TIMEOUT
     try:
-        returncode_, stdout_, stderr_ = exec_command(command, timeout=30)
+        returncode_, stdout_, _ = exec_command(command, timeout=timeout)
         installed = (returncode_ == 0)
 
         assert installed
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1,0 +1,81 @@
+from mock import patch
+
+from requests import Response
+
+from dcos import config, http
+
+
+@patch('requests.request')
+def test_request_default_timeout_without_config(requests_mock):
+    timeout = True
+    toml_config = config.Toml({})
+    expected_timeout = http.DEFAULT_TIMEOUT
+
+    _assert_request_timeout(
+        requests_mock,
+        timeout,
+        toml_config,
+        expected_timeout)
+
+
+@patch('requests.request')
+def test_request_default_timeout_with_config(requests_mock):
+    timeout = 5
+
+    toml_config = config.Toml({
+        'core': {'timeout': timeout}
+    })
+
+    expected_timeout = (http.DEFAULT_CONNECT_TIMEOUT, timeout)
+
+    _assert_request_timeout(
+        requests_mock,
+        timeout,
+        toml_config,
+        expected_timeout)
+
+
+@patch('requests.request')
+def test_request_timeout_with_numeric_argument(requests_mock):
+    timeout = 30
+    toml_config = config.Toml({})
+    expected_timeout = (http.DEFAULT_CONNECT_TIMEOUT, timeout)
+
+    _assert_request_timeout(
+        requests_mock,
+        timeout,
+        toml_config,
+        expected_timeout)
+
+
+@patch('requests.request')
+def test_request_timeout_with_tuple_argument(requests_mock):
+    timeout = (5, 30)
+    toml_config = config.Toml({})
+
+    _assert_request_timeout(requests_mock, timeout, toml_config, timeout)
+
+
+def _assert_request_timeout(
+        requests_mock, timeout, toml_config, expected_timeout):
+    resp = Response()
+    resp.status_code = 200
+    requests_mock.return_value = resp
+
+    method = 'GET'
+    url = 'https://www.example.com'
+
+    http.request(
+        method,
+        url,
+        timeout=timeout,
+        headers={},
+        toml_config=toml_config)
+
+    requests_mock.assert_called_once_with(
+        method=method,
+        url=url,
+        auth=None,
+        timeout=expected_timeout,
+        headers={},
+        verify=None)


bamarni:
run all integration tests

dcos/http.py
```diff
@@ -13,7 +13,14 @@
 
 logger = util.get_logger(__name__)
 
-DEFAULT_TIMEOUT = 5
+DEFAULT_TIMEOUT = 15
```
adamtheturtle: Let's make sure that the public documentation matches this: https://jira.mesosphere.com/browse/DCOS_OSS-1957.

dcos/http.py
```diff
@@ -93,7 +100,7 @@ def _verify_ssl(url, verify=None, toml_config=None):
 def _request(method,
              url,
              is_success=_default_is_success,
-             timeout=DEFAULT_TIMEOUT,
+             timeout=-1,
```
adamtheturtle: Please document this in the docstring.

adamtheturtle: This `-1` system is not expected behaviour as a user of this API.

What do you think of instead making `timeout` a required parameter? Perhaps there could be a helper `http_timeout_from_configuration` and some callers would use that.

dcos/cluster.py
```diff
@@ -138,7 +138,9 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     cluster_name = cluster_id
     try:
         url = dcos_url.rstrip('/') + '/mesos/state-summary'
-        name_query = http.get(url, toml_config=cluster.get_config())
+        name_query = http.get(url,
```
adamtheturtle: This changes the default timeout for this operation from 5 seconds to 3. Is that desired?

dcos/cluster.py
```diff
@@ -381,7 +383,7 @@ def get_dcos_version(self):
             resp = requests.request(
                 'GET',
                 url=endpoint,
-                timeout=http.DEFAULT_TIMEOUT,
+                timeout=http.FAIL_FAST_TIMEOUT,
```
adamtheturtle: This changes the default timeout for this operation from 5 seconds to 3. Is that desired?

dcos/cluster.py
```diff
@@ -106,7 +106,7 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     try:
         # find cluster id
         cluster_url = dcos_url.rstrip('/') + '/metadata'
-        res = http.get(cluster_url, timeout=1)
+        res = http.get(cluster_url, timeout=http.FAIL_FAST_TIMEOUT)
```
adamtheturtle: This changes the default timeout for this operation from 1 second to 3. Is that desired?

It also is not clear from the public documentation what will and what will not be affected by the user-set timeout configuration variable.

dcos/http.py
adamtheturtle: Why 15 seconds?

I built a binary from this branch and tested it with the issue I presented on https://jira.mesosphere.com/browse/DCOS_OSS-1988. The problem remained:

```
Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --log-level=debug --debug package install --cli dcos-enterprise-cli --yes
MainThread: 2017-12-18 19:36:04,525 dcos/http.py:_request:155 - Sending HTTP ['get'] to ['https://172.17.0.2/capabilities']: {'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Accept': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1'}
MainThread: 2017-12-18 19:36:04,531 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAccept-Encoding: gzip, deflate\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
MainThread: 2017-12-18 19:36:04,691 urllib3/connectionpool.py:_make_request:396 - https://172.17.0.2:443 "GET /capabilities HTTP/1.1" 200 144
MainThread: 2017-12-18 19:36:04,695 dcos/http.py:_request:185 - Received HTTP response [200]: {'Server': 'openresty/1.13.6.1', 'Date': 'Mon, 18 Dec 2017 19:36:34 GMT', 'Content-Length': '144', 'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Connection': 'keep-alive'}
MainThread: 2017-12-18 19:36:04,695 dcos/util.py:timer:609 - duration: dcos.http._request: 0.17s
MainThread: 2017-12-18 19:36:04,702 dcos/http.py:_request:155 - Sending HTTP ['post'] to ['https://172.17.0.2/package/describe']: {'Content-Type': 'application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1', 'Accept': 'application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3'}
MainThread: 2017-12-18 19:36:04,705 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nAccept-Encoding: gzip, deflate\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
reply: 'HTTP/1.1 200 OK\r\n'
MainThread: 2017-12-18 19:36:15,758 urllib3/connectionpool.py:_make_request:396 - https://172.17.0.2:443 "POST /package/describe HTTP/1.1" 200 1346
MainThread: 2017-12-18 19:36:15,762 dcos/http.py:_request:185 - Received HTTP response [200]: {'Server': 'openresty/1.13.6.1', 'Date': 'Mon, 18 Dec 2017 19:36:45 GMT', 'Content-Length': '1346', 'Content-Type': 'application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3', 'Connection': 'keep-alive'}
MainThread: 2017-12-18 19:36:15,762 dcos/util.py:timer:609 - duration: dcos.http._request: 11.06s
header: Server header: Date header: Content-Type header: Content-Length header: Connection By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#certified-services
Installing CLI subcommand for package [dcos-enterprise-cli] version [1.3.0]
MainThread: 2017-12-18 19:36:15,773 dcos/util.py:ensure_dir_exists:203 - Creating directory: '/Users/Adam/.dcos/clusters/edfc460f-77b4-476e-993a-c0910aa5f3e9/subcommands/dcos-enterprise-cli'
MainThread: 2017-12-18 19:36:15,800 dcos/http.py:_request:155 - Sending HTTP ['get'] to ['https://172.17.0.2/package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli']: {'Accept': 'application/json'}
MainThread: 2017-12-18 19:36:15,804 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
send: b'GET /package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli HTTP/1.1\r\nHost: 172.17.0.2\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nAccept: application/json\r\nAccept-Encoding: gzip, deflate\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
MainThread: 2017-12-18 19:36:30,916 dcos/http.py:_request:177 - HTTP Timeout
  File "urllib3/connectionpool.py", line 387, in _make_request
  File "<string>", line 2, in raise_from
  File "urllib3/connectionpool.py", line 383, in _make_request
  File "http/client.py", line 1198, in getresponse
  File "http/client.py", line 297, in begin
  File "http/client.py", line 258, in _read_status
  File "socket.py", line 576, in readinto
  File "ssl.py", line 937, in recv_into
  File "ssl.py", line 799, in read
  File "ssl.py", line 583, in read
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

  File "requests/adapters.py", line 440, in send
  File "urllib3/connectionpool.py", line 639, in urlopen
  File "urllib3/util/retry.py", line 357, in increment
  File "urllib3/packages/six.py", line 686, in reraise
  File "urllib3/connectionpool.py", line 601, in urlopen
  File "urllib3/connectionpool.py", line 389, in _make_request
  File "urllib3/connectionpool.py", line 309, in _raise_timeout
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)

During handling of the above exception, another exception occurred:

  File "dcos/http.py", line 164, in _request
  File "requests/api.py", line 58, in request
  File "requests/sessions.py", line 508, in request
  File "requests/sessions.py", line 618, in send
  File "requests/adapters.py", line 521, in send
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)
MainThread: 2017-12-18 19:36:30,918 dcos/util.py:timer:609 - duration: dcos.http._request: 15.12s
Request to URL [https://172.17.0.2/package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli] timed out.
       27.78 real         0.84 user         0.62 sys
```

That isn't to say that this change is not part of a meaningful change, but I'd like to know of the reason for changing from 5 to 15 - are there requests which take somewhere between those values?

dcos/http.py
bamarni: Looking closer at the underlying library (requests), their timeout can actually be splitted in a tuple : 
```
 :param timeout: (optional) How many seconds to wait for the server to send data
                            before giving up, as a float, or a :ref:`(connect timeout, read
                            timeout) <timeouts>` tuple.
```

I think we need to leverage this, the `FAIL_FAST_TIMEOUT` would become the connection timeout, `DEFAULT_TIMEOUT` would become the timeout for reading the response. Then I am inclined to bump the default timeout way more (even to the documented 180 seconds).

The timeout being a required parameter would make the http module way less user-friendly. I agree though that `-1` seems odd. We need to have the default being read from configuration or falling back to the constant, I can suggest instead of `-1` to use a default of `True`, which makes more sense imo.

WDYT?

adamtheturtle:
Thank you Bilal.
Unfortunately, if my testing is valid, this does not address the problem fully.

The following is an attempt to set the timeout to 600, and then you can see that there is still an error when installing the Enterprise CLI at a timeout of 15 seconds.

```
Adam@MacBook-Pro ~/D/m/m/dcos-cli> git branch
* (HEAD detached at bilal/DCOS_OSS-1956)
Adam@MacBook-Pro ~/D/m/m/d/cli> rm -rf dist/
Adam@MacBook-Pro ~/D/m/m/d/cli> make binary
...
Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos config set core.timeout 60
[core.timeout]: already set to '60'
Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos --log-level=debug --debug package install --cli dcos-enterprise-cli --yes
MainThread: 2017-12-18 20:53:05,680 dcos/http.py:_request:155 - Sending HTTP ['get'] to ['https://172.17.0.2/capabilities']: {'Accept': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1'}
MainThread: 2017-12-18 20:53:05,687 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
MainThread: 2017-12-18 20:53:05,858 urllib3/connectionpool.py:_make_request:396 - https://172.17.0.2:443 "GET /capabilities HTTP/1.1" 200 144
MainThread: 2017-12-18 20:53:05,863 dcos/http.py:_request:185 - Received HTTP response [200]: {'Date': 'Mon, 18 Dec 2017 20:53:40 GMT', 'Connection': 'keep-alive', 'Server': 'openresty/1.13.6.1', 'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Content-Length': '144'}
MainThread: 2017-12-18 20:53:05,863 dcos/util.py:timer:609 - duration: dcos.http._request: 0.18s
MainThread: 2017-12-18 20:53:05,875 dcos/http.py:_request:155 - Sending HTTP ['post'] to ['https://172.17.0.2/package/describe']: {'Accept': 'application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3', 'Content-Type': 'application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1'}
MainThread: 2017-12-18 20:53:05,877 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
MainThread: 2017-12-18 20:53:21,010 dcos/http.py:_request:177 - HTTP Timeout
  File "urllib3/connectionpool.py", line 387, in _make_request
  File "<string>", line 2, in raise_from
  File "urllib3/connectionpool.py", line 383, in _make_request
  File "http/client.py", line 1198, in getresponse
  File "http/client.py", line 297, in begin
  File "http/client.py", line 258, in _read_status
  File "socket.py", line 576, in readinto
  File "ssl.py", line 937, in recv_into
  File "ssl.py", line 799, in read
  File "ssl.py", line 583, in read
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

  File "requests/adapters.py", line 440, in send
  File "urllib3/connectionpool.py", line 639, in urlopen
  File "urllib3/util/retry.py", line 357, in increment
  File "urllib3/packages/six.py", line 686, in reraise
  File "urllib3/connectionpool.py", line 601, in urlopen
  File "urllib3/connectionpool.py", line 389, in _make_request
  File "urllib3/connectionpool.py", line 309, in _raise_timeout
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)

During handling of the above exception, another exception occurred:

  File "dcos/http.py", line 164, in _request
  File "requests/api.py", line 58, in request
  File "requests/sessions.py", line 508, in request
  File "requests/sessions.py", line 618, in send
  File "requests/adapters.py", line 521, in send
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)
MainThread: 2017-12-18 20:53:21,013 dcos/util.py:timer:609 - duration: dcos.http._request: 15.14s
Request to URL [https://172.17.0.2/package/describe] timed out.
```

adamtheturtle:
Thank you Bilal.
Unfortunately, if my testing is valid, this does not address the problem fully.

The following is an attempt to set the timeout to 600, and then you can see that there is still an error when installing the Enterprise CLI at a timeout of 15 seconds.

```
Adam@MacBook-Pro ~/D/m/m/dcos-cli> git branch
* (HEAD detached at bilal/DCOS_OSS-1956)
Adam@MacBook-Pro ~/D/m/m/d/cli> rm -rf dist/
Adam@MacBook-Pro ~/D/m/m/d/cli> make binary
...
Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos config set core.timeout 60
[core.timeout]: already set to '60'
Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos --log-level=debug --debug package install --cli dcos-enterprise-cli --yes
MainThread: 2017-12-18 20:53:05,680 dcos/http.py:_request:155 - Sending HTTP ['get'] to ['https://172.17.0.2/capabilities']: {'Accept': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1'}
MainThread: 2017-12-18 20:53:05,687 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
MainThread: 2017-12-18 20:53:05,858 urllib3/connectionpool.py:_make_request:396 - https://172.17.0.2:443 "GET /capabilities HTTP/1.1" 200 144
MainThread: 2017-12-18 20:53:05,863 dcos/http.py:_request:185 - Received HTTP response [200]: {'Date': 'Mon, 18 Dec 2017 20:53:40 GMT', 'Connection': 'keep-alive', 'Server': 'openresty/1.13.6.1', 'Content-Type': 'application/vnd.dcos.capabilities+json;charset=utf-8;version=v1', 'Content-Length': '144'}
MainThread: 2017-12-18 20:53:05,863 dcos/util.py:timer:609 - duration: dcos.http._request: 0.18s
MainThread: 2017-12-18 20:53:05,875 dcos/http.py:_request:155 - Sending HTTP ['post'] to ['https://172.17.0.2/package/describe']: {'Accept': 'application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3', 'Content-Type': 'application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1'}
MainThread: 2017-12-18 20:53:05,877 urllib3/connectionpool.py:_new_conn:824 - Starting new HTTPS connection (1): 172.17.0.2
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDA1NDc4MX0.IBOY8X8Bc7UmzalNMU51MowtM25JOhr8CPopcaAgqnAbKEJPslpMplPE_rZep39qGqj_ZueZCMKEkhTGAkgMsDpqIBgm1Wx-p6Cy_rg6BpVqVg3q-RBvnQyc1D4hUBN6BM5RLoPrWo1wwHGveuQ-RRE9hodXkr1f4U9kId2lQLtxYDN6DSeyLTI78aLsrFS7gGdNxSshTkAemAAdpsoQVgPg6MkM2aohcVu9XR9pGcGvpek266K7zDUMZWg0Gz0ocps_0c1JaQYhDaDIpfWaaM0lzQUoTEU3Q4Y04KyXJM4rr1qEhi9atFA26gfIRY2V00vYs-lWMRRxbCFEsqzRKQ\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
MainThread: 2017-12-18 20:53:21,010 dcos/http.py:_request:177 - HTTP Timeout
  File "urllib3/connectionpool.py", line 387, in _make_request
  File "<string>", line 2, in raise_from
  File "urllib3/connectionpool.py", line 383, in _make_request
  File "http/client.py", line 1198, in getresponse
  File "http/client.py", line 297, in begin
  File "http/client.py", line 258, in _read_status
  File "socket.py", line 576, in readinto
  File "ssl.py", line 937, in recv_into
  File "ssl.py", line 799, in read
  File "ssl.py", line 583, in read
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

  File "requests/adapters.py", line 440, in send
  File "urllib3/connectionpool.py", line 639, in urlopen
  File "urllib3/util/retry.py", line 357, in increment
  File "urllib3/packages/six.py", line 686, in reraise
  File "urllib3/connectionpool.py", line 601, in urlopen
  File "urllib3/connectionpool.py", line 389, in _make_request
  File "urllib3/connectionpool.py", line 309, in _raise_timeout
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)

During handling of the above exception, another exception occurred:

  File "dcos/http.py", line 164, in _request
  File "requests/api.py", line 58, in request
  File "requests/sessions.py", line 508, in request
  File "requests/sessions.py", line 618, in send
  File "requests/adapters.py", line 521, in send
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='172.17.0.2', port=443): Read timed out. (read timeout=15)
MainThread: 2017-12-18 20:53:21,013 dcos/util.py:timer:609 - duration: dcos.http._request: 15.14s
Request to URL [https://172.17.0.2/package/describe] timed out.
```

bamarni:
@adamtheturtle : my bad this was quite messy, I've adjusted the PR to match with the doc and introduced a connection timeout. This should fix your problem.

dcos/http.py
```diff
@@ -106,8 +117,10 @@ def _request(method,
     :type url: str
     :param is_success: Defines successful status codes for the request
     :type is_success: Function from int to bool
-    :param timeout: request timeout
-    :type timeout: int
+    :param timeout: How many seconds to wait for the server to send data
```
adamtheturtle: It is not clear from this what `True` means.

dcos/http.py
```diff
@@ -106,8 +117,10 @@ def _request(method,
...
+                    before giving up, as a float, a (connect timeout, read
```
adamtheturtle: I'd prefer a more precise definition of "giving up".

dcos/http.py
```diff
@@ -120,6 +133,15 @@ def _request(method,
     :rtype: Response
     """
 
+    if timeout is True:
+        if toml_config is None:
+            toml_config = config.get_config()
+
+        timeout = config.get_config_val("core.timeout", toml_config)
+        timeout = (DEFAULT_CONNECT_TIMEOUT, timeout or DEFAULT_TIMEOUT)
```
adamtheturtle: If `timeout` is `0` or `None`, I believe that this will set `timeout` to `(DEFAULT_CONNECT_TIMEOUT, (DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT))`.

That is to say, I think that instead of `DEFAULT_TIMEOUT`, this should say `DEFAULT_READ_TIMEOUT`.

This indicates that there is a missing test.

It is not clear from the public documentation that giving a value of `0` uses the default timeout.

dcos/http.py
```diff
@@ -13,7 +13,18 @@
 
 logger = util.get_logger(__name__)
 
-DEFAULT_TIMEOUT = 5
+DEFAULT_READ_TIMEOUT = 180
```
adamtheturtle: It is probably useful to link to http://docs.python-requests.org/en/master/user/advanced/#timeouts somewhere.

dcos/http.py
adamtheturtle: This is quite a complicated parameter now.

What do you think of this:

```python
def _request(
    ...
    timeout_from_config=True,
    custom_timeout=None,
):
    if timeout_from_config and custom_timeout is not None:
        raise Exception()
```

?

adamtheturtle: And perhaps that logic can be done in `request` and this function can be kept basically the same as it was before.

dcos/http.py
bamarni: hum.. I'll fix this and add some tests :+1:

dcos/http.py
bamarni: Indeed the timeout can accept different kind of values, this comes from the requests package, which can accept None, a float or a tuple. Adding a default to True doesn't impact the usage (eg. this will never be needed to do `request(timeout=True)`, rather this is more of an internal way to specify a default (because the default we want has complex logic).

Having to do `request(timeout_from_config=False, custom_timeout=30)` in order to specify a timeout seems very unnatural to me.

dcos/http.py
```diff
@@ -106,8 +117,10 @@ def _request(method,
     :type url: str
     :param is_success: Defines successful status codes for the request
     :type is_success: Function from int to bool
-    :param timeout: request timeout
-    :type timeout: int
+    :param timeout: How many seconds to wait for the server to send data
+                    before giving up, as a float, a (connect timeout, read
+                    timeout) tuple or `True`. `None` means unlimited timeout.
+    :type timeout: True | None | float | (None | float, None | float)
```
adamtheturtle: `True` is not a `type`. I believe that this should be `bool`, but I am not hugely familiar with this documentation style.

dcos/http.py
adamtheturtle: This is an implementation detail and so I will leave the final decision on the internal API to you and the other core maintainers of this project. However, I do feel that using `True` on top of the `requests` API is hard to understand.

I'll restate the choice that I would make, but I understand that this is not your preference:

`timeout` is a required field.
Consumers can use a helper to get a value from a configuration, or a default.
That might look like:

```python
timeout = helpers.get_timeout_from_config(config=toml_config)
request(timeout=timeout, ...)
```

dcos/http.py
adamtheturtle: It is up to you whether to keep this, but this PR now does two things:
This PR now does at least the following:

1. It uses the configuration variable in more situations.
2. It changes some requests' defaults to 180 from 5 seconds.

If instead, this were limited to just the former, as per https://jira.mesosphere.com/browse/DCOS_OSS-1956, this could be a much simpler change.

I would personally prefer that. Then, changing the default timeout can be done in another PR which can maybe involve @jgehrcke as a reviewer, and a more thorough look into the optimal timeout. That could even look into removing potentially duplicated logic such as https://github.com/dcos/dcos-cli/blob/3402f831528bca6985eaad1f8308c6a911e6036f/cli/dcoscli/node/main.py#L290.

adamtheturtle: What I mean is, consider setting `DEFAULT_READ_TIMEOUT` to `5` here. I will still have to use the configuration option, but that will unblock me.

adamtheturtle:
Thanks @bamarni - let me know when you'd like another review.

dcos/http.py
bamarni: I am also missing the exact documentation of the pydoc format we use, and work mostly by comparing the existing we have...

As we currently have `None` in some docs, and it is not a type either, I assume `True` is also valid.

adamtheturtle: That is not a safe assumption. For example, `None` is shorthand for `type(None)` in `mypy`. `True`'s type is `bool` and that is the only way I have seen its type documented anywhere.

bamarni:
@adamtheturtle : I've updated the PR and added tests. Only fixing the code to use the `core.timeout` config value is not enough, I'm seeing more and more reports of timeout errors (mainly in the context of cosmos) lately. From DC/OS 1.11, 5 seconds is not enough anymore, therefore the default has to be bumped.

As this PR now makes the distinction between the connection and read timeout, I'm fine with using the documented 180 seconds.

dcos/http.py
bamarni: Indeed, the place you've linked could be simplified too. I also would have prefer to have the default read from the config only in `http.request` instead of as deep as `http._request`, but as the `auth` module uses that private function directly I put it there.

Otherwise the changes in this PR are in my view the bare minimum we must do in order to fix the current situation.

dcos/http.py
```diff
@@ -106,8 +117,14 @@ def _request(method,
...
-    :param timeout: request timeout
-    :type timeout: int
+    :param timeout: How many seconds to wait for the server to send data
+                    before aborting and raising an exception. A timeout is
+                    either a float, a (connect timeout, read timeout) tuple
+                    or `None`. `None` means unlimited timeout. If no timeout
+                    is passed it defaults to the DEFAULT_TIMEOUT tuple, where
+                    the connect timeout can optionally be overridden by the
+                    `core.timeout` config.
+    :type timeout: float | None | bool | (None | float, None | float)
```
adamtheturtle: These `float`s can also be `int`s, right?

dcos/http.py
```diff
@@ -120,6 +137,15 @@ def _request(method,
     :rtype: Response
     """
 
+    if timeout is True:
+        if toml_config is None:
+            toml_config = config.get_config()
+
+        timeout = config.get_config_val("core.timeout", toml_config)
+        timeout = (DEFAULT_CONNECT_TIMEOUT, timeout or DEFAULT_READ_TIMEOUT)
```
adamtheturtle: This sets the read timeout to `DEFAULT_READ_TIMEOUT` if `timeout` is set to 0. Is this the desired behaviour? Please add a test for this case.

dcos/cluster.py
```diff
@@ -106,7 +106,7 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     try:
         # find cluster id
         cluster_url = dcos_url.rstrip('/') + '/metadata'
-        res = http.get(cluster_url, timeout=1)
+        res = http.get(cluster_url, timeout=5)
```
adamtheturtle: Why change this?

Please add a comment describing how the timeout was chosen.

dcos/cluster.py
```diff
@@ -138,7 +138,9 @@ def setup_cluster_config(dcos_url, temp_path, stored_cert):
     cluster_name = cluster_id
     try:
         url = dcos_url.rstrip('/') + '/mesos/state-summary'
-        name_query = http.get(url, toml_config=cluster.get_config())
+        name_query = http.get(url,
+                              toml_config=cluster.get_config(),
+                              timeout=5)
```
adamtheturtle: Please add a comment describing how the timeout was chosen.

dcos/cluster.py
```diff
@@ -381,7 +383,7 @@ def get_dcos_version(self):
             resp = requests.request(
                 'GET',
                 url=endpoint,
-                timeout=http.DEFAULT_TIMEOUT,
+                timeout=5,
```
adamtheturtle: Please add a comment describing how the timeout was chosen.

adamtheturtle:
Thank you @bamarni - I want to do two tests with the built binary before approving this:

1. Make sure that I can install the Enterprise CLI
2. Set the default timeout to 1 second and see the timeout error that I saw before

Otherwise, I'm happy. We've discussed some code organisation, but I am a consumer and will leave it to you to make the decisions about how to organise the code you are maintaining.

Please can you link to the JIRA issue(s) that you propose this closes, so that reviewers can assess whether this is a comprehensive fix.

Other than that, this changes the default read timeout to 180 seconds. I'd like @jgehrcke to have the opportunity to comment on that. I really don't have well informed opinions on that matter.

jgehrcke:
> Other than that, this changes the default read timeout to 180 seconds. I'd like @jgehrcke to have the opportunity to comment on that. I really don't have well informed opinions on that matter.

I assume it changes the read timeout to 180 seconds from a rather small value. This is a good choice.

Here is how I see what matters in this context in practice:

### The connect timeout
In practice, a TCP connection attempt should rarely take longer than a few seconds. Even under extremely bad transport layer conditions. Therefore, a very long `connect()` timeout rarely makes sense for a program that a human user interacts with. If that program could not `connect()` to the remote end within `5 s` to `10 s` then the user deserves to see a corresponding error message and might want to retry manually.

### The receive/read timout
In practice, after a TCP connection has been established and after an HTTP request has been sent through it, it _might_ take the remote system a long while to process the HTTP request and to generate a corresponding HTTP response. The time between receiving the request and emitting the response cannot in general be assumed to be limited by the network, but by anything (think of a compute job that takes `10 s` or `1 min` to complete before a response can be emitted). Before the response is generated the client hangs in a `recv()` system call and the receive/read timeout applies. Of course, the goal is to provide an error message to the user as early as possible, whenever it makes sense. However, an even more important goal is to *not* abort the `recv()` prematurely, e.g. 1 second before the HTTP response _would have arrived_ (I think that's the case that you, Adam, ran into). That is why for setting a reasonable receive timeout ideally we understand what the _conceptual_ upper bound is for the duration of the operation in question. If we don't know that (as seems to be the case for some operations performed by Cosmos) the only reasonable option then is to set a large receive timeout to make sure that the client waits "long enough". I think that's what "this changes the default read timeout to 180 seconds" achieves here. It's a change pro reliability, and not so much for UX. When we have to pick one, reliability is more important than UX, so that's the correct first step.

adamtheturtle:
Thanks @jgehrcke . @bamarni - that confirms to me that is a good change. Please make the minor changes I suggested and then I will do a manual test as described above. Then I expect to approve this PR.

bamarni:
@adamtheturtle : about the 3 places with the 5 seconds timeout you've commented on, they correspond to
- `/metadata` : used when setting-up a cluster
- `/dcos-metadata/dcos-version.json`
- `/mesos/state-summary`

These requests are either made when setting-up a cluster or listing clusters (to have their version in the list). Historically the timeout has been set to something low because they might not be reachable and we want to fail early in such cases (eg. it means the user made a typo when doing `cluster setup`, or that a cluster is down and the version shown in `cluster list` should be `N/A`).

So here what matters the most here is the connection timeout which has to be low, in order to fail fast when the connection cannot be established (this is something expected and valid for the use-cases I mentioned before). As for reading the responses, they are pretty small and delivered fast from my experience, there were already a small timeout there (even 1sec). It is very unlikely that they fail because of the read timeout (maybe the mesos state summary is a little slower than the others, but it is already using a 5 seconds timeout currently).

Thanks @jgehrcke for the detailed explanation, I also agree that the connect timeout should be pretty low and the read timeout higher. In the future we might adjust the read timeout on a component per component basis.

adamtheturtle:
@bamarni that's a great explanation and I would like it or similar to be in the codebase.

When the next maintainer wants to change it, they should know the context.

dcos/http.py
bamarni: `timeout` to 0 doesn't really makes sense, it is not possible to set it like this through the config : 
```
$ dcos config set core.timeout 0                                                                      
Error: 0 is less than the minimum of 1
Path: timeout
Value: 0
```

adamtheturtle: Great, that satisfies me.

bamarni:
I've added comments on top of the 3 places.

adamtheturtle:
Thank you Bilal.

I ran the command which had been failing a few times.
With the CLI `master`, it failed every time.
With this change, it failed a few times on a different timeout error.

I think that this means that your change works - more testing to come - but another issue has been uncovered.
Let's try to understand before merging this.

Below are the logs for two attempts with a binary built from this branch.

The first is a failure:

```
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
reply: 'HTTP/1.1 504 Gateway Time-out\r\n'
Error while fetching [https://172.17.0.2/package/describe]: HTTP 504: "Gateway Time-out".
header: Server header: Date header: Content-Type header: Content-Length header: Connection ⏎                                                                                                    (e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
reply: 'HTTP/1.1 504 Gateway Time-out\r\n'
Error while fetching [https://172.17.0.2/package/describe]: HTTP 504: "Gateway Time-out".
header: Server header: Date header: Content-Type header: Content-Length header: Connection        62.33 real         0.85 user         0.71 sys
```

The second is a success:

```
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#certified-services
Installing CLI subcommand for package [dcos-enterprise-cli] version [1.3.0]
send: b'GET /package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli HTTP/1.1\r\nHost: 172.17.0.2\r\nUser-Agent: python-requests/2.18.4\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/json\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection header: Content-Disposition New commands available: dcos backup, dcos security
```

This is compared to the kinds of failures I see with `master`:

```
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nUser-Agent: python-requests/2.18.4\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nUser-Agent: python-requests/2.18.4\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
Request to URL [https://172.17.0.2/package/describe] timed out.
        7.12 real         0.93 user         0.77 sys
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#certified-services
Installing CLI subcommand for package [dcos-enterprise-cli] version [1.3.0]
send: b'GET /package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/json\r\nConnection: keep-alive\r\nUser-Agent: python-requests/2.18.4\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
Request to URL [https://172.17.0.2/package/resource?url=https://downloads.mesosphere.io/cli/binaries/darwin/x86-64/1.3.0/da2e75e78079662dd9940c7e848770ad8e55c97124a57a2d5786a679e6dfe6d0/dcos-enterprise-cli] timed out.
        7.36 real         1.01 user         0.82 sys
```

It looks as though the `core.timeout` does apply to this new work as hoped:

```
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> dist/dcos config set core.timeout 2
[core.timeout]: already set to '2'
(e2e) Adam@MacBook-Pro ~/D/m/m/d/cli> time dist/dcos --debug package install --cli dcos-enterprise-cli --yes
send: b'GET /capabilities HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nUser-Agent: python-requests/2.18.4\r\nContent-Type: application/vnd.dcos.capabilities+json;charset=utf-8;version=v1\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server header: Date header: Content-Type header: Content-Length header: Connection send: b'POST /package/describe HTTP/1.1\r\nHost: 172.17.0.2\r\nAccept-Encoding: gzip, deflate\r\nConnection: keep-alive\r\nAccept: application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3\r\nUser-Agent: python-requests/2.18.4\r\nContent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1\r\nContent-Length: 38\r\nAuthorization: token=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJ1aWQiOiJhZG1pbiIsImV4cCI6MTUxNDIxMTg5OH0.EhTiqJMAu_DfTv0TkXw0l0pemEBqPtEEx8AS4lwHKGHNWAmvng7Z1_xYVL1kac49As5wRTmomyorrEJ0LkrmtBvbZcIgYzLTmDpaYqFPK8FSLySkwF8ghi3EKiISicCIC473Ucyejv4hMAFNXkU17BZZkiUkO7B_CSwEdt2DeAZ_YruoF_5YowE9FiwGi89qt0_H27ywjgT4eyVIg6frh9J2KBwwBZk-rjZRN0xhhS88k1uddlpafMLTiGhfcdTUoo8KAglHdI3HsB77KjCpRjHVGDIBvPJlRbshNcGrAQ3DdvFrsbS4GtNsE3D2BMX9na7NxuR2nsIJVXuA8q10Lw\r\n\r\n'
send: b'{"packageName": "dcos-enterprise-cli"}'
Request to URL [https://172.17.0.2/package/describe] timed out.
        3.90 real         0.88 user         0.73 sys
```

bamarni:
@adamtheturtle : the 504 timeout error seems like a cluster-side problem, the CLI receives a successful error response there.

It seems like admin router is timing-out while it doesn't get any data to read from cosmos.

bamarni:
run all integration tests

bamarni:
Thanks for the reviews @adamtheturtle @jgehrcke @mnaboka

PR #1128, opened 2017-12-18
fix dcos-cli CI

mnaboka:
installer would not pass exhibitor_storage_backend implicitly anymore, we
need to include it in dcos-launch config.


mnaboka:
run all integration tests

PR #1131, opened 2018-01-10
Improve output from metrics summary

philipnrmn:
Previously, as an outcome of `dcos node metrics summary`, the CLI reported the percentage of the disk that was available, eg 1GB (99%) for a 100GB disk.

This PR updates that calculation so we now see the used percentage, so in the above case the user would now see a disk entry of 1GB (1%). 

Additionally, I expanded the output from the --json endpoint after noticing that it only reported the total system memory and disk usage.

### DIFF
--- a/cli/dcoscli/metrics.py
+++ b/cli/dcoscli/metrics.py
@@ -98,6 +98,8 @@ def _node_summary_json(datapoints):
     summary_datapoints = [
         _get_datapoint(datapoints, 'cpu.total'),
         _get_datapoint(datapoints, 'memory.total'),
+        _get_datapoint(datapoints, 'memory.free'),
+        _get_datapoint(datapoints, 'filesystem.capacity.total', {'path': '/'}),
         _get_datapoint(datapoints, 'filesystem.capacity.used', {'path': '/'})
     ]
     return json.dumps(summary_datapoints)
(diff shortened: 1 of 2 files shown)


cli/dcoscli/metrics.py
```diff
@@ -125,7 +127,7 @@ def _node_summary_data(datapoints):
     disk_free = _get_datapoint_value(
         datapoints, 'filesystem.capacity.used', {'path': '/'})
```
bamarni: the variable is called disk_free while it seems to be disk_used (`filesystem.capacity.used`).

cli/dcoscli/metrics.py
```diff
@@ -98,6 +98,8 @@ def _node_summary_json(datapoints):
     summary_datapoints = [
         _get_datapoint(datapoints, 'cpu.total'),
         _get_datapoint(datapoints, 'memory.total'),
+        _get_datapoint(datapoints, 'memory.free'),
```
bamarni: here for consistency, could we have `memory.used`? A few lines below there is `filesystem.capacity.used`. So that'd make datapoint pairs of `used` and `total`.

philipnrmn: Hey @bamarni I do agree that this would make more sense! Unfortunately `memory.used` isn't available in our stats - we only have `memory.free`.

cli/dcoscli/metrics.py
philipnrmn: Fixed :)

klueska:
@bamarni any other comments or can we start the CI on this one?

bamarni:
run linux integration tests

PR #1132, opened 2018-01-11
Catch exceptions when matching configured clusters hosts

bamarni:
During cluster linking, the host to be linked gets its IP address
resolved in order to get matched against the currently configured ones.

Resolving configured clusters IP address might throw an exception (when
they are not reachable anymore), this continues silently in such cases.

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -257,7 +257,12 @@ def _link(dcos_url, provider_id):
     for configured_cluster in cluster.get_clusters():
         configured_cluster_host = \
             urlparse(configured_cluster.get_url()).netloc
-        configured_cluster_ip = socket.gethostbyname(configured_cluster_host)
+
+        try:
+            configured_cluster_ip = \
+                socket.gethostbyname(configured_cluster_host)
+        except OSError as error:
+            continue
 
         if linked_cluster_ip == configured_cluster_ip:
             linked_cluster_id = configured_cluster.get_cluster_id()


bamarni:
run all integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -257,7 +257,12 @@ def _link(dcos_url, provider_id):
     for configured_cluster in cluster.get_clusters():
         configured_cluster_host = \
             urlparse(configured_cluster.get_url()).netloc
-        configured_cluster_ip = socket.gethostbyname(configured_cluster_host)
+
+        try:
+            configured_cluster_ip = \
+                socket.gethostbyname(configured_cluster_host)
+        except OSError as error:
+            continue
```
klueska: What will `configured_cluster_ip` be after this point if we ignore the exception?
What is the desired behaviour overall?

bamarni: `continue` moves to the top of the loop with the next cluster item, `configured_cluster_ip` will be overriden in this next iteration.

The goal of this loop is to find if the cluster to be linked is already configured (by resolving  matching the master ip addresses), which is at the moment a requirement since we don't support setting-it up on the fly.

klueska: OK. I'm not a huge fan of how this is all put together with a `break` and an `else` on the for loop, but I can at least understand what you are trying to do now.

bamarni: How would you have rather wrote it? IMO the for else clause is mainly useful in this kind of situation, to avoid making a boolean flag to track if there is an item found.

PR #1133, opened 2018-01-18
Adding Metronome job queue functionality

kensipe:
command:   `job queue`, `job queue startdeadline`, `job queue --json`, `job queue -q`

```
dcos job queue 
ID                         JOB RUN ID              
startdeadline  /startdeadline/201801172321181BlLt  
```

```
 dcos job queue --json
[
  {
    "jobId": "startdeadline",
    "runs": [
      {
        "runId": "/startdeadline/201801172321181BlLt"
      }
    ]
  }
]
```

### DIFF
--- a/cli/tests/data/metronome/jobs/gyarados.json
+++ b/cli/tests/data/metronome/jobs/gyarados.json
@@ -1,0 +1,10 @@
+{
+  "id": "gyarados",
+  "description": "so large I do not fit",
+  "run": {
+    "cmd": "sleep 20000",
+    "cpus": 10,
+    "mem": 32,
+    "disk": 0
+  }
+}
--- a/cli/tests/unit/data/job_queue.txt
+++ b/cli/tests/unit/data/job_queue.txt
@@ -1,0 +1,3 @@
+ID              JOB RUN ID      
+queue-job  20180123185638P0Dxf  
+queue-job  20180123185640uv3PZ  
--- a/python/lib/dcos/dcos/metronome.py
+++ b/python/lib/dcos/dcos/metronome.py
@@ -339,7 +339,8 @@ class Client(object):
         :rtype: None
         """
         if job_id is None or run_id is None:
-            raise DCOSException("job_id and run_id are required to kill a jobrun.")
+            raise DCOSException(
+                "job_id and run_id are required to kill a jobrun.")
         job_id = util.normalize_marathon_id_path(job_id)
         run_id = util.normalize_marathon_id_path(run_id)
         path = '/v1/jobs{}/runs{}/actions/stop'.format(job_id, run_id)
--- a/python/lib/dcos/dcos/metronome.py
+++ b/python/lib/dcos/dcos/metronome.py
@@ -345,7 +345,7 @@ class Client(object):
         path = '/v1/jobs{}/runs{}/actions/stop'.format(job_id, run_id)
         self._rpc.http_req(http.post, path)
 
-    def get_queued_jobruns(self, job_id):
+    def get_queued_job_runs(self, job_id):
         """Returns the content of the launch queue,
         including the jobruns for each job_id which should be scheduled.
 
@@ -356,7 +356,7 @@ class Client(object):
         if job_id is not None:
             deployments = [
                 deployment for deployment in response.json()
-                if job_id in deployment['jobId']
+                if job_id == deployment['jobId']
             ]
         else:
             deployments = response.json()
--- a/python/lib/dcos/dcos/metronome.py
+++ b/python/lib/dcos/dcos/metronome.py
@@ -338,11 +338,30 @@ class Client(object):
         :type run_id: str
         :rtype: None
         """
-
+        if job_id is None or run_id is None:
+            raise DCOSException("job_id and run_id are required to kill a jobrun.")
         job_id = util.normalize_marathon_id_path(job_id)
         run_id = util.normalize_marathon_id_path(run_id)
         path = '/v1/jobs{}/runs{}/actions/stop'.format(job_id, run_id)
         self._rpc.http_req(http.post, path)
+
+    def get_queued_jobruns(self, job_id):
+        """Returns the content of the launch queue,
+        including the jobruns for each job_id which should be scheduled.
+
+        :returns: a list of jobruns to launch
+        :rtype: list of dict
+        """
+        response = self._rpc.http_req(http.get, 'v1/queue')
+        if job_id is not None:
+            deployments = [
+                deployment for deployment in response.json()
+                if job_id in deployment['jobId']
+            ]
+        else:
+            deployments = response.json()
+
+        return deployments
 
     @staticmethod
     def _job_id_path_format(url_path_template, id_path):
--- a/python/lib/dcoscli/dcoscli/data/help/job.txt
+++ b/python/lib/dcoscli/dcoscli/data/help/job.txt
@@ -18,6 +18,7 @@ Description:
         dcos job schedule remove <job-id> <schedule-id>
         dcos job schedule update <job-id> <schedule-file>
         dcos job show runs <job-id> [--run-id <run-id>][--json|--quiet]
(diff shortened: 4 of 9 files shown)


cli/dcoscli/job/main.py
```diff
@@ -228,6 +233,39 @@ def _remove(job_id, stop_current_job_runs=False):
     return 0
 
 
+def _queue(job_id, json_flag=False, quiet=False):
+    """
+    :param job_id: Id of the job
+    :type job_id: str
+    :param json_flag: output json if True
```
ichernetsky: It should start with a capital `O`, based on the style of other docstrings in this file.

cli/dcoscli/job/main.py
```diff
@@ -228,6 +233,39 @@ def _remove(job_id, stop_current_job_runs=False):
...
+    :type json_flag: bool
+    :returns: process return code
+    :rtype: int
+    """
+    client = metronome.create_client()
```
ichernetsky: There should be an empty line before this one.

cli/dcoscli/job/main.py
```diff
@@ -228,6 +233,39 @@ def _remove(job_id, stop_current_job_runs=False):
...
+    deployment_list = client.get_queued_jobruns(job_id)
+
+    if not deployment_list and not json_flag:
+        msg = "There are no deployments in the queue"
+        if job_id:
```
ichernetsky: Why an exception is thrown if `job_id` evaluates to `True`?

cli/dcoscli/job/main.py
```diff
@@ -228,6 +233,39 @@ def _remove(job_id, stop_current_job_runs=False):
...
+            msg += " for '{}'".format(job_id)
+            raise DCOSException(msg)
+        else:
```
ichernetsky: Why is just `msg` published and it returns? Why depends on `job_id`?

cli/dcoscli/tables.py
```diff
@@ -296,6 +296,32 @@ def job_table(job_list):
...
+    :type job_queue: [job]
+    :rtype: PrettyTable
+    """
+
+    flatten_job_list = []
+
+    for job in job_queue:
+        job_id = job.get("jobId")
+        for jobruns in job.get("runs"):
+            flatten_job_list.append({ 'jobId': job_id, 'jobrun_id': jobruns.get("runId")})
```
ichernetsky: Extra space after `{`.

dcos/metronome.py
```diff
@@ -338,12 +338,31 @@ def kill_run(self, job_id, run_id):
...
+        including the jobruns for each job_id which should be scheduled.
+
+        :returns: a list of jobruns to launch
+        :rtype: list of dict
+        """
+        response = self._rpc.http_req(http.get, 'v1/queue')
+        if job_id is not None:
+            deployments = [
+                deployment for deployment in response.json()
+                if job_id in deployment['jobId']
```
alenkacz: I believe that in on strings is actually doing substring so this would return true for "jobId" in "jobId2" so I would actually prefer to compare on equals here

dcos/metronome.py
```diff
@@ -338,12 +338,31 @@ def kill_run(self, job_id, run_id):
...
+            ]
+        else:
+            deployments = response.json()
+
+        return deployments
```
alenkacz: you don't have to assign to variable and then return, I would return directly from the if/else

dcos/metronome.py
```diff
@@ -338,12 +338,31 @@ def kill_run(self, job_id, run_id):
         :type run_id: str
         :rtype: None
         """
-
+        if job_id is None or run_id is None:
```
alenkacz: 👍

cli/dcoscli/tables.py
```diff
@@ -296,6 +296,32 @@ def job_table(job_list):
...
+
+    :param job_queue: jobs with jobruns in the queue to render
+    :type job_queue: [job]
+    :rtype: PrettyTable
+    """
+
+    flatten_job_list = []
+
+    for job in job_queue:
+        job_id = job.get("jobId")
```
alenkacz: can we do ['jobId'] instead of get?

cli/dcoscli/job/main.py
alenkacz: I am actually a little bit lost here too...

cli/dcoscli/job/main.py
```diff
@@ -228,6 +233,39 @@ def _remove(job_id, stop_current_job_runs=False):
     return 0
 
 
+def _queue(job_id, json_flag=False, quiet=False):
```
alenkacz: I don't understand what quiet means and it is not documented even in the other usages, so can you maybe document it?

cli/dcoscli/job/main.py
alenkacz: Maybe it is error when asking for concrete job id but not an error in general? But I would say it should not raise exception in both cases...

kensipe:
run linux integration tests

cli/dcoscli/job/main.py
kensipe: this is when the deployment_list doesn't exist... I considered an empty list to be normal however if you got an empty list and you specified a job_id then it is an error.   exit 0 is normal... errors are handled through exceptions in the CLI and will be transformed into a non-zero exit.    I believe this approach is consistent with previous marathon deployment code.

cli/dcoscli/job/main.py
kensipe: explanation provided above... the job_id is added the msg if possible to provide the most information

cli/dcoscli/job/main.py
kensipe: quiet or q is docker inspired... docker ps -lq returns just the sha of the last run container.  I've been adding these in a number of places in the CLI to make them more composable.. it is step 1... in other words you should be able to do this `dcos job kill jobID $(dcos job queue -q)` and it will kill all the jobs in the queue for the jobID.

dcos/metronome.py
kensipe: I would much rather have 1 return point...

kensipe:
run linux integration tests

kensipe:
run linux integration tests

kensipe:
run linux integration tests

cli/dcoscli/job/main.py
alenkacz: oh ok, so can we add a documentation for this somewhere? At least one sentence as it is in https://docs.docker.com/engine/reference/commandline/ps/

cli/dcoscli/job/main.py
alenkacz: I still don't agree that when I ask for a queue of specific job that it is error when that job is not in the queue - I would expect to get an empty result instead if it is not in the queue...

bamarni:
@kensipe : could you please add an integration test? Then we'd know that this functionality works and doesn't break in the future.

kensipe:
run linux integration tests

kensipe:
this is tested working.. however it requires the next release of metronome.   The failure `STDERR: Error: Unknown route.`  is because the test cluster doesn't have the latest metronome.

bamarni:
Thanks @kensipe, let's wait for it to be in dcos master then we can re-run the build and merge.

kensipe:
rebased

kensipe:
run linux integration tests

bamarni:
run linux integration tests

kensipe:
let me double check master on this :(

kensipe:
sorry on this... the PR to fix master didn't land yet... I will follow up there.

kensipe:
run linux integration tests

ArmandGrillet:
@kensipe Should this PR be closed and recreated for https://github.com/dcos/dcos-core-cli ?

bamarni:
@kensipe : please re-open the PR against the core CLI

kensipe:
👍

PR #1135, opened 2018-01-23
Consider that dcos version metadata might be invalid

bamarni:
I had an issue I cannot reproduce where the JSON returned by this endpoint was invalid, probably because of a wrongly installed / unhealthy cluster.

In such case it becomes impossible to run `dcos cluster list` for example as an exception is thrown, this change makes the code more resilient to this kind of problem.

### DIFF
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -391,10 +391,10 @@ class Cluster():
                 url=endpoint,
                 timeout=5,
                 verify=False)
+
+            return resp.json().get("version", VERSION_UNKNOWN)
         except Exception as e:
             return VERSION_UNKNOWN
-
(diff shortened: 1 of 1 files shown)


bamarni:
run linux integration tests

bamarni:
run linux integration tests

dcos/cluster.py
```diff
@@ -391,11 +391,11 @@ def get_dcos_version(self):
                 url=endpoint,
                 timeout=5,
                 verify=False)
+
+            return resp.json().get("version", VERSION_UNKNOWN)
```
mnaboka: would that be appropriate here to try catch jsonification here?

bamarni: what do you mean? it is currently in a catch-all block.

PR #1136, opened 2018-01-23
Cluster link error messages improvements

bamarni:
- Update a parameter pydoc description
- Don't require the cluster to be configured when unlinking
- Show an explicit error message when recreating an existing link
- Display a custom error message when unlinking an invalid cluster

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -327,7 +327,12 @@ def _unlink(name):
     endpoint = urllib.parse.urljoin(
             dcos_url, '/cluster/v1/links/' + name)
 
-    http.delete(endpoint)
+    try:
+        http.delete(endpoint)
+    except DCOSHTTPException as e:
+        if e.status() == 404:
+            raise DCOSException('Unknown cluster link {}.'.format(name))
+        raise
 
     return 0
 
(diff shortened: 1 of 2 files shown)


bamarni:
run all integration tests

bamarni:
ru all integration tests

bamarni:
run all integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -308,21 +313,26 @@ def _unlink(name):
...
     endpoint = urllib.parse.urljoin(
-            dcos_url, '/cluster/v1/links/' + c.get_cluster_id())
+            dcos_url, '/cluster/v1/links/' + name)
 
-    http.delete(endpoint)
+    try:
+        http.delete(endpoint)
+    except DCOSHTTPException as e:
+        if e.status() == 404:
+            raise DCOSException('Unknown cluster link {}.'.format(name))
```
collinvandyck: Would something like `"Cluster '{}' could not be found".format(name)` be better here?

bamarni: I wanted to add the `link` semantic to make it clear what it is about, if we adapt this message we should probably also align the other one with it. Currently it is : 
- `This cluster link already exists.`
- `Unknown cluster link <id_or_name>.`

cc @pavisandhu @klueska

PR #1137, opened 2018-01-23
 dcos-log v2: add a custom handler for 500 error

mnaboka:
- add 500 response code to is_success function to handle the 500 error
   in caller instead of `http` module.
- if the user encounters 500 when trying to fetch the logs, use
  `DCOSUnprocessableException` to include the text with the actual error
   message.

### DIFF
--- a/cli/dcoscli/log.py
+++ b/cli/dcoscli/log.py
@@ -13,6 +13,7 @@ from dcos.cosmos import get_cosmos_url
 from dcos.errors import (DCOSAuthenticationException,
                          DCOSAuthorizationException,
                          DCOSException,
+                         DCOSHTTPException,
                          DefaultError)
 
 logger = util.get_logger(__name__)
@@ -336,7 +337,7 @@ def is_success(code):
     :param code: HTTP response codes
     :type code: int
     """
-    if (200 <= code < 300) or code == 404:
+    if (200 <= code < 300) or code in [404, 500]:
         return True
     return False
 
@@ -351,6 +352,9 @@ def print_logs_range(url):
     with contextlib.closing(http.get(url, is_success=is_success,
                                      headers={'Accept': 'text/plain'})) as r:
 
+        if r.status_code == 500:
+            raise DCOSException(
+                '{} Error message: {}'.format(DCOSHTTPException(r), r.text))
         if r.status_code == 404:
             raise DCOSException('No files exist. Exiting.')
         if r.status_code == 204:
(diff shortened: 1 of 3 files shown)


mnaboka:
run all integration tests

mnaboka:
run all integration tests

cli/dcoscli/log.py
```diff
@@ -351,6 +352,8 @@ def print_logs_range(url):
     with contextlib.closing(http.get(url, is_success=is_success,
                                      headers={'Accept': 'text/plain'})) as r:
 
+        if r.status_code == 500:
+            raise DCOSUnprocessableException(r)
```
bamarni: The class pydoc states `A wrapper around Response objects for HTTP 422 error codes`, so this seems not suited for a 500 response.

What about adding a `text()` method to `DCOSHTTPException`, and if it's a 500 raise a `DCOSException` with the custom message you want (which would include the response content)?

mnaboka: what if i declare a new exception inherited from `DCOSHTTPException` and override `__str__` method to include text? would that work?

bamarni: Creating an exception class for 500 error seems fine, though I don't see why the string representation should vary depending on the class. Eg. why would a 500 error print the response content by default while a 502 wouldn't? That seems arbitrary.

I think we should expose the response content from DCOSHTTPException in all cases, and leave it to the caller if they want to print some customized/augmented error messages.

mnaboka:
run all integration tests

mnaboka:
run all integration tests

bamarni:
@mnaboka : windows and mac failures seem related to your changes : 
```
assert 'foo bar error1' in 'Error while fetching [<MagicMock name=\'get().request.url\' id=\'4542321104\'>]: HTTP 500: "<MagicMock name=\'get().reason\' id=\'4542420920\'>". Error message: foo bar error'
```

mnaboka:
run all integration tests

mnaboka:
@bamarni right, that was a typo and it's fixed now, re-running the tests

mnaboka:
@bamarni some tests are failing, do you know what could be the issue?
```
[mac-tests] >       assert stderr == b"This cluster link already exists.\n"
[mac-tests] E       AssertionError: assert b'URL [https:...nreachable.\n' == b'This cluster...ady exists.\n'
[mac-tests] E         At index 0 diff: 85 != 84
[mac-tests] E         Left contains more items, first extra item: 105
[mac-tests] E         Full diff:
[mac-tests] E         - b'URL [https://34.215.174.248/acs/api/v1/auth/providers] is unreachable.\n'
[mac-tests] E         + b'This cluster link already exists.\n'
```

bamarni:
@mnaboka : mac integration tests have been flaky lately as they fail in many places due to connectivity issue. I don't exactly know the root cause, either since we upgraded the mac mini or because of network problems in the sf office. This is tracked in jira.

PR #1140, opened 2018-01-29
Add missing `__init__.py` to dcoscli/test folder

bamarni:
This makes it a proper package which can then be exposed an used by other projects when importing the `dcoscli` dependency.

The package contains helper functions useful in tests (to run commands, interact with Marathon, etc.).


bamarni:
run linux integration tests

adamtheturtle:
If the goals stated in the description are appropriate, this is a suitable PR.

However, I thought @klueska was hoping that we decrease the dependency on importing from `dcoscli` - Kevin, can you comment to that?

Perhaps such a discussion would be better placed in JIRA.

bamarni:
Our main problem is that we have 2 Pypi packages instead of one, and we want to be able to break / refactor the package without making a guarantee about backward compatibility. Here if I'm relying on dcoscli by importing it, I'm doing this knowing that I can expect breakages that I'd need to fix.

And even though we'd remove the packages from Pypi, one would still be able to import them using git, so I think this is a different discussion. Nonetheless I think this fix is valid.

PR #1144, opened 2018-02-06
Properly query for DC/OS version on Windows 

bamarni:
Currently it is using `os.path`, which would use a backslash on Windows, this makes the HTTP request to fail.

Discovered while looking at https://jira.mesosphere.com/browse/COPS-2400


bamarni:
run windows integration tests

bamarni:
run windows integration tests

bamarni:
run linux integration tests

PR #1143, opened 2018-02-02
Improvements

bamarni:
- Use a variable in a test
- Use a try / except / else clause 
- Revert "Moved cluster attachment to not lose track of attached cluster."

PR #1147, opened 2018-02-08
Give priority to Cluster over LinkedCluster when listing

bamarni:
Otherwise the name returned from the cluster linker service will take precedence over a potential custom name in the config file.

https://jira.mesosphere.com/browse/DCOS-21082

### DIFF
--- a/dcos/cluster.py
+++ b/dcos/cluster.py
@@ -251,13 +251,6 @@ def get_clusters(include_linked=False):
     util.ensure_dir_exists(clusters_path)
     clusters = set()
 
-    if include_linked:
-        try:
-            linked_clusters = get_linked_clusters()
-            clusters.update(linked_clusters)
-        except DCOSException:
-            pass
-
     uuid_regex = re.compile((r'^[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-'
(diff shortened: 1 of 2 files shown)


bamarni:
run linux integration tests

bamarni:
run mac integration tests

bamarni:
Failing build mac is a known issue, seems like the mac mini currently has some connectivity issues.

dcos/cluster.py
```diff
@@ -267,6 +260,15 @@ def get_clusters(include_linked=False):
             c = Cluster(entry)
             clusters.add(c)
 
+    # Search for linked clusters after configured clusters, this makes sure
+    # the set gives priority to Cluster over LinkedCluster.
+    if include_linked:
+        try:
+            linked_clusters = get_linked_clusters()
+            clusters.update(linked_clusters)
```
adamtheturtle: ❓ This should go in an `else` clause, only what can raise the exception should be in `try`. Optional because this is not new.

adamtheturtle:
@bamarni Please can you confirm:

* You manually reproduced the issue
* The same steps could not be manually reproduced with this change

Then ping me for approval, thanks!

dcos/cluster.py
bamarni: Makes sense :+1: I'll push it tomorrow, here or otherwise into #1143

PR #1149, opened 2018-02-09
Disable self-linking

bamarni:
User test reports have shown this to be confusing, as once they configure a second cluster, they forget to attach back to the original one.

https://jira.mesosphere.com/browse/DCOS-21047


bamarni:
run linux integration tests

PR #1150, opened 2018-02-12
Moved cluster attachment to not lose track of attached cluster.

ArmandGrillet:


ArmandGrillet:
run linux integration tests

cli/dcoscli/cluster/main.py
```diff
@@ -368,10 +369,10 @@ def setup(dcos_url,
     :rtype: int
     """
 
-    with cluster.setup_directory() as temp_path:
+    with util.tempdir() as tempdir:
```
bamarni: Is the `cluster.setup_directory()` function still in use after this change? It can probably be removed.

ArmandGrillet: Still used in https://github.com/dcos/dcos-cli/blob/a705fa3587cb5ba873aede859a9aa42c37f5f40a/tests/test_cluster.py#L117 but it's the test failing in CI thus I'm gonna update it now!

ArmandGrillet:
run linux integration tests

ArmandGrillet:
run linux integration tests

ArmandGrillet:
run linux integration tests

cli/dcoscli/cluster/main.py
bamarni: If it's now only used in a test we should remove that function and either : 
- update the test accordingly
- remove the test (if you think it's not needed anymore)

ArmandGrillet:
run linux integration tests

ArmandGrillet:
run linux integration tests

ArmandGrillet:
run linux integration tests

bamarni:
run windows integration tests

PR #1152, opened 2018-02-13
Fix nightly slack notification

bamarni:
I've added a `BRANCH_NAME=master` environment variable in the nightly build config, so we should now receive slack notifications on failures.

This patch is just about using `task_wrapper()` on the main node instead of the inner one. This would send notifications for failures during `dcos_launch` too.


bamarni:
run all integration tests

PR #1153, opened 2018-02-13
Deprecate installation of global subcommands

bamarni:
Passing the `--global` flag to the `dcos package install` command
is now deprecated and support for this will be removed in 1.12.

Subcommands should always get installed for a given cluster instead.

https://jira.mesosphere.com/browse/DCOS_OSS-2157


bamarni:
run linux integration tests

PR #1155, opened 2018-02-16
Ignore DCOS_CLUSTER_NAME env var

bamarni:
This env variable causes inconsistent behaviours, eg. `dcos cluster list` would display all clusters with the same name when DCOS_CLUSTER_NAME is set.

One should instead run `dcos cluster rename <cluster> <new_name>` in order to change a cluster name.

https://jira.mesosphere.com/browse/DCOS-21098

### DIFF
--- a/dcos/config.py
+++ b/dcos/config.py
@@ -170,12 +170,16 @@ def get_config_val_envvar(name, config=None):
 
     section, subkey = split_key(name.upper())
 
-    env_var = None
+    env_var = ''
     if section == "CORE":
         if subkey.startswith("DCOS") and os.environ.get(subkey):
             env_var = subkey
         else:
             env_var = "DCOS_{}".format(subkey)
+    elif section == "CLUSTER" and subkey == "NAME":
+        # Ignore DCOS_CLUSTER_NAME, this environment variable is confusing.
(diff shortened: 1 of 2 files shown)


bamarni:
run linux integration tests

dcos/config.py
```diff
@@ -170,12 +170,16 @@ def get_config_val_envvar(name, config=None):
...
     section, subkey = split_key(name.upper())
 
-    env_var = None
+    env_var = ''
     if section == "CORE":
         if subkey.startswith("DCOS") and os.environ.get(subkey):
             env_var = subkey
         else:
             env_var = "DCOS_{}".format(subkey)
+    elif section == "CLUSTER" and subkey == "NAME":
```
klueska: I'm not a huge fan of special casing just this one ENV variable and ignoring it. Is there some more general solution to this problem?

bamarni: Thinking about all the env variables we want to support and explicitly whitelisting them would sound cleaner to me. I went for the quickfix here as this will get replaced soon during the 1.12 refactoring.

bamarni:
@klueska @ArmandGrillet : any news about this one?

PR #1159, opened 2018-02-22
Added config management in Go CLI.

bamarni:
# Bootstrap the config for the new CLI

## Technical design doc

https://docs.google.com/document/d/1SumMm9eqXPTjVek4L__n7VXBbTi0N-YVPUyxViApsvA/edit?usp=sharing

## Summary

It adds a `Config` struct which can be generated from a TOML file, it also accepts `DCOS_*` env vars.

Usage example : 
``` go
package main

import (
	"fmt"
	"github.com/dcos/dcos-cli/pkg/config"
)

var cfgStr = `
[core]
dcos_acs_token = "zj8Tb0vhQw"
dcos_url = "https://54.245.134.104"
ssl_verify = "/path/to/dcos_ca.crt"

[cluster]
name = "jkshsfhjsg"
`

func main() {
	conf, err := config.FromString(cfgStr)
	// OR... conf, err := config.FromPath("/path/to/config")

	// handle error
	if err != nil {
	}

	fmt.Println(conf.URL)

	// proceed with the config
	// conf.URL == "https://54.245.134.104"
	// conf.TLS.Insecure == false
	// conf.TLS.RootCAsPath == "/path/to/dcos_ca.crt"
	// etc.
}
```

It also adds the following commands : 
- `dcos config show`
- `dcos config set`
- `dcos config unset`

## Next steps

Validation is still missing but I think this should be done later in a separate PR, maybe not even right after this one as it needs some abstraction we don't yet have (installing subcommands and reading their associated JSON schema).


pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+// Additional sections such as "marathon" are stored in a Tree.
+// This difference is due to the fact that we know the content of "core" and "cluster"
+// but not the content of additional sections.
+type Config struct {
+	Core    Core    `toml:"core"`
+	Cluster Cluster `toml:"cluster"`
+	tree    *toml.Tree
+}
+
+// Core is the main section of the DC/OS CLI configuration. It contains the DC/OS URL, an authentication
```
ArmandGrillet: When cutting a comment, let's make the length similar for each line or cut using periods. E.g. here:
```
// Core is the main section of the DC/OS CLI configuration. 
// It contains  the DC/OS URL, an authentication token and various other config options.
```

pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
+package config
+
+import (
+	"strconv"
+
+	"github.com/pelletier/go-toml"
+)
+
+// DefaultConfig for the DC/OS CLI.
```
ArmandGrillet: What is the purpose of this `var`? Why a Timeout of `180`? The code already describes what is currently in the comment.

pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+	// The token generated by authenticating to DC/OS using the Admin Router Access Control Service.
+	ACSToken string `toml:"dcos_acs_token" env:"DCOS_ACS_TOKEN"`
+
+	// Whether to verify server certs for HTTPS or path to a root CA cert. Valid values are:
+	// - A path to a root CA certifacte.
+	// - 1, t, T, TRUE, true, True - will use CA Certificates from the system.
+	// - 0, f, F, FALSE, false, False - will send insecure requests.
+	SSL SSL `toml:"ssl_verify" env:"DCOS_SSL_VERIFY"`
+
+	// Request timeout in seconds once the connection is established.
```
ArmandGrillet: Let's add a `The` at the beginning of the comment.

pkg/config/decode.go
```diff
@@ -0,0 +1,114 @@
...
+	case reflect.Struct:
+		// The current type is a struct, iterate through its fields.
+		for i := 0; i < ctx.typ.NumField(); i++ {
+			field := ctx.typ.Field(i)
+			// Lookup for the "toml" tag and skip this field if not present.
+			tomlKey, ok := field.Tag.Lookup("toml")
+			if !ok {
+				continue
+			}
+			// Generate a new decodeContext from the struct field and continue unmarshaling.
```
ArmandGrillet: Similarly to what you did in `decode`, you can add a blank line before this comment.

pkg/config/decode.go
```diff
@@ -0,0 +1,114 @@
...
+		ctx.typ = ctx.typ.Elem()
+		ctx.val = ctx.val.Elem()
+	}
+
+	switch ctx.typ.Kind() {
+	case reflect.Struct:
+		// The current type is a struct, iterate through its fields.
+		for i := 0; i < ctx.typ.NumField(); i++ {
+			field := ctx.typ.Field(i)
+			// Lookup for the "toml" tag and skip this field if not present.
```
ArmandGrillet: Similarly to what you did in `decode`, you can add a blank line before this comment.

pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Cluster contains cluster-related configurations.
+type Cluster struct {
+	// Custom name for the cluster. If not set it defaults to what is returned by the /mesos/state-summary
+	// endpoint.
+	Name string `toml:"name"`
+}
+
+// Trust policy for server certificates during HTTPS.
```
ArmandGrillet: Can we explain why we create a new type here?

pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+	MesosMasterURL string `toml:"mesos_master_url"`
+
+	// Whether to prompt the user to log in when token expired.
+	// If set to `false`, otherwise automatically initiate login
+	PrompLogin bool `toml:"prompt_login"`
+}
+
+// Cluster contains cluster-related configurations.
+type Cluster struct {
+	// Custom name for the cluster. If not set it defaults to what is returned by the /mesos/state-summary
```
ArmandGrillet: Let's cut the comment another way:
```
// Custom name for the cluster. 
// If not set, defaults to what is returned by the /mesos/state-summary endpoint.

File: pkg/config/decode.go [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/pelletier/go-toml"
+)
+
+// decode parses a TOML tree (and environment variables) according to field tags set in the Config struct.
+//
+// Essentially, it expects `toml` and `env` field tags to be set. Cf. Config for reference.
+func decode(tree *toml.Tree) (conf Config, err error) {
+	// The use of reflection and user-provided data is a very sensitive combination,
+	// searching for "panic" in https://golang.org/pkg/reflect/ shows 82 results.
```
ArmandGrillet: This line is maybe not necessary.

pkg/config/decode.go
```diff
@@ -0,0 +1,114 @@
...
+	//
+	// Let's be on the safe side and recover from any panic during unmarshalling in order to
```
ArmandGrillet: s/`Let's be on the safe side and `/`We`

pkg/config/decode.go
```diff
@@ -0,0 +1,114 @@
...
+		// Check if the current type supports env var overriding and look-it up if needed.
+		if ctx.envVar != "" {
+			if envValStr, ok := os.LookupEnv(ctx.envVar); ok {
+				if envVal, err := convertStringTo(envValStr, ctx.typ); err == nil {
+					ctx.val.Set(reflect.ValueOf(envVal))
+					return
+				}
+			}
+		}
+		// Fallback to reading from the TOML tree.
```
ArmandGrillet: Similarly to what you did in `decode`, you can add a blank line before this comment.

pkg/config/decode_test.go
```diff
@@ -0,0 +1,57 @@
...
+	cleanup, err := clitesting.WithEnv(map[string]string{
+		"DCOS_URL":          "https://dcos-env.example.com",
+		"DCOS_CLUSTER_NAME": "nope",
+	})
+	require.NoError(t, err)
+	defer cleanup()
+
+	treeMap := map[string]interface{}{
+		"core": map[string]interface{}{
+			"dcos_url": "https://dcos.example.com", // this should get overriden by DCOS_URL
```
ArmandGrillet: Let's stick to a comment before `treeMap := map[string]interface{}{` instead of inline comments.

pkg/config/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package config manages DC/OS CLI configurations for your clusters.
```
ArmandGrillet: s/`manages DC/OS CLI configurations for your clusters.`/`manages CLI configurations for your DC/OS clusters.`

pkg/config/doc.go
```diff
@@ -0,0 +1,5 @@
...
+//
+// Each configured cluster has an associated TOML configuration file which can be parsed,
+// updated or displayed through the Config struct.
```
ArmandGrillet: s/`updated or displayed`/`updated,  or displayed` https://en.wikipedia.org/wiki/Serial_comma

pkg/testing/env.go
```diff
@@ -0,0 +1,33 @@
+package testing
+
+import (
+	"os"
+)
+
+// WithEnv updates environment variables according to the vars map argument.
+// A cleanup function is returned in order to revert the environment back to its original state.
```
ArmandGrillet: This is cool ✨ !

pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+	// endpoint.
+	Name string `toml:"name"`
+}
+
+// Trust policy for server certificates during HTTPS.
+type SSL string
```
mhrabovcin: Can we call it `TLS`? The `SSL` is outdated name that shouldn't be used.

pkg/config/config_test.go
```diff
@@ -0,0 +1,27 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSSL(t *testing.T) {
```
mhrabovcin: Also here, lets use `TLS`.

bamarni:
@ArmandGrillet @mhrabovcin : I've updated the PR according to your comments.

pkg/config/decode.go
```diff
@@ -0,0 +1,113 @@
...
+	"reflect"
+	"strconv"
+
+	"github.com/pelletier/go-toml"
+)
+
+// decode parses a TOML tree (and environment variables) according to field tags set in the Config struct.
+//
+// Essentially, it expects `toml` and `env` field tags to be set. Cf. Config for reference.
+func decode(tree *toml.Tree) (conf Config, err error) {
```
jdef: it would be nice if this was a pure function instead of embedding a dependency on calls to the `os` package to resolve environment variables. Some suggestions for alternative approaches:

1. Consider adding an `env []string` parameter to this func (and to `decodeContext`) which a caller MAY populate with the results of `https://golang.org/pkg/os/#Environ`, or else;
2. Consider adding a `lookupEnv func(string)(string,bool)` parameter to this func (and to `decodeContext`) which a caller may populate with a reference to `os.LookupEnv`.

I tend to prefer (2). This would also simplify unit testing: you wouldn't need to muck around with setting/restoring the process environment just to unit test the decoding logic here.

.gitignore
```diff
@@ -18,3 +18,5 @@ bundle.zip
 Vagrantfile
 .vagrant
 .idea
+
+vendor
```
jdef: if this project is going to produce binaries, then why is `vendor` excluded?

pkg/config/config.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"strconv"
+
+	"github.com/pelletier/go-toml"
+)
+
+// DefaultConfig for the DC/OS CLI.
+//
+// All config defaults are their zero value except for the timeout which is 180 seconds.
+var DefaultConfig = Config{
```
jdef: is the intent that some caller might want to change the global default here? if that's not critical, then please don't use a `var` and consider instead defining this as a func that returns a `Config`

pkg/config/decode.go
bamarni: Indeed that makes unit tests easier, 1. might have been good as it makes only one syscall, but I went for the second which is also nice.

pkg/config/config.go
bamarni: I've added a comment in the first commit, this also doesn't seem like a hard suggestion from dep (https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory).

.gitignore
bamarni: I've added a comment in the first commit, this also doesn't seem like a hard suggestion from dep (https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory).

pkg/config/config.go
bamarni: It is not, I've moved it to a function.

.gitignore
jdef: some consequences of NOT committing a vendor directory are::
(1) reproducible builds are difficult to do offline if you don't already have vendored sources sitting somewhere local
(2) CI builds will have external dependencies because they need to download external sources for every build. this can also make CI builds slower
(3) if the project isn't pinning **exact** dependency versions via `dep` config, then dependencies can drift from build to build across different systems leading to subtle bugs (by default, dependencies via `dep` are not pinned to exact versions, there's an implicit `^` prefix injected when referencing project dependencies by semantic version)

bamarni: About dependencies drifting, we could specify tags, have a CI step which fails if the lock file gets altered after dep ensure, there should be solutions to prevent this.

This is just my personal preference, I'll make sure to raise your point in our team sync and we'll see what the consensus is.

benclarkwood: I second that committing the vendor directory solves and guards against a lot of problems.

bamarni: If by problems you mean for eg. a dependency remote repo getting deleted, committing dependencies would hide the problems instead of solving them imo. These kind of problems would only get stacked until someone eventually needs to bump dependencies and actually solve them (by using another dep, finding a fork, moving it to the project codebase, or other mean).

I'd rather see our nightly integration tests failing in such situations, so we can be notified early and take appropriate action.

We also have pip dependencies in the Python CLI, eg. docopt, requests, etc. The setup is not ideal right now as we don't require tagged versions, but would you also argue that we should commit them as part of our codebase?

@klueska @ArmandGrillet  wdyt?

jdef:
Reproducible builds are very important. The fact that dep, by default,
doesn't pin exact versions works against this goal - which is compounded
when code in vendor/ is not committed.

The argument that (paraphrasing) "we don't have this w/ python, so we don't
need it for golang" seems a bit shaky. Let's call the python-approach
"deprecated" and do better with the golang-based project.

.gitignore
mhrabovcin: From security perspective its a problem if we cannot build new version of program if one of the dependencies isn't available because of repo deletion. There is a possibility of new `golang` version coming out with a security fix that would require us to rebuild binary in order to get security fix in place.

Committing `/vendor` directory is a standard practice in `golang` projects and most of ours projects (if not all) are following same pattern. We should be consistent in order to not confuse developers.

https://github.com/mesosphere/dcos-secrets
https://github.com/mesosphere/dcos-go-enterprise
https://github.com/mesosphere/dcos-cluster-linker
https://github.com/mesosphere/dcos-ca

bamarni: ok, I've added dependencies to this PR.

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"strconv"
+
+	"github.com/pelletier/go-toml"
+)
+
+// Config for the DC/OS CLI. The "core" and "cluster" sections are mapped to structs.
+// Additional sections such as "marathon" are stored in a Tree.
+// This difference is due to the fact that we know the content of "core" and "cluster"
```
ArmandGrillet: You can add an empty line for comments before this line.

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+	// If not set it defaults to what is returned by the /mesos/state-summary endpoint.
+	Name string `toml:"name"`
+}
+
+// TLS defines the trust policy for server certificates during HTTPS.
+// It has its dedicated type as it can be configured as a string or bool-like string.
+type TLS string
+
+// Insecure specifies if server certificates should be accepted without verification for HTTPS requests.
+// Skipping verification against the system's CA bundle or a cluster-specific CA is highly discouraged
```
ArmandGrillet: You can add an empty line for comments before this line.

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+// When empty, the CLI will default to the system's CA bundle.
+func (s TLS) CAPath() string {
+	if _, err := strconv.ParseBool(string(s)); err == nil {
+		return ""
+	}
+	return string(s)
+}
+
+// DefaultConfig returns the default config for the DC/OS CLI.
+//
```
ArmandGrillet: Not needed, the second line being strongly linked to the first one.

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+// CAPath returns the path to a custom root CA used to verify server certificates.
+// When empty, the CLI will default to the system's CA bundle.
+func (s TLS) CAPath() string {
+	if _, err := strconv.ParseBool(string(s)); err == nil {
+		return ""
+	}
+	return string(s)
+}
+
+// DefaultConfig returns the default config for the DC/OS CLI.
```
ArmandGrillet: s/`config`/`configuration`

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+// All config defaults are their zero value except for the timeout which is 180 seconds.
+func DefaultConfig() Config {
+	return Config{
+		Core: Core{
+			Timeout: 180,
+		},
+	}
+}
+
+// FromPath creates a Config based on a path to a .toml file.
```
ArmandGrillet: s/`based on a`/`using a path`

pkg/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+func FromPath(path string) (Config, error) {
+	tree, err := toml.LoadFile(path)
+	if err != nil {
+		return Config{}, err
+	}
+	return decode(tree, nil)
+}
+
+// FromString creates a Config using a string as input.
```
ArmandGrillet: s/` as input`/` representing the configuration formatted as a TOML document`

pkg/config/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package config manages CLI configurations for your DC/OS clusters.
+//
+// Each configured cluster has an associated TOML configuration file which can be parsed,
```
ArmandGrillet: Let's cut the comment just after `file` and bring the rest in the next line to have something of the same length.

pkg/config/decode.go
```diff
@@ -0,0 +1,120 @@
...
+	"os"
+	"reflect"
+	"strconv"
+
+	"github.com/pelletier/go-toml"
+)
+
+// decode parses a TOML tree (and environment variables) according to field tags set in the Config struct.
+// lookupEnv is used to look-up for env variables, if not provided it defaults to os.LookupEnv.
+// Essentially, decode expects `toml` and `env` field tags to be set. Cf. Config for reference.
```
ArmandGrillet: What does `field tags` means?

pkg/config/decode.go
```diff
@@ -0,0 +1,120 @@
...
+func convertStringTo(str string, targetType reflect.Type) (interface{}, error) {
+	switch targetType.Kind() {
+	case reflect.String:
+		return str, nil
+	case reflect.Bool:
+		return strconv.ParseBool(str)
+	case reflect.Int:
+		return strconv.Atoi(str)
+	default:
+		return nil, fmt.Errorf("couldn't convert string \"%s\" to \"%s\"", str, targetType)
```
ArmandGrillet: According to the Go codebase, it's OK to have contractions in error messages (e.g. https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/misc/cgo/testsigfwd/main.go#L107) :+1:.

pkg/config/decode_test.go
```diff
@@ -0,0 +1,60 @@
...
+	require.Equal(t, "https://dcos.example.com", conf.Core.URL)
+	require.Equal(t, true, conf.Core.TLS.Insecure())
+	require.Equal(t, 15, conf.Core.Timeout)
+	require.Equal(t, true, conf.Core.Reporting)
+	require.Equal(t, "mr-cluster", conf.Cluster.Name)
+}
+
+func TestDecodeWithEnvVars(t *testing.T) {
+	// core.dcos_url should get overriden by DCOS_URL.
+	// cluster.name shouldn't get overriden by DCOS_CLUSTER_NAME.
```
ArmandGrillet: According to the Go codebase, it's OK to have contractions in comments :+1: https://github.com/golang/go/blob/29461ccc7f7b6a73bc0104731cfbe0c2cbe86733/src/go/parser/interface.go#L131

pkg/config/decode.go
bamarni: These are metadata one can add to struct fields, they can then be read through the reflect package. This is how we define the `toml` and `env` expected values for each config property.

pkg/config/config.go
```diff
@@ -95,3 +95,21 @@ func DefaultConfig() Config {
...
+func FromPath(path string) (Config, error) {
+	tree, err := toml.LoadFile(path)
+	if err != nil {
+		return Config{}, err
+	}
+	return decode(tree, nil)
+}
+
+// FromString creates a Config using a string representing the configuration formatted as a TOML document.
+func FromString(tomlData string) (Config, error) {
```
collinvandyck: Maybe have a `FromReader(io.Reader) (Config, error)` method that actually calls the toml lib, and have the other methods call that...

pkg/config/decode.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+type decodeContext struct {
+	val      reflect.Value
+	typ      reflect.Type
+	tomlPath []string
+	envVar   string
+}
+
+func (d *decoder) unmarshal(ctx decodeContext) {
```
collinvandyck: It seems weird that we have to do all of this work to unmarshal a toml document.

bamarni: At first I was considering to use third-party libs to unmarshal the config into a struct.
As we have 2 sources for the config (toml file and env variables to override),
it'd have required to unmarshal the struct twice, and the libraries I've found
had some quircks. For eg. https://github.com/caarlos0/env says :

> if the environment variable is not set, the zero-value of the type will be used

So it would overwrite the values set during the previous toml unmarshaling, unless we use
2 different structs and something like https://github.com/imdario/mergo. Eventually I felt
like it would be a cleaner solution to just have our own logic. The hard work is still being
done by the toml lib (parsing the configuration and generating a tree out of it), we just have
our own way of mapping that tree into a struct.

bamarni:
One limitation worth mentioning right now is that, once we'll have the opposite of decode (encoding a config into toml), it won't be possible to all in a raw decode a config, change values in the struct and encode it afterwards (as some env variables might then get injected in the toml config).

I don't foresee this as something we would suffer from, the only command I can think of which would do something close to this is `dcos config set <toml_key> <value>`, and in that case we would set the value in the toml tree directly and not the struct.

pkg/config/config.go
bamarni: I've introduced a Store and added a `fromStore` function which are called by the other functions. As I think the data source here can't be limited to the config file (we also have env vars).

bamarni:
On @collinvandyck 's previous point (`It seems weird that we have to do all of this work to unmarshal a toml document.`), I tend to agree that this could be simplified.

Reflection is useful when traversing unknown interfaces, in our case we only support a well-known Config struct. Thus setting tags to its fields and using reflection for unmarshaling the data source might be overkill.

We could have something like this : 
``` go
func Unmarshal(conf *Config, store *Store) error {
	for _, key := range store.Keys() {
		val, _ := store.Get(key)
		switch key {
		case "core.dcos_url":
			conf.Core.URL = val
		case "core.dcos_acs_token":
			conf.Core.ACSToken = val
		// etc.
		}
	}
}
```

(Though the switch statement would probably be in a ` func (c *Config) Set(key string, val interface{})`.)

What do you guys think?

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+		opts.EnvLookup = os.LookupEnv
+	}
+	return &Store{
+		tree:         tree,
+		envWhitelist: opts.EnvWhitelist,
+		envLookup:    opts.EnvLookup,
+	}
+}
+
+// Get a config value from the Store.
```
ArmandGrillet: I would make this more explicit: `Get returns a value from the Store using a key`.

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+func (s *Store) Get(key string) (interface{}, bool) {
```
ArmandGrillet: The second value returned seems unnecessary.

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+	// Check if the given key is whitelisted for env var override.
```
ArmandGrillet: `for env var override.` 🤔

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+	if envVar, ok := s.envWhitelist[key]; ok {
+		// If so, look-it up.
+		if envVal, ok := s.envLookup(envVar); ok {
+			return envVal, true
+		}
+	}
+
+	// Fallback to reading from the TOML tree if present.
```
ArmandGrillet: `reading from` seems unnecessary.

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+	return nil, false
+}
+
+// Set a key in the store.
+func (s *Store) Set(key string, val interface{}) {
+	switch key {
+	case keyTimeout:
+		// go-toml requires int64
+		val = cast.ToInt64(val)
+	case keyPagination, keyReporting, keyPrompLogin:
```
ArmandGrillet: Having this list here seems not that readable.

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+					s.tree, _ = toml.TreeFromMap(treeMap)
+				} else {
+					return
+				}
+			}
+		}
+	}
+}
+
+// Keys returns all keys present in the Store.
```
ArmandGrillet: s/`all keys present`/`all the keys`

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+		if keys[i] == keys[i-1] {
+			keys = append(keys[:i], keys[i+1:]...)
+			keysLen--
+			i--
+		}
+	}
+	return keys
+}
+
+func searchKeys(tree *toml.Tree, keys *[]string, keyPath []string) {
```
ArmandGrillet: I would still add some comments regarding the parameters we take and the basic functionality of this function, even if this is private.

pkg/cmd/dcos_config_show.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// configShowCmd represents the config command
+var configShowCmd = &cobra.Command{
+	Use:  "show",
+	Args: cobra.MaximumNArgs(1),
+	RunE: runConfigShowCmd,
+}
+
+var maskedKeys = []string{
```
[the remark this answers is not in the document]
bamarni: I've added it here as the show command shouldn't be part of the config package imo, it is just one opinionated way to print the config to users, and not generic enough.

pkg/cmd/dcos_config_show.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func runConfigShowCmd(cmd *cobra.Command, args []string) error {
+	store := attachedCluster().Config.Store
+
+	if len(args) == 1 {
+		return configShow(store, args[0])
+	}
+
+	configShowAll(store)
```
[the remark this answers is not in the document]
bamarni: This is also the case for `configShow()`, the intent for having functions there is readability. When looking at this function it is clear what it does. If I were to remove these private functions, the code would then be : 
``` go
func runConfigShowCmd(cmd *cobra.Command, args []string) error {
	store := attachedCluster().Config.Store()

	if len(args) == 1 {
		if val, ok := store.Get(key); ok {
			fmt.Printf("%v\n", val)
			return nil
		}
		return fmt.Errorf("unknown key \"%s\"", key)
	}

	for _, key := range store.Keys() {
		if val, ok := store.Get(key); ok {
			for _, maskedKey := range maskedKeys {
				if maskedKey == key {
					val = "********"
				}
			}
			fmt.Printf("%s %v\n", key, val)
		}
	}
	return nil
}
```
Which is much harder to read through.

pkg/config/store.go
```diff
@@ -0,0 +1,150 @@
...
+	"sort"
+	"strings"
+
+	toml "github.com/pelletier/go-toml"
+	"github.com/spf13/cast"
+)
+
+// StoreOpts are functional options for a Store.
+type StoreOpts struct {
+	EnvWhitelist map[string]string
```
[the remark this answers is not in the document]
bamarni: I've added a comment, I think EnvWhitelist makes more sense according to what it is. WhitelistEnv implies more of an action (to whitelist the environment), while it is actually just a list of env variables.

pkg/cmd/dcos_config_show.go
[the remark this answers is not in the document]
ArmandGrillet: You're right :+1:

pkg/cmd/dcos.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+type Cluster struct {
+	Config config.Config
+}
+
+// This is a temporary function until we have the proper abstraction for this.
+func attachedCluster() *Cluster {
+	path := "/home/bamarni/.dcos/dcos.toml"
```
ArmandGrillet: Let's replace this line by:
```
usr, _ := user.Current()
path := filepath.Join(usr.HomeDir, ".dcos", "dcos.toml")
```
This requires `"os/user"` and `"path/filepath"`.

pkg/cmd/dcos_config_show.go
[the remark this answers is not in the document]
ArmandGrillet: True :+1:, let's just inverse the order of the functions `configShowAll` and `configShow` then.

pkg/config/config.go
```diff
@@ -0,0 +1,133 @@
...
+	store   *Store
+	Path    string
+}
+
+// Store returns the config Store.
+func (conf *Config) Store() *Store {
+	return conf.store
+}
+
+// Save the config to the path specified in the Config.Path field.
```
ArmandGrillet: s/`Save the`/`Save writes the`.

pkg/config/doc_test.go
```diff
@@ -0,0 +1,39 @@
+package config_test
+
+import (
+	"fmt"
+
+	"github.com/dcos/dcos-cli/pkg/config"
+)
+
+// Create a config from a TOML file path.
```
ArmandGrillet: s/`Create a config`/`Creates a Config`

pkg/config/doc_test.go
```diff
@@ -0,0 +1,39 @@
...
+	conf, err := config.FromPath("/path/to/config.toml")
+	if err != nil {
+		// Handle error.
+	}
+
+	// Displays the DC/OS URL if it exists in the file.
+	fmt.Println(conf.Core.URL)
+}
+
+// Create a config from a TOML string.
```
ArmandGrillet: s/`Create a config`/`Creates a Config`

pkg/config/doc_test.go
```diff
@@ -0,0 +1,39 @@
...
+	conf, err := config.FromPath("/path/to/config.toml")
+	if err != nil {
+		// Handle error.
+	}
+
+	// Displays the DC/OS URL if it exists in the file.
+	fmt.Println(conf.URL)
+}
+
+// Create a config from a TOML string.
```
ArmandGrillet: s/`Create a config`/`Creates a Config`

pkg/config/marshal.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Try to cast the value to a bool, true means we verify
+	// server certificates, false means we skip verification.
+	if verify, err := strconv.ParseBool(strVal); err == nil {
+		return TLS{
+			Insecure: !verify,
+		}
+	}
+
+	// If the value is not a bool-like string, it means it's a path to a root CA bundle.
```
ArmandGrillet: s/`is not a bool-like string`/`is not a string representing a bool`

pkg/config/marshal.go
```diff
@@ -0,0 +1,124 @@
...
+
+// UnmarshalTLS creates a TLS struct from a string or bool-like string.
+//
+// Valid values are:
+// - A path to a root CA bundle.
+// - "1", "t", "T", "TRUE", "true", "True" - will use the system's CA bundle.
+// - "0", "f", "F", "FALSE", "false", "False" - will send insecure requests.
+func UnmarshalTLS(val interface{}) TLS {
+	strVal := cast.ToString(val)
+	if strVal == "" {
```
ArmandGrillet: A blank line before and a comment explaining that condition could be useful.

pkg/config/marshal.go
```diff
@@ -0,0 +1,124 @@
...
+	// If the value is not a bool-like string, it means it's a path to a root CA bundle.
+	rootCAsPEM, err := ioutil.ReadFile(strVal)
+	if err != nil {
+		return TLS{
+			Insecure:    true,
+			RootCAsPath: strVal,
+		}
+	}
+
+	certPool := x509.NewCertPool()
```
ArmandGrillet: What are we doing here? It is explained in `config.go` that it is a `pool of root CAs to verify server certificates against.` but writing for which reasons would we arrive in that part of the function and what does it imply could be useful.

pkg/config/marshal.go
```diff
@@ -0,0 +1,124 @@
...
+	if !certPool.AppendCertsFromPEM(rootCAsPEM) {
+		return TLS{
+			Insecure:    true,
+			RootCAsPath: strVal,
+		}
+	}
+
+	return TLS{
```
ArmandGrillet: Could you add a comment explaining in which cases do we go until this `return` and why are we not setting `Insecure`?

pkg/cmd/dcos.go
bamarni: makes sense :+1:

pkg/cmd/dcos_config_show.go
```diff
@@ -0,0 +1,55 @@
...
+	Args: cobra.MaximumNArgs(1),
+	RunE: runConfigShowCmd,
+}
+
+var maskedKeys = []string{
+	"core.dcos_acs_token",
+}
+
+func runConfigShowCmd(cmd *cobra.Command, args []string) error {
+	store := attachedCluster().Config.Store
```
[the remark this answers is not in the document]
bamarni: In the PR first message I've added a link to a technical design doc explaining the Config / Store difference.

pkg/cmd/dcos_config_show.go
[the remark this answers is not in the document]
bamarni: for what reason?

pkg/config/doc_test.go
bamarni: I've changed the name of the example functions, now they appear right after the doc for their associated functions, I think it's more readable like this and doesn't need a comment, so I've removed it and the one for FromString too.

ArmandGrillet:
run linux integration tests

PR #1160, opened 2018-02-27
Simplify setup_cluster_config()

bamarni:
This simplifies the setup_cluster_config() function and updates tests to make sure that the core.ssl_verify config with a CA downloaded from the cluster points to the correct cluster path.

https://jira.mesosphere.com/browse/DCOS_OSS-2201

### DIFF
--- a/cli/dcoscli/cluster/main.py
+++ b/cli/dcoscli/cluster/main.py
@@ -420,8 +420,8 @@ def setup(dcos_url,
             raise DCOSException(msg)
 
         # configure cluster directory
-        cluster_path = cluster.setup_cluster_config(dcos_url, real_config_dir,
-                                                    stored_cert)
+        cluster_path = cluster.setup_cluster_config(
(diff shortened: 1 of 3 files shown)


bamarni:
run linux integration tests

bamarni:
@mnaboka : I couldn't reproduce that problem, however I've updated the related code and added a test against this. Are you able to provide the exact steps to reproduce it? And does this PR fixes it?

klueska:
@bamarni To reproduce, just manually set the `ssl_verify` value to a non-existent path in your `dcos.toml`

bamarni:
Indeed there is no validation when setting ssl_verify to a path (https://jira.mesosphere.com/browse/DCOS-11460). Here I'm more concerned about the cause, I don't think @mnaboka explicitely did explicitly set it to this tmp path, it seems instead related to a recent change (#1150).

See also https://mesosphere.slack.com/archives/C047TFUJA/p1519333331000290

klueska:
He was using the CLI from source and did the setup before updating to master.

PR #1161, opened 2018-02-28
Prepare 0.6.x

bamarni:
- Revert "Moved cluster attachment to not lose track of attached cluster."
- Ignore DCOS_CLUSTER_NAME env var
- Update DC/OS installer credentials to use 1.11


bamarni:
run all integration tests

Issue #1167, opened 2018-03-06
--password-file options does not strip newlines from file


dirkjonker:
### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?
```
$ dcos --version
dcoscli.version=0.5.7
dcos.version=1.10.4
dcos.commit=2e05da6b45f9f0e486de4d65404a0081099861c8
```
When running `dcos cluster setup` or `dcos auth login`, the `--password-file` option only works if the file contains nothing but the password (newlines not permitted).
I wonder if this is intended behavior, as many common file editors such as vim will automatically add a newline to the end of each file. I would suggest stripping the contents of the file when reading it, as password usually cannot contain a newline character anyway.

bamarni:
Indeed that seems unfortunate, and I don't think that passwords ending with a newline are actually a thing (or at least I hope not :sweat_smile:), see #1168

PR #1168, opened 2018-03-06
Right-strip newlines when reading a secure file

bamarni:
This function is called when reading a password or service account key file.

It common to have a trailing newline in such files, it shouldn't be considered as part of the credentials.

https://github.com/dcos/dcos-cli/issues/1167


python/lib/dcos/tests/test_util.py
```diff
@@ -16,6 +16,14 @@ def test_open_file():
         in str(excinfo.value)
 
 
+def test_read_file_secure_with_trailing_newline():
```
adamtheturtle: Please add a test with multiple newlines at the end. I imagine that this is also easily accidentally done.

dirkjonker: Definitely a good test case. Should pass without modifications when using `rstrip` https://docs.python.org/3/library/stdtypes.html?highlight=rstrip#str.rstrip

bamarni:
Thanks for the feedback, I've addressed your comments.

bamarni:
run linux integration tests

python/lib/dcos/dcos/util.py
```diff
@@ -280,7 +280,7 @@ def read_file_secure(path):
     enforce_file_permissions(path)
 
     with open_file(path) as file_:
-        return file_.read()
+        return file_.read().rstrip()
```
dirkjonker: Perhaps only strip newline-characters, e.g. `.rstrip('\r\n')` will not strip tabs or spaces, which may be valid password characters?

bamarni: That's what I had in the beginning then followed @jgehrcke's suggestion, I think both are fine.

If users are bit by this I'd be interested to know why they think putting passwords with trailing spaces / tabs in files and expecting them to be processed successfully by other users / tools is a good idea :confused:

PR #1170, opened 2018-03-09
Config improvements

bamarni:
- Add spf13/afero as a dependency
- Display code coverage when running tests
- Use afero in the config package 
- Move the Save method from Config to Store


bamarni:
run linux integration tests

PR #1171, opened 2018-03-12
Implement config marshalling

bamarni:
### Summary

- Implement config marshalling and saving
- Make small improvements to the Store
- Give more context to the config design document

#### Create a new config

``` go
conf := config.New()

// Set the DC/OS URL and ACS token.
conf.SetURL("https://dcos.example.com")
conf.SetACSToken("token_ABC")

// Set a path for the config and save it.
conf.SetPath("/path/to/config.toml")
err := conf.Save()
if err != nil {
	// Handle error.
}
```

#### Read / update a config from a path

``` go
conf, err := config.FromPath("/path/to/config.toml")
if err != nil {
	// Handle error.
}

// Change the cluster name.
conf.SetClusterName("my-new-cluster-name")

// Save the config file.
err = conf.Save()
if err != nil {
	// Handle error.
}
```


pkg/config/marshal.go
```diff
@@ -137,3 +119,48 @@ func UnmarshalTLS(val interface{}) TLS {
 		RootCAs:     certPool,
 	}
 }
+
+// Marshal updates the Store with Config dirty fields.
+func Marshal(conf *Config, store *Store) {
```
ArmandGrillet: In https://golang.org/pkg/encoding/json/#Marshal, `Marshal` returns a new Store instead of taking it as a parameter and customizes it. Should we do the same?

bamarni: Interesting, it's possible there because they can generate the JSON bytes from scratch. In our case we can't always do so because we might have an already existing store with a path and potential config data that we only need to update.

PR #1173, opened 2018-03-13
Config tweaks

bamarni:
- Break once a masked key is found in `config show`
- Clear dirty config fields after Marshalling
- Switch Config.timeout from int to time.Duration


bamarni:
@ArmandGrillet : thanks for the review, there is already a comment, let met know if that's not clear enough.

https://github.com/dcos/dcos-cli/blob/master/pkg/config/config.go#L149-L150

PR #1174, opened 2018-03-13
[WIP] Add HTTP client.

ArmandGrillet:
Only accepts GET and POST requests.


pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+	httpClient := &http.Client{
+		Transport: transport,
+		Timeout:   time.Duration(conf.Timeout()) * time.Second,
+	}
+
+	return &Client{conf.URL(), make(http.Header), httpClient}
```
bamarni: I think it is clearer to use named fields and have it as multi-line.

pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (c *Client) Get(endpoint string) (resp *http.Response, err error) {
```
bamarni: `endpoint` -> `path`
`endpoint` suggests this might as well be a full URL.

pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+	req, err := http.NewRequest("GET", c.baseURL+endpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+	return c.do(req)
+}
+
+func (c *Client) Post(endpoint string, body io.Reader) (resp *http.Response, err error) {
```
bamarni: Shall we use the same signature as the stdlib client method? https://golang.org/pkg/net/http/#Client.Post

pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+	req, err := http.NewRequest("POST", c.baseURL+endpoint, body)
+	if err != nil {
+		return nil, err
+	}
+	return c.do(req)
+}
+
+func (c *Client) do(req *http.Request) (resp *http.Response, err error) {
+	req.Close = true
```
bamarni: What is the reason for setting this to false? This disallows re-using the same underlying TCP connection for future requests, which shouldn't be the case imo.

pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
```
bamarni: In order to support APIs potentially relying on different MIME types than JSON, we should leave such decisions to the caller. Also, `Content-type` is not to be used in GET requests for example.

How about exporting the `Do` method, and have `func (c *Client) NewRequest(method, url string, body io.Reader) (*Request, error)` that callers can use to customize the request as they wish?

pkg/http/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+		w.Write([]byte(query))
+	}))
+	defer ts.Close()
+
+	conf, err := config.FromString(`
+[core]
+
+dcos_url = "` + ts.URL + `"
+`)
+	require.NoError(t, err)
```
bamarni: Please use `conf := config.New()` and `conf.SetURL(ts.URL)`. Here we should focus on testing the http client's Get method.

pkg/http/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	client := New(conf)
+
+	resp, err := client.Get("/")
```
bamarni: Similarly to what is done for TestPost, we should make assertions in the http test server handler to make sure the client sends the expected request (a `GET` request to the `/` path).

pkg/http/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestPost(t *testing.T) {
+	type query_struct struct {
+		Cluster string
+	}
+	query := `{"cluster":"DC/OS"}`
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
+			t.Errorf("Expected ‘POST’ request, got ‘%s’", r.Method)
+		}
```
bamarni: As we started using it and for consistency across our tests, let's use testify here. It turns these in a one-liner :slightly_smiling_face:

pkg/http/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/dcos/dcos-cli/pkg/config"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGet(t *testing.T) {
+	query := `{"cluster":"DC/OS"}`
```
bamarni: `query` -> `expectedBody`?

pkg/http/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+		require.Equal(t, query_sent.Cluster, "DC/OS")
+	}))
+	defer ts.Close()
+
+	conf, err := config.FromString(`
+[core]
+
+dcos_url = "` + ts.URL + `"
+`)
+	require.NoError(t, err)
```
bamarni: Same as my previous comment, this shouldn't be the place to test config parsing.

bamarni:
I also assumed the client would automatically set the Authorization header with the ACS token. Could you add a design page for this package? Even if it's minimal this would at least make sure we are all on the same track about the scope of our HTTP client.

pkg/http/client.go
```diff
@@ -0,0 +1,62 @@
...
+		tlsConfig = &tls.Config{InsecureSkipVerify: true}
+	} else {
+		tlsConfig = &tls.Config{InsecureSkipVerify: conf.TLS().Insecure, RootCAs: conf.TLS().RootCAs}
+	}
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+	httpClient := &http.Client{
+		Transport: transport,
+		Timeout:   time.Duration(conf.Timeout()) * time.Second,
```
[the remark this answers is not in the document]
ArmandGrillet: Yes, while keeping `SetTimeout(timeout int)`.

pkg/config/config.go
```diff
@@ -58,8 +59,8 @@ func (conf *Config) SetTLS(tls TLS) {
 }
 
 // Timeout returns the HTTP request timeout in seconds once the connection is established.
-func (conf *Config) Timeout() int {
-	return conf.timeout
+func (conf *Config) Timeout() time.Duration {
+	return time.Duration(conf.timeout) * time.Second
```
bamarni: CI failed with : 
```
--- FAIL: TestFromFullConfigString (0.00s)
	assertions.go:247: 
                          
	Error Trace:	config_test.go:76
		
	Error:      	Not equal: 
		
	            	expected: int(15)
		
	            	actual  : time.Duration(15000000000)
		
	Test:       	TestFromFullConfigString
```

Additionally, moving from an int to a duration should also be done in the getter and associated field. This would make our domain object cleaner as callers won't have to think in seconds but stick to the proper Go type.

pkg/httpclient/client.go
```diff
@@ -0,0 +1,70 @@
...
+func (c *Client) NewRequest(method, path string, body io.Reader) (req *http.Request, err error) {
+	return http.NewRequest(method, c.baseURL+path, body)
+}
+
+// Do sends an HTTP request and returns an HTTP response, following
+// policy (such as redirects, cookies, auth) as configured on the
+// client. Also sends the authorization header with the ACS token
+// to work with the DC/OS cluster we are linked to.
+func (c *Client) Do(req *http.Request) (resp *http.Response, err error) {
+	req.Header.Set("Authorization", "token="+c.acsToken)
```
bamarni: I'd do this in NewRequest.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+		if r.Method != "GET" {
+			t.Errorf("Expected `GET` request, got `%s`", r.Method)
+		}
+		if r.URL.EscapedPath() != "/get" {
+			t.Errorf("expected request to `/get, got `%s`", r.URL.EscapedPath())
+		}
+
+		w.Write([]byte(body))
+	}))
+	defer ts.Close()
```
bamarni: This test is essentially a duplicate of `TestGet`. As we are testing `NewRequest` here, we shouldn't do any request, just assert that the returned request has the expected method and URL set.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		decoder := json.NewDecoder(r.Body)
+		var body_sent body_struct
+		err := decoder.Decode(&body_sent)
+		require.NoError(t, err)
+		defer r.Body.Close()
+
+		require.Equal(t, body_sent.Cluster, "DC/OS")
+	}))
+	defer ts.Close()
```
bamarni: This test is essentially a duplicate of `TestPost`. As we are testing `NewRequest` here, we shouldn't do any request, just assert that the returned request has the expected method and URL set.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
+			t.Errorf("Expected `POST` request, got `%s`", r.Method)
+		}
+		if r.URL.EscapedPath() != "/post" {
+			t.Errorf("expected request to `/post, got `%s`", r.URL.EscapedPath())
+		}
+
+		decoder := json.NewDecoder(r.Body)
+		var body_sent body_struct
```
bamarni: I forgot to add linting in the CI, it complains here as underscores vars shouldn't be used. There are also 3 other locations in this file.

pkg/config/config.go
bamarni: If you want you can stick to int for this PR and I'll do a follow-up PR to change the config package.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+		err := decoder.Decode(&body_sent)
+		require.NoError(t, err)
+		defer r.Body.Close()
```
bamarni: This is not needed : 
```
// The Server will close the request body. The ServeHTTP
// Handler does not need to.
```

pkg/httpclient/client.go
bamarni: The reasoning being that this client might also be created with no ACS token yet, and later on one would want to log in and make a request with the returned token, but here it would always be overwritten with an empty one.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		require.Equal(t, body_sent.Cluster, "DC/OS")
```
bamarni: This introduces a race condition as it's not executed in the test's goroutine. It should use `assert` instead of `require`.

pkg/httpclient/client.go
bamarni: There should also be an if condition to only send this header if the acs token is not empty.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,144 @@
...
+func TestPost(t *testing.T) {
+	type bodyStruct struct {
+		Cluster string
+	}
+	body := `{"cluster":"DC/OS"}`
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
+			t.Errorf("Expected `POST` request, got `%s`", r.Method)
+		}
```
bamarni: All 3 liners like this in test http handlers can become one-liners, eg. `assert.Equal(t, "POST", r.Method)`

pkg/httpclient/client.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", contentType)
+	return c.Do(req)
+}
+
+// NewRequest returns a new Request given a method, path, and optional body.
+func (c *Client) NewRequest(method, path string, body io.Reader) (req *http.Request, err error) {
+	req, err = http.NewRequest(method, c.baseURL+path, body)
```
bamarni: This assumes `baseURL` doesn't contain a trailing slash, but we don't enforce it in the code. We should either right trim any slash when creating the client, or normalize the config value not to contain the trailing slash.

pkg/httpclient/client_test.go
```diff
@@ -0,0 +1,144 @@
...
+	defer ts.Close()
+
+	conf := config.New()
+	conf.SetURL(ts.URL)
+	conf.SetACSToken("secret")
+
+	client := New(conf)
+
+	_, err := client.Get("/get")
+	require.NoError(t, err)
```
bamarni: If the `client.Get()` method were not doing any HTTP request at all (eg. just `return nil, nil`), the test would still succeed.

We are making assertions in the test HTTP handler func without making sure that the handler function is actually called. This can be done by either returning a response with a well-known message and asserting that the client returns the expected reponse, or in a more generic way by having the handler setting a flag or something indicating it has been called.

bamarni:
As we will have quite some tests around HTTP across the whole CLI. It might be worth investigating on possible helper packages for it, or creating our own. With this PR we already seem to have some boilerplate code in tests (related to the httptest server).

PR #1176, opened 2018-03-15
CI: Run syntax and unit tests on the linux-binary check

bamarni:
It allows to get quicker feedback if something is obviously broken without the need to wait for linux-integration tests to be triggered and the costly AWS cluster to be created.

### DIFF
--- a/ci/Jenkinsfile-linux-binary
+++ b/ci/Jenkinsfile-linux-binary
@@ -43,10 +43,31 @@ def binaryBuilder = builder('py35', '/workspace')({
         }
     }
 
-    stage ("Build dcos-cli binary") {
-        dir('dcos-cli/python/lib/dcoscli') {
+    dir('dcos-cli/python/lib/dcoscli') {
+        stage ("Build dcos-cli binary") {
             sh "make binary"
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile-linux-binary
```diff
@@ -43,11 +43,32 @@ def binaryBuilder = builder('py35', '/workspace')({
...
             sh "make binary"
             sh "dist/dcos"
         }
+
+        stage ("Run syntax checks on dcoscli") {
+            sh "make env"
+            sh '''
+                bash -c " \
+                  source env/bin/activate && \
+                  tox -e py35-syntax"
```
bamarni: I've left out the dcoscli unit tests as they unfortunately currently require a configured cluster.

ArmandGrillet: Do we have an open ticket regarding this? Unit tests should be able to run without a configured cluster in the future 🤔.

bamarni: Indeed :+1:  I've created https://jira.mesosphere.com/browse/DCOS_OSS-2270

PR #1181, opened 2018-03-23
 Use mac-hh-yosemite Jenkins node for MacOS builds

bamarni:
According to PyInstaller requirements :

> PyInstaller builds apps that are compatible with the Mac OS X release in which you run it, and following releases.

Building from an old enough Mac makes sure we support all versions onwards. The latest Yosemite version was released 3 years ago, so it already brings a wide range of supported MacOS versions for the CLI.

The Jenkins node has Python 3.5.5 compiled with OpenSSL 1.0.2n.

https://jira.mesosphere.com/browse/DCOS_OSS-2290

### DIFF
--- a/ci/Jenkinsfile-mac-binary
+++ b/ci/Jenkinsfile-mac-binary
@@ -36,7 +36,7 @@ def builder(String nodeId, String workspace = null) {
 /**
  * Run a builder to create the dcos-cli binary and execute it.
  */
-def binaryBuilder = builder('mac')({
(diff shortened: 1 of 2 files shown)


bamarni:
run mac integration tests

bamarni:
run all integration tests

bamarni:
Not yet, I will still try to run them on High Sierra with a binary built from Yosemite, as we didn't try this yet.

PR #1187, opened 2018-03-27
Auth list providers

br-lewis:
Adds the `list-providers` command to the `auth` subcommand. It also makes some changes to how the CLI finds the config file in the new `clusters/...` directory structure.


pkg/cmd/dcos.go
```diff
@@ -9,16 +9,20 @@ import (
 
 	"github.com/dcos/dcos-cli/pkg/config"
 	"github.com/spf13/cobra"
+	"io"
```
bamarni: this should go in the previous import section (https://github.com/golang/go/wiki/CodeReviewComments#imports).

pkg/cmd/dcos.go
```diff
@@ -9,16 +9,20 @@ import (
...
 )
 
 // rootCmd represents the base command when called without any subcommands.
 var rootCmd = &cobra.Command{
 	Use: "dcos",
 }
 
+var dcosConfig *Cluster
```
bamarni: we should avoid global variables in general, I know there are already some but this will be refactored. How about calling `attachedCluster()` at the beginning of the list-provider command into a local variable?

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+	authCmd.AddCommand(authListProvidersCmd)
+	authListProvidersCmd.Flags().BoolVar(&jsonOutput, "json", false,
+		"returns providers in json format")
+}
+
+func listProviders(cmd *cobra.Command, args []string) {
+	providers, err := getProviders()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
bamarni: Instead of printing and calling exit like here or below, other commands define the `RunE` func field and it can return an error.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+const (
+	AuthTypeDCOSUidPassword = "dcos-uid-password"
+	AuthTypeDCOSUidServiceKey = "dcos-uid-servicekey"
+	AuthTypeDCOSUidPasswordLDAP = "dcos-uid-password-ldap"
+	AuthTypeSAMLSpInitiated = "saml-sp-initiated"
+	AuthTypeOIDCAuthCodeFlow = "oidc-authorization-code-flow"
+	AuthTypeOIDCImplicitFlow = "oidc-implicit-flow"
+)
+
+var jsonOutput bool
```
bamarni: This would require having constructor-based commands rather than global variables like `authListProviders`, but once this is done that'd be nice to have structs for options. Similar to the pattern here : https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/auth/cani.go#L39-L53

(I'm leaving this comment for reference not necessarily to tackle in this PR).

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+
+// These are the different auth types that DC/OS supports with the names that they'll be given from the providers
+// endpoint.
+const (
+	AuthTypeDCOSUidPassword = "dcos-uid-password"
+	AuthTypeDCOSUidServiceKey = "dcos-uid-servicekey"
+	AuthTypeDCOSUidPasswordLDAP = "dcos-uid-password-ldap"
+	AuthTypeSAMLSpInitiated = "saml-sp-initiated"
+	AuthTypeOIDCAuthCodeFlow = "oidc-authorization-code-flow"
+	AuthTypeOIDCImplicitFlow = "oidc-implicit-flow"
```
bamarni: JP suggests using the "login provider" semantic here, as authentication is only about sending the ACS token to DC/OS. We still have quite some docs using "auth provider" wording, but I'd rather use the wording recommended by the security team.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+		desc = fmt.Sprintf("Log in using SAML 2.0 (%s)", provider.Description)
+	} else if authType == AuthTypeOIDCAuthCodeFlow || authType == AuthTypeOIDCImplicitFlow {
+		desc = fmt.Sprintf("Log in using OpenID Connect(%s)", provider.Description)
+	} else {
+		return "", errors.New("unknown authentication type")
+	}
+	return desc, nil
+}
+
+type authProvider struct {
```
bamarni: As commented above : `LoginProvider`, this would also make sense to create an `auth` package as this would be used in other places (eg. cluster setup).

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+	} else if authType == AuthTypeDCOSUidServiceKey {
+		desc = "Log in using a DC/OS service user account (username and private key)"
+	} else if authType == AuthTypeDCOSUidPasswordLDAP {
+		desc = "Log in in using an LDAP user account (username and password)"
+	} else if authType == AuthTypeSAMLSpInitiated {
+		desc = fmt.Sprintf("Log in using SAML 2.0 (%s)", provider.Description)
+	} else if authType == AuthTypeOIDCAuthCodeFlow || authType == AuthTypeOIDCImplicitFlow {
+		desc = fmt.Sprintf("Log in using OpenID Connect(%s)", provider.Description)
+	} else {
+		return "", errors.New("unknown authentication type")
```
bamarni: A switch statement would be more readable here.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+			table.Append([]string{name, desc})
+		}
+		table.Render()
+	}
+}
+
+func getProviders() (*map[string]authProvider, error) {
+	var config = dcosConfig.Config
+	var client = httpclient.New(config)
+	var response, err = client.Get("/acs/api/v1/auth/providers")
```
bamarni: the short variable declaration is more common in go, that's what we currently use, eg. `response, err := client.Get("/acs/api/v1/auth/providers")`

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+	AuthTypeSAMLSpInitiated = "saml-sp-initiated"
+	AuthTypeOIDCAuthCodeFlow = "oidc-authorization-code-flow"
+	AuthTypeOIDCImplicitFlow = "oidc-implicit-flow"
+)
+
+var jsonOutput bool
+
+// authCmd represents the `dcos auth` subcommand.
+var authListProvidersCmd = &cobra.Command{
+	Use: "list-providers",
```
bamarni: `dcos auth list-providers` accepts an optional parameter, which refers to the cluster URL to get providers from instead of using the attached cluster's URL.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
+package cmd
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+	"os"
```
bamarni: imports should be sorted as mentioned before.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+// authCmd represents the `dcos auth` subcommand.
+var authListProvidersCmd = &cobra.Command{
+	Use: "list-providers",
+	Run: listProviders,
+}
+
+func init() {
+	authCmd.AddCommand(authListProvidersCmd)
+	authListProvidersCmd.Flags().BoolVar(&jsonOutput, "json", false,
+		"returns providers in json format")
```
bamarni: There is no line-length limit in Go, we use ~110 characters as a rule of thumb, a single line is fine here.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+
+func listProviders(cmd *cobra.Command, args []string) {
+	providers, err := getProviders()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if jsonOutput {
+		// re-marshal it into json with indents added in for pretty printing
```
bamarni: We use sentences starting with uppercase and ending with a dot. The cmd package was a quick proof of concept but we started doing this in other packages.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+			table.Append([]string{name, desc})
+		}
+		table.Render()
+	}
+}
+
+func getProviders() (*map[string]authProvider, error) {
+	var config = dcosConfig.Config
+	var client = httpclient.New(config)
```
bamarni: `httpclient` has been changed recently: https://github.com/dcos/dcos-cli/commit/3560d639cb5926bee89e23b5b728d05f350c7c35

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	defer response.Body.Close()
+
+	var resp map[string]authProvider
+	err = json.NewDecoder(response.Body).Decode(&resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return &resp, nil
```
bamarni: nit, this is enough as callers must check for the error and not for the pointer to be nil :
``` go 
err = json.NewDecoder(response.Body).Decode(&resp)
return &resp, err
```

br-lewis:
Build failed due to a linter error (used `Url` instead of `URL` in a variable name) but why did only 1 fail?

bamarni:
> Build failed due to a linter error (used Url instead of URL in a variable name) but why did only 1 fail?

I added Go tests on the linux builder Jenkinsfile, I will work on creating its own status check for the new CLI.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,124 @@
...
+	RunE: listProviders,
+}
+
+func init() {
+	authCmd.AddCommand(authListProvidersCmd)
+	authListProvidersCmd.Flags().BoolVar(&jsonOutput, "json", false, "returns providers in json format")
+}
+
+func listProviders(cmd *cobra.Command, args []string) error {
+	conf := attachedCluster().Config
```
bamarni: This is unnecessary when an argument is passed, could you execute this only when no arg is passed?

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -0,0 +1,124 @@
...
+	var resp map[string]loginProvider
+	err = json.NewDecoder(response.Body).Decode(&resp)
+	return &resp, err
+}
+
+func loginTypeDescription(loginType string, provider loginProvider) (string, error) {
+	var desc string
+	switch loginType {
+	case LoginTypeDCOSUidPassword:
+		desc = "Log in using a standard DC/OS user account (username and password)"
```
bamarni: nit: returning directly from switch cases removes the need for the lines surrounding the switch block.

PR #1189, opened 2018-03-27
Fixed pattern matching for task IDs in 'task exec'.

klueska:
Previously, there were problems with the way patterns were matched for
the task ID when calling `task exec`. Specifically, it didn't follow the
semantics indicated in the help string for (1) matching regular
expressions and (2) matching any substring within the ID, not just the
prefix of the ID.

The reason we did not adhere to these semantics is because we were
trying to match the exact semantics of `docker exec`, which only allows
prefix matching, not arbitrary regex or 'contains()' like semantics for
pattern matching the ID.

With this change, we have decided to break from the docker semantics and
use the default pattern matching semantics provided by the dcos-cli
library command for pattern matching task ids.

Unfortunately, there is a bug in this library, and regular expressions
are not supported! However, arbitrary substring pattern mathing is.

This commit, therefore, updates the semantics of `task exec` to adhere
to 'contains()' like smeantics within the task ID for the 'task_id'
parameter passed to it, but does not yet support proper regular
expression pattern matching.

A subsequent commit will be added to fix regex matching in library call
implementing task ID pattern matching.

### DIFF
--- a/python/lib/dcos/dcos/mesos.py
+++ b/python/lib/dcos/dcos/mesos.py
@@ -493,33 +493,17 @@ class Master(object):
 
         return tasks
 
-    def get_container_id(self, task_id):
-        """Returns the container ID for a task ID matching `task_id`
+    def get_container_id(self, task_obj):
+        """Returns the container ID for a task.
 
-        :param task_id: The task ID which will be mapped to container ID
-        :type task_id: str
-        :returns: The container ID associated with 'task_id'
+        :param task__obj: The task which will be mapped to container ID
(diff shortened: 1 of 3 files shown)


klueska:
run linux integration tests

klueska:
run linux integration tests

klueska:
run linux integration tests

klueska:
run linux integration tests

klueska:
run mac integration tests

klueska:
run windows integration tests

PR #1191, opened 2018-03-28
Update `dcos task exec` help message about regexes

bamarni:
This makes it more precise as it accepts Unix file pattern wildcards instead of regexes.


bamarni:
run linux integration tests

PR #1193, opened 2018-04-01
Config refactoring

bamarni:
### Refactor the config package

This refines the goal of the package and adjusts the implementation plan accordingly.

The Config struct becomes Cluster and is moved into a `core` subpackage. Decoupling it from the main package makes the config agnostic from any domain model, rather, Cluster is just an example of how one can use a Config to fits their application needs.

The marshalling/unmarshalling logic has been removed in favor of a lighter proxy object (Cluster) which gets/sets values in the Store on the fly. Also, only relevant `core` and `cluster` fields now have getters/setters.

### Introduce a ConfigManager 

The config manager acts as a repository for configurations. It is currently only able to find configs but in the future it will be responsible for creating and removing configs as well.

This commit also adds several test fixtures to make sure the config manager behaves correctly under different scenarios and edge-cases.

### Introduce a CLI context

Introduce a Context struct. It is created based on various abstractions (eg. io.Reader for stdout/stderr) or by calling DefaultContext() which will be backed by the OS.

Once created, the context must be passed as an argument to each subcommand constructor and they must use it to interact with the environment (print output, read env vars or files, etc.) or instantiate environment-dependent structs from other packages.


pkg/config/config.go
```diff
@@ -1,201 +1,274 @@
...
+		}
+	}
 
-// setDirty sets a field as "dirty", this is used by the Marshal
-// function to know which fields must be persisted to the Store.
-func (conf *Config) setDirty(field interface{}) {
-	conf.dirtyFields[field] = true
+	// Fallback to the TOML tree if present.
+	switch node := c.tree.Get(key).(type) {
+	case *toml.Tree, []*toml.Tree:
```
ArmandGrillet: I don't see how could the value returned be of type `[]*toml.Tree` with https://github.com/pelletier/go-toml/blob/66540cf1fcd2c3aee6f6787dfa32a6ae9a870f12/toml.go#L78 🤔

pkg/config/config.go
```diff
@@ -1,201 +1,274 @@
...
-func New() Config {
-	return Config{
-		store:       NewStore(StoreOpts{}),
-		dirtyFields: make(map[interface{}]bool),
+	sort.Strings(keys)
+
+	// Remove duplicates, this happens when there is both an env var and a TOML key.
+	// Keys are already sorted so this is done by comparing each key with the next one
+	// and removing the latter if it's the same.
+	for i := 1; i < keysLen; i++ {
```
ArmandGrillet: If we are sure that `keys` contains only two identical values at max and that they are sorted, the loop can be simplified:
```
// a = ["a", "a", "b", "c", "c"]
for i := 0; i < len(a) - 1; i++ {
	if a[i] == a[i+1] {
		a = append(a[:i], a[i+1:]...)
	}
}
// a = ["a", "b", "c"]
```
If they can be more than 2 times the same value, a `i--` has to be added in the loop.

pkg/cli/cluster_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+func TestGetTLS(t *testing.T) {
+
+	expectedTLSInsecureSkipVerify := []struct {
+		val      interface{}
+		insecure bool
+	}{
+		{"True", false},
```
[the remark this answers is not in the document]
bamarni: I've tested values which make sense to me, `T` or `F` for eg. is very uncommon. For simplicity we've relied on ParseBool for now but I think we should move away from this and define what we want instead of depending on what a language does.

This is a topic I'd like to bring to the UX team, see https://jira.mesosphere.com/browse/DCOS-11460 for example.

pkg/cmd/dcos_auth_listproviders.go
```diff
@@ -21,62 +21,61 @@ const (
...
-		table.SetRowSeparator(" ")
-		table.SetColumnSeparator(" ")
-		table.SetCenterSeparator(" ")
+// newCmdAuthListProviders creates the `dcos auth list-providers` subcommand.
+func newCmdAuthListProviders(ctx *cli.Context) *cobra.Command {
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:  "list-providers",
+		Args: cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
bamarni: In this PR I've explicitly moved away from global variables and functions as much as possible. Having the command function inlined forces us to keep it thin as it is supposed to be, and move relevant things into their own packages.

This function is not a good example as we haven't extracted things from it yet so indeed it looks quite big.

pkg/config/config.go
bamarni: This happens with tables (cf. https://github.com/toml-lang/toml#array-of-tables). We don't support it in our CLI configuration so here I'd rather return `nil`.

pkg/config/config.go
bamarni: Good catch, indeed there cannot be more than 2 duplicates (one from env var and one from the config file), I've removed the `i--`.

PR #1196, opened 2018-04-02
Httpclient updates

bamarni:
- add a default transport
- set a default timeout of 3 minutes
- fix a flaky test
- add a functional option type for Requests

### DIFF
--- a/pkg/httpclient/client.go
+++ b/pkg/httpclient/client.go
@@ -1,15 +1,18 @@
 package httpclient
 
 import (
+	"context"
 	"io"
 	"net"
 	"net/http"
+	"time"
 )
 
 // A Client is an HTTP client.
 type Client struct {
 	acsToken   string
 	baseURL    string
+	timeout    time.Duration
 	baseClient *http.Client
 }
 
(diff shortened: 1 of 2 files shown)


bamarni:
run windows integration tests

pkg/httpclient/client_test.go
```diff
@@ -59,39 +60,70 @@ func TestPost(t *testing.T) {
...
+func TestNewRequestWithoutTimeout(t *testing.T) {
+	client := New("https://dcos.io", func(client *Client) {
+		client.timeout = 60 * time.Second
+	})
+
+	req, err := client.NewRequest("GET", "/path", nil, NoTimeout())
+	require.NoError(t, err)
+	deadline, ok := req.Context().Deadline()
+	require.False(t, ok)
+	require.True(t, deadline.IsZero())
```
jdef: why bother testing the value of `deadline` if `ok` is false?

pkg/httpclient/client_test.go
```diff
@@ -59,39 +60,70 @@ func TestPost(t *testing.T) {
...
+}
+
+func TestCancelRequest(t *testing.T) {
+	done := make(chan struct{})
+	stuckedHandler := make(chan struct{})
```
jdef: nit: s/stuckedHandler/stuckHandler/

pkg/httpclient/client.go
```diff
@@ -46,10 +85,18 @@ func (c *Client) Post(path string, contentType string, body io.Reader) (*http.Re
 	return c.Do(req)
 }
 
+// NoTimeout disables the timeout for an HTTP request.
+func NoTimeout() RequestOption {
+	return func(req *http.Request) {
+		ctx := context.WithValue(req.Context(), keyNoTimeout, true)
```
jdef: nit: I probably would have used `struct{}{}` instead of `true` since a value is not actually needed here

PR #1202, opened 2018-04-05
Rewrite Cosmos binary resource URLs when needed

### LINKED ISSUE #1190: dcos package install --cli is not working

### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?

```
[dcos-sandbox]~ $ dcos --version
dcoscli.version=0.6.0
dcos.version=1.11.0
dcos.commit=b6d6ad4722600877fde2860122f870031d109da3
dcos.bootstrap-id=a0654657903fb68dff60f6e522a7f241c1bfbf0f
[dcos-sandbox]~ $ 
```

osx

### What did you do?

```
[dcos-sandbox]~ $ dcos package install elastic --cli --yes
By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#certified-services
Installing CLI subcommand for package [elastic] version [2.2.0-5.6.5]
URL [http://admin.<cropped>.io/package/resource?url=https://downloads.mesosphere.com/elastic/assets/2.2.0-5.6.5/dcos-service-cli-darwin] is unreachable.
[dcos-sandbox]~ $ 
```

I was expecting the cli package to be installed.

bamarni:
Depending on the user setup, they might contain an incorrect scheme.
This is because Cosmos relies in the scheme used to query admin router.
If a CLI is configured to use HTTPS but a load balancer sits in front
of admin router and does TLS termination, these URLs will use the HTTP
scheme.

Not only that it is a security concern, it would simply fail on setups
which don't support HTTPS. This is hopefully a temporary fix until we
come-up with a long-term fix in DC/OS.

Fixes #1190
https://jira.mesosphere.com/browse/COPS-3052
https://jira.mesosphere.com/browse/DCOS_OSS-2325

### DIFF
--- a/python/lib/dcos/dcos/subcommand.py
+++ b/python/lib/dcos/dcos/subcommand.py
@@ -9,7 +9,9 @@ import stat
 import subprocess
 import sys
 import zipfile
+
 from distutils.version import LooseVersion
+from urllib.parse import urlparse
 
 from dcos import config, constants, http, util
 from dcos.errors import DCOSException
@@ -515,6 +517,10 @@ def _install_with_binary(
 
     binary_url, kind = binary_cli.get("url"), binary_cli.get("kind")
 
+    binary_url = _rewrite_binary_url(
+        binary_url,
+        config.get_config_val("core.dcos_url"))
+
     try:
         env_bin_dir = os.path.join(env_directory, BIN_DIRECTORY)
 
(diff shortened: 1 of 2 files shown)


python/lib/dcos/dcos/subcommand.py
```diff
@@ -567,6 +573,31 @@ def _install_with_binary(
...
 
 
+def _rewrite_binary_url(binary_url, dcos_url):
+    """ Rewrite Cosmos binary URLs when needed.
+
+    A rewrite happens when a resource refers to the currently configured
+    DC/OS cluster but uses the wrong scheme. Ideally this "hack" will be
+    removed in the future.
+
+    For more context, see https://jira.mesosphere.com/browse/COPS-3052.
```
ArmandGrillet: We should not mention a private ticket in a public repo. Updating [DCOS_OSS-2325](https://jira.mesosphere.com/browse/DCOS_OSS-2325) to add more context and of this comment line to give a public link would be enough.

bamarni: updated

PR #1208, opened 2018-04-13
Add `dcos cluster list` subcommand.

ArmandGrillet:
Same features as in the current DC/OS CLI, I have just renamed `CLUSTER ID` by `ID` in the output as the command already states that we list clusters.

Also abstracts the table creation so that we use the same for all our commands.

### DIFF
--- a/pkg/cli/context.go
+++ b/pkg/cli/context.go
@@ -71,6 +71,17 @@ func (ctx *Context) Cluster() (*Cluster, error) {
 	return NewCluster(conf), nil
 }
 
+// Clusters returns the clusters.
+func (ctx *Context) Clusters() []*Cluster {
+	confs := ctx.ConfigManager().All()
+	var clusters []*Cluster
+	for _, conf := range confs {
+		clusters = append(clusters, NewCluster(conf))
+	}
+
+	return clusters
+}
+
 // HTTPClient creates an httpclient.Client for a given cluster.
 func (ctx *Context) HTTPClient(c *Cluster, opts ...httpclient.Option) *httpclient.Client {
 	if c.Timeout() > 0 {
--- a/pkg/cmd/dcos_cluster_list.go
+++ b/pkg/cmd/dcos_cluster_list.go
@@ -1,0 +1,117 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/dcos/dcos-cli/pkg/cli"
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/spf13/cobra"
+)
+
+// newCmdClusterList lists the clusters.
+func newCmdClusterList(ctx *cli.Context) *cobra.Command {
+	var onlyAttached bool
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:  "list",
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clusters := ctx.Clusters()
+
+			currentCluster, err := ctx.Cluster()
+			var currentClusterID string
+			if err == nil {
+				currentClusterID = currentCluster.ID()
+			}
+
+			// infos is composed of clusters information: attachment, name, id, status, version, and url.
+			infos := []clusterInfo{}
+			var mu sync.Mutex
+
+			var wg sync.WaitGroup
+			for _, cluster := range clusters {
+				// Add should execute before the statement creating the goroutine.
+				wg.Add(1)
+				go func(cluster *cli.Cluster) {
+					defer wg.Done()
+
+					var info clusterInfo
+					info.Name = cluster.Name()
+					info.ID = cluster.ID()
+					if info.ID == currentClusterID {
+						info.Attached = true
+					} else if onlyAttached {
+						return
+					}
+					info.URL = cluster.URL()
+					info.Status = "AVAILABLE"
+					info.Version, err = getVersion(ctx, cluster)
+					if err != nil {
+						info.Status = "UNAVAILABLE"
+					}
+
+					mu.Lock()
+					infos = append(infos, info)
+					mu.Unlock()
+				}(cluster)
+			}
+			wg.Wait()
+
+			if jsonOutput {
+				// Re-marshal it into json with indents added in for pretty printing.
+				out, err := json.MarshalIndent(infos, "", "\t")
+				if err != nil {
+					return err
+				}
+				fmt.Fprintln(ctx.Out, string(out))
+			} else {
+				table := cli.NewTable(ctx.Out, []string{"", "NAME", "ID", "STATUS", "VERSION", "URL"})
+				for i := range infos {
+					var attached string
+					if infos[i].Attached {
+						attached = "*"
+					}
+					table.Append([]string{attached, infos[i].Name, infos[i].ID, infos[i].Status, infos[i].Version, infos[i].URL})
+				}
+				table.Render()
+			}
+			return nil
+		},
+	}
+	cmd.Flags().BoolVar(&onlyAttached, "attached", false, "returns attached cluster only")
+	cmd.Flags().BoolVar(&jsonOutput, "json", false, "returns clusters in json format")
+	return cmd
+}
+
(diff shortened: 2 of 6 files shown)


ArmandGrillet:
run all integration tests

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/spf13/cobra"
+)
+
+// newCmdClusterList lists the clusters.
+func newCmdClusterList(ctx *cli.Context) *cobra.Command {
+	var onlyAttached bool
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:  "list",
+		Args: cobra.MaximumNArgs(1),
```
bamarni: `cobra.NoArgs`

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			manager := ctx.ConfigManager()
+			confs := manager.All()
+
+			currentConf, err := manager.Current()
+			if err != nil {
+				return err
+			}
+			currentConfID := filepath.Base(filepath.Dir(currentConf.Path()))
```
bamarni: As we currently have `Cluster()`, what about creating a `Clusters()` method in the context too? Then the Cluster struct can expose an `ID()` method.

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+			clusters := [][]string{}
+
+			var wg sync.WaitGroup
+			for _, conf := range confs {
+				// Add should execute before the statement creating the goroutine.
+				wg.Add(1)
+				go func(conf *config.Config) {
+					defer wg.Done()
+
+					name := conf.Get("cluster.name").(string)
```
bamarni: This would panic if that config is not a string, if we switch to cluster structs as mentioned, it can use the explicit getters (eg. `cluster.URL()`).

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+	cmd := &cobra.Command{
+		Use:  "list",
+		Args: cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			manager := ctx.ConfigManager()
+			confs := manager.All()
+
+			currentConf, err := manager.Current()
+			if err != nil {
+				return err
```
bamarni: Having no attached cluster shouldn't trigger an error, we would then just not display any star in the list.

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+					} else if onlyAttached {
+						return
+					}
+					url := conf.Get("core.dcos_url").(string)
+					status := "AVAILABLE"
+					version, err := getVersion(url)
+					if err != nil {
+						status = "UNAVAILABLE"
+					}
+					clusters = append(clusters, []string{name, id, status, version, url})
```
bamarni: read and writes to `clusters` happen concurrently, it needs to be protected with a mutex.

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+				}(conf)
+			}
+			wg.Wait()
+
+			if jsonOutput {
+				// Re-marshal it into json with indents added in for pretty printing.
+				out, err := json.MarshalIndent(clusters, "", "\t")
```
bamarni: This creates a JSON with arrays, here the command is supposed to print explicit fields, eg : 
```
[
  {
    "attached": true,
    "cluster_id": "9fa2d099-b3e3-4d69-9156-67a8ccdff961",
    "name": "DC/OS CLI Integration Tests",
    "status": "UNAVAILABLE",
    "url": "https://34.209.232.52",
    "version": "N/A"
  }
]
```

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,106 @@
...
+	return cmd
+}
+
+// getVersion returns the version of the cluster as a string.
+func getVersion(url string) (string, error) {
+	if url == "" {
+		return "UNKNOWN", errors.New("cluster URL undefined")
+	}
+
+	client := httpclient.New(url)
```
bamarni: The client here should respect the cluster's TLS config, we should also probably set a timeout (currently it's 5 seconds).

br-lewis: Maybe it would be better to add a Client method to Clusters to create a client using that cluster's config?

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,122 @@
...
+				go func(cluster *cli.Cluster) {
+					defer wg.Done()
+
+					var info clusterInfo
+					info.Name = cluster.Name()
+					info.ID = cluster.ID()
+					if info.ID == currentClusterID {
+						info.Attached = true
+						if !jsonOutput {
+							info.Name += "*"
```
bamarni: As discussed yesterday, could this be in an empty column at the beginning instead?

pkg/cli/context.go
```diff
@@ -71,6 +71,17 @@ func (ctx *Context) Cluster() (*Cluster, error) {
 	return NewCluster(conf), nil
 }
 
+// Clusters returns the clusters.
+func (ctx *Context) Clusters() ([]*Cluster, error) {
+	confs := ctx.ConfigManager().All()
+	clusters := []*Cluster{}
```
bamarni: nit: `var clusters []*Cluster`

A nil slice is similar here but saves an unnecessary allocation.

pkg/cli/context.go
```diff
@@ -71,6 +71,17 @@ func (ctx *Context) Cluster() (*Cluster, error) {
 	return NewCluster(conf), nil
 }
 
+// Clusters returns the clusters.
+func (ctx *Context) Clusters() ([]*Cluster, error) {
```
bamarni: `func (ctx *Context) Clusters() []*Cluster {` ?

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// getVersion returns the version of the cluster as a string.
+func getVersion(ctx *cli.Context, cluster *cli.Cluster) (string, error) {
+	if cluster.URL() == "" {
+		return "UNKNOWN", errors.New("cluster URL undefined")
+	}
+
+	client := ctx.HTTPClient(cluster)
+	resp, err := client.Get("/dcos-metadata/dcos-version.json")
```
bamarni: This might hang for some time under some circumstances, could we hardcode a 5 seconds timeout here?

`client := ctx.HTTPClient(cluster, httpclient.Timeout(5 * time.Second))` on the line above.

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,122 @@
...
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clusters, err := ctx.Clusters()
+			if err != nil {
+				return err
+			}
+
+			currentCluster, err := ctx.Cluster()
+			if err != nil {
+				return err
```
bamarni: This shouldn't stop the command's execution with an error, it is fine to continue with an empty currentClusterID here.

pkg/cmd/dcos_cluster_list.go
```diff
@@ -0,0 +1,122 @@
...
+				table := cli.NewTable(ctx.Out, []string{"NAME", "ID", "STATUS", "VERSION", "URL"})
+				for i := range infos {
+					table.Append([]string{infos[i].Name, infos[i].ID, infos[i].Status, infos[i].Version, infos[i].URL})
+				}
+				table.Render()
+			}
+			return nil
+		},
+	}
+	cmd.Flags().BoolVar(&onlyAttached, "attached", false, "returns attached clusters only")
```
bamarni: nit: `returns attached cluster only`

PR #1211, opened 2018-04-17
Accept an --all flags for `dcos cluster remove`

bamarni:
This flags removes all configured clusters.

### DIFF
--- a/pkg/cmd/dcos_cluster_remove.go
+++ b/pkg/cmd/dcos_cluster_remove.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/dcos_cluster_remove.go
```diff
@@ -9,16 +10,38 @@ import (
...
+				return errors.New("either a cluster name or the --all option must be passed")
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			conf, err := ctx.ConfigManager().Find(args[0], false)
-			if err != nil {
-				return err
+			// Remove a single cluster.
+			if len(args) == 1 {
```
ArmandGrillet: Having an `if removeAll {` to handle the two cases would be more straightforward here.

PR #1212, opened 2018-04-17
Add logging

bamarni:
- Add logrus as a dependency
- Add flags for verbose outputs


pkg/cmd/dcos.go
```diff
@@ -3,14 +3,29 @@ package cmd
...
+				ctx.Logger.SetLevel(logrus.InfoLevel)
+			} else if verbose > 1 {
+				// -vv sets the logger level to debug. This also happens for -vvv
+				// and above, in such cases we set the logging level to its maximum.
+				ctx.Logger.SetLevel(logrus.DebugLevel)
+			}
+		},
 	}
+
+	cmd.PersistentFlags().CountVarP(&verbose, "", "v", "verbose (-v, or -vv)")
```
ArmandGrillet: s/`verbose (-v, or -vv)`/`verbosity (-v or -vv)`. Also, the current output `-v, -- count   verbose (-v, or -vv)` is not very meaningful (what does `-- count` means?), is there any way we can customize it?

Gopkg.lock
```diff
@@ -55,14 +61,14 @@
 [[projects]]
   name = "github.com/spf13/cobra"
   packages = ["."]
-  revision = "7b2c5ac9fc04fc5efafb60700713d4fa609b777b"
-  version = "v0.0.1"
+  revision = "a1f051bc3eba734da4772d60e2d677f47cf93ef4"
+  version = "v0.0.2"
```
ArmandGrillet: Why are we doing the update in this commit? Is it necessary for logrus?

PR #1213, opened 2018-04-18
Print a clear error when no cluster is attached

bamarni:
This came-up multiple times, when running `dcos config set` commands and
no cluster is currently attached, a stack trace is shown :

  TypeError: join() argument must be str or bytes, not 'NoneType'

This change prevents that stack trace to be shown.

https://jira.mesosphere.com/browse/COPS-2857
https://jira.mesosphere.com/browse/DCOS_OSS-2319


bamarni:
run all integration tests

bamarni:
run mac integration tests

PR #1215, opened 2018-04-19
Changed order of types parsed by jsonitem.

ArmandGrillet:
Without this, `dcos marathon app update test-app env='{"APISERVER_PORT":"25502"}'` would consider `env` as a string instead of a JSON. 

This solves both commands displayed in https://docs.mesosphere.com/1.11/deploying-services/update-user-service/ that do not work at the moment.


ArmandGrillet:
run all integration tests

ArmandGrillet:
You're right, I'm working on a test for it and should be there by tomorrow. 

Shall we add as a feature the possibility to modify one key of a JSON? Currently changing `env` would replace the old `end` by the value given, there is no way of changing only one specific key in a JSON.

ArmandGrillet:
run all integration tests

PR #1216, opened 2018-04-20
Add logs for HTTP requests and responses

bamarni:
This relies on the httputil package to dump requests and reponses, the INFO logging level displays requests/responses without the body while the DEBUG level displays them.

Eg. :
```
$ dcos -v auth list-providers https://dcos.example.com
INFO[0000] Outgoing request:
GET /acs/api/v1/auth/providers HTTP/1.1
Host: dcos.example.com
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip

INFO[0000] Incoming response:
HTTP/1.1 200 OK
Content-Length: 531
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Fri, 20 Apr 2018 13:28:19 GMT
Server: openresty

   PROVIDER ID                           AUTHENTICATION TYPE
  dcos-users     Log in using a standard DC/OS user account (username and password)
  dcos-services  Log in using a DC/OS service user account (username and private key)
```

### DIFF
--- a/pkg/httpclient/client.go
+++ b/pkg/httpclient/client.go
@@ -6,7 +6,10 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httputil"
 	"time"
+
+	"github.com/sirupsen/logrus"
 )
 
 // A Client is an HTTP client.
@@ -15,6 +18,7 @@ type Client struct {
 	baseURL    string
 	timeout    time.Duration
 	baseClient *http.Client
+	logger     *logrus.Logger
 }
 
 // Option is a functional option for an HTTP client.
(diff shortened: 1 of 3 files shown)


pkg/httpclient/client.go
```diff
@@ -145,5 +156,33 @@ func (c *Client) NewRequest(method, path string, body io.Reader, opts ...Request
 // policy (such as redirects, cookies, auth) as configured on the
 // client.
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
-	return c.baseClient.Do(req)
+	var dumpBody bool
+	var logFn func(format string, args ...interface{})
```
ArmandGrillet: What about `logFn := c.logger. Infof ` and then the `if c.logger != nil && c.logger.Level >= logrus.DebugLevel {`?

PR #1219, opened 2018-04-23
Update Golint dependency.

ArmandGrillet:


bamarni:
cf. https://github.com/golang/lint/issues/396

PR #1220, opened 2018-04-24
Fix bad rebase

bamarni:
- Use ctx.Logger() instead of ctx.Logger
- Update an incorrect log message


ArmandGrillet:
run linux-go-tests

PR #1224, opened 2018-04-25
Support the auth header challenge to detect login providers

bamarni:
This introduces a login package and implements the auth header challenge as a fallback solution when the `/acs/api/v1/auth/providers` endpoint is not available.

This allows eg. to list providers with DC/OS OSS :
```
$ dcos auth list-providers http://54.202.72.66
  PROVIDER ID                     LOGIN METHOD                      
  dcos-oidc    Log in using OpenID Connect (Google, GitHub, or Microsoft)
```

With DC/OS EE 1.7/1.8 : 
```
$ dcos auth list-providers http://54.244.69.190
  PROVIDER ID                             LOGIN METHOD                             
  dcos-users   Log in using a standard DC/OS user account (username and password)
```

### DIFF
--- a/pkg/cmd/auth/auth_listproviders.go
+++ b/pkg/cmd/auth/auth_listproviders.go
@@ -42,7 +42,7 @@ func newCmdAuthListProviders(ctx *cli.Context) *cobra.Command {
 				}
 				fmt.Fprintln(ctx.Out(), string(out))
 			} else {
-				table := cli.NewTable(ctx.Out(), []string{"PROVIDER ID", "AUTHENTICATION TYPE"})
+				table := cli.NewTable(ctx.Out(), []string{"PROVIDER ID", "LOGIN METHOD"})
 				for name, provider := range providers {
 					table.Append([]string{name, provider.String()})
 				}
--- a/pkg/cmd/auth/auth_listproviders.go
+++ b/pkg/cmd/auth/auth_listproviders.go
@@ -6,18 +6,8 @@ import (
 
 	"github.com/dcos/dcos-cli/pkg/cli"
 	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/dcos/dcos-cli/pkg/login"
 	"github.com/spf13/cobra"
-)
-
-// These are the different auth types that DC/OS supports with the names that they'll be given from the providers
-// endpoint.
-const (
-	LoginTypeDCOSUidPassword     = "dcos-uid-password"
-	LoginTypeDCOSUidServiceKey   = "dcos-uid-servicekey"
-	LoginTypeDCOSUidPasswordLDAP = "dcos-uid-password-ldap"
-	LoginTypeSAMLSpInitiated     = "saml-sp-initiated"
-	LoginTypeOIDCAuthCodeFlow    = "oidc-authorization-code-flow"
-	LoginTypeOIDCImplicitFlow    = "oidc-implicit-flow"
 )
 
 // newCmdAuthListProviders creates the `dcos auth list-providers` subcommand.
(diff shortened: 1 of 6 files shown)


pkg/cmd/auth/auth_listproviders.go
```diff
@@ -52,13 +43,8 @@ func newCmdAuthListProviders(ctx *cli.Context) *cobra.Command {
 				fmt.Fprintln(ctx.Out(), string(out))
 			} else {
 				table := cli.NewTable(ctx.Out(), []string{"PROVIDER ID", "AUTHENTICATION TYPE"})
```
jgehrcke: Can we change `AUTHENTICATION TYPE` to `LOGIN METHOD`?

pkg/httpclient/client.go
```diff
@@ -31,6 +31,13 @@ func TLS(tlsConfig *tls.Config) Option {
 	}
 }
 
+// ACSToken sets the authorization token for HTTP requests.
```
jgehrcke: Can we rename this to "authentication token?"

pkg/login/login.go
```diff
@@ -0,0 +1,80 @@
...
+
+	c.logger.Info("Falling back to the WWW-Authenticate challenge.")
+	authHeader, err := c.challengeAuth()
+	if err != nil {
+		return nil, err
+	}
+
+	switch authHeader {
+	case "oauthjwt":
+		providers["dcos-oidc"] = defaultOIDCImplicitFlowProvider()
```
bamarni: @jgehrcke : Ideally this will match the future content of the login providers endpoint in DC/OS OSS, especially for the provider ID. It currently uses : 
``` json 
{
        "dcos-oidc": {
                "authentication-type": "oidc-implicit-flow",
                "client-method": "browser-prompt-authtoken",
                "config": {
                        "start_flow_url": "/login?redirect_uri=urn:ietf:wg:oauth:2.0:oob"
                },
                "description": "Google, GitHub, or Microsoft"
        }
}
```

Let me know if I should adjust here.

pkg/login/login.go
```diff
@@ -0,0 +1,83 @@
...
+	resp, err := c.http.Get("/acs/api/v1/auth/providers")
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	providers := make(map[string]*Provider)
+	if resp.StatusCode == 200 {
+		err := json.NewDecoder(resp.Body).Decode(&providers)
+		if err == nil {
```
ArmandGrillet: What is `err != nil`? We are already in the case of a status code 200 thus we should not expect one at this point.

pkg/login/login.go
```diff
@@ -0,0 +1,83 @@
...
+	case "acsjwt":
+		// DC/OS EE 1.7/1.8
+		providers["dcos-users"] = defaultDCOSUIDPasswordProvider()
+	default:
+		return nil, fmt.Errorf("unsupported WWW-Authenticate challenge '%s'", authHeader)
+	}
+	return providers, nil
+}
+
+// challengeAuth sends an unauthenticated HTTP request to a DC/OS well-known resource.
```
ArmandGrillet: Why `challenge` in the function name?

pkg/login/provider.go
```diff
@@ -0,0 +1,68 @@
...
+
+// String converts a login provider to a string.
+func (provider *Provider) String() string {
+	switch provider.Type {
+	case DCOSUIDPassword:
+		return "Log in using a standard DC/OS user account (username and password)"
+	case DCOSUIDServiceKey:
+		return "Log in using a DC/OS service user account (username and private key)"
+	case DCOSUIDPasswordLDAP:
+		return "Log in in using an LDAP user account (username and password)"
```
ArmandGrillet: One `in` should be removed.

pkg/login/login.go
bamarni: From https://tools.ietf.org/html/rfc2617 : 
> HTTP provides a simple challenge-response authentication mechanism that MAY be used by a server to challenge a client request and by a client to provide authentication information.

The current one in Python also has the `challenge` semantic : `header_challenge_auth`.

pkg/login/login.go
bamarni: This is indeed a code path which shouldn't happen, unless something is broken in DC/OS, or the API endpoint introduced some breaking changes. However in the unlikely case where it happens I'd rather fail with an explicit error rather than silencing it and having the CLI failing/misbehaving later on.

ArmandGrillet: > However in the unlikely case where it happens I'd rather fail with an explicit error rather than silencing it and having the CLI failing/misbehaving later on.
But that is the case at the moment, no? I would expect a:
```
		if err == nil {
			return providers, nil
		} else {
			return nil, err
		}
```

bamarni: Ah yes indeed, I was confused, updated.

pkg/login/login.go
bamarni: Follow-up from JP : 
> `dcos-oidc-auth0`
that identity provider is specifically Auth0
serves as a "proxy" to other identity providers

PR #1225, opened 2018-04-25
Revert fix for golint issue.

ArmandGrillet:


Dockerfile
```diff
@@ -2,9 +2,4 @@ FROM golang:1.10.0
 
 RUN go get -u \
     github.com/golang/dep/cmd/dep
```
bamarni: misses a trailing backslash

PR #1226, opened 2018-04-26
[WIP] Add a render package

bamarni:
It for now only contains a table helper.

PR #1227, opened 2018-04-29
Add a prompt package

bamarni:
This package provides helpers to ask for interactive input. It currently supports text input, password and choosing from a list.

Usage : 
``` go
prompt := prompt.New(os.Stdin, os.Stdout)

username := prompt.Input("Username: ")

pass := prompt.Password("Password:")

choices := []string{"a", "b", "c"}
i, err := prompt.Select("Please choose:", choices)
if err != nil {
	return err
}
choice = choices[i]
```

### DIFF
--- a/pkg/prompt/prompt.go
+++ b/pkg/prompt/prompt.go
@@ -1,0 +1,80 @@
+package prompt
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"reflect"
+	"strconv"
+
+	"golang.org/x/crypto/ssh/terminal"
+)
+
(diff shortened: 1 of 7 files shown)


pkg/prompt/prompt.go
```diff
@@ -0,0 +1,80 @@
+package prompt
```
ArmandGrillet: I find it strange to have a package for `prompt` but not for `table`.

bamarni: cf. #1226 , if we make our own package for table / json it should probably be outside of `cli`

PR #1228, opened 2018-05-02
Add a design page for CLI login

bamarni:
This document defines the goals and the high-level implementation of the CLI login flow.

### DIFF
--- a/design/login.md
+++ b/design/login.md
@@ -75,8 +75,8 @@ A login happens through the `dcos auth login` command, which accepts the followi
 - **--password-file** : Specify the path to a file that contains the password.
 - **--private-key** : Specify the path to the private key for service account login.
 
-The password can also be read from the `DCOS_PASSWORD` environment variable. The `--password` or
-`--password-file` flags take precedence over the environment variable.
+The username and password can also be read from the `DCOS_USERNAME` and `DCOS_PASSWORD` environment
+variables. Flags (`--username`, `--password`, `--password-file`) take precedence over environment variables.
 
 (A login also happens at the end of the `dcos cluster setup` command, which accepts all the flags
 from the `dcos auth login` command.)
--- a/design/login.md
+++ b/design/login.md
@@ -1,0 +1,102 @@
+# login
+
+The login package handles DC/OS login related operations.
+
+## Goals
+
+The goals of the login package are to :
+
+- Detect available login providers for a DC/OS cluster.
+- Log in to a DC/OS cluster using a given login provider.
+
+## Login providers detection
+
+To detect available login providers for a given DC/OS cluster, the CLI issues a GET request to the
+`/acs/api/v1/auth/providers` endpoint. This endpoint returns a list of providers in JSON. Each
+top-level key refers to the login provider ID, while its value is the login provider definition:
+
+``` json
+{
+  "dcos-users": {
+    "authentication-type": "dcos-uid-password",
+    "client-method": "dcos-usercredential-post-receive-authtoken",
+    "config": {
+      "start_flow_url": "/acs/api/v1/auth/login"
+    },
+    "description": "Default DC/OS authenticator"
+  }
+}
+```
+
+The above definition is a default DC/OS login provider, its ID is **dcos-users**.
+The ID has no special semantic apart from uniquely identifying a login provider.
+
+When the `/acs/api/v1/auth/providers` endpoint is not present, the available login provider is
+detected by making an unauthenticated `HEAD` request to a well-known protected DC/OS resource and
+reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
+
+- `acsjwt` : the cluster only supports the `dcos-users` provider as defined above.
+- `oauthjwt` : the cluster only supports the OpenID implicit flow provider, as defined below:
+
+``` json
+{
+  "dcos-oidc-auth0": {
+    "authentication-type": "oidc-implicit-flow",
+    "client-method": "browser-prompt-authtoken",
+    "config": {
+      "start_flow_url": "/login?redirect_uri=urn:ietf:wg:oauth:2.0:oob"
+    },
+    "description": "Google, GitHub, or Microsoft"
+  }
+}
+```
+
+## Log in to a DC/OS cluster
+
+Log in to a DC/OS cluster is the action of presenting some credentials to the cluster in order to
+get an authorization token in return (also known as `ACS token`).
+
+The DC/OS CLI supports 6 login providers types (`authentication-type`):
+
+- **dcos-uid-password** : Log in using a standard DC/OS user account (username and password).
+- **dcos-uid-password-ldap** : Log in using an LDAP user account (username and password).
+- **dcos-uid-servicekey** : Log in using a DC/OS service user account (username and private key).
+- **saml-sp-initiated** : Log in using SAML 2.0.
+- **oidc-authorization-code-flow** : Log in using OpenID Connect authorization code flow.
+- **oidc-implicit-flow** : Log in using OpenID Connect implicit flow.
+
+A login happens through the `dcos auth login` command, which accepts the following flags:
+
+- **--provider** (string) : Specify the login provider ID to use.
+- **--username** (string) : Specify the username.
+- **--password** (string) : Specify password on the command line (insecure).
+- **--password-env** (string) : Specify an environment variable name that contains the password.
+- **--private-key** (string) : Specify the path to a file that contains the private key for service
+    account login.
+
+(A login also happens at the end of the `dcos cluster setup` command, which accepts all the flags
+from the `dcos auth login` command.)
+
+In order to perform a login, the CLI must select a login provider to use. This selection can occur
+in 3 different ways:
+
+- **explicitly** : the user passed the `--provider` flag with a given login provider ID.
+- **implicitly** : there is a single available login provider, or the user passed some flags
+    which are specific to a single login provider.
+- **manually** : when the login provider is not explicit nor implicit, a list is prompted and
+    the user is expected to select a login provider manually.
+
+Once the login provider is selected, its relevant credentials are read from command-line flags
+and the user is being prompted for the missing ones (if any). The login provider's `client-method`
+is then triggered.
+
+Each login provider has one of these 4 login methods associated to it (`client-method`):
+
+- **dcos-usercredential-post-receive-authtoken**, **dcos-credential-post-receive-authtoken** :
+    POST username and password to the `start_flow_url` endpoint.
+- **dcos-servicecredential-post-receive-authtoken** : Generate a login token from a service account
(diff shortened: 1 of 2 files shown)


design/login.md
```diff
@@ -0,0 +1,102 @@
+# login
+
+The login package handles DC/OS login related operations.
+
+## Goals
+
+The goals of the login package are to :
```
ArmandGrillet: s/`to :`/`to:`

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+    },
+    "description": "Default DC/OS authenticator"
+  }
+}
+```
+
+The above definition is a default DC/OS login provider, its ID is **dcos-users**.
+The ID has no special semantic apart from uniquely identifying a login provider.
+
+When the `/acs/api/v1/auth/providers` endpoint is not present, the available login provider is
```
ArmandGrillet: Could you add for which DC/OS versions will this endpoint not be present?

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+- **dcos-uid-servicekey** : Log in using a DC/OS service user account (username and private key).
+- **saml-sp-initiated** : Log in using SAML 2.0.
+- **oidc-authorization-code-flow** : Log in using OpenID Connect authorization code flow.
+- **oidc-implicit-flow** : Log in using OpenID Connect implicit flow.
+
+A login happens through the `dcos auth login` command, which accepts the following flags:
+
+- **--provider** (string) : Specify the login provider ID to use.
+- **--username** (string) : Specify the username.
+- **--password** (string) : Specify password on the command line (insecure).
```
ArmandGrillet: `on the command line` seems unnecessary.

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+- **--password-env** (string) : Specify an environment variable name that contains the password.
+- **--private-key** (string) : Specify the path to a file that contains the private key for service
+    account login.
+
+(A login also happens at the end of the `dcos cluster setup` command, which accepts all the flags
+from the `dcos auth login` command.)
+
+In order to perform a login, the CLI must select a login provider to use. This selection can occur
```
ArmandGrillet: s/`must select a login provider to use`/`lets the user select a login provider`

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+in 3 different ways:
+
+- **explicitly** : the user passed the `--provider` flag with a given login provider ID.
+- **implicitly** : there is a single available login provider, or the user passed some flags
```
ArmandGrillet: s/`there is a single available login provider, or`/`a single login provider is available or`

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+    which are specific to a single login provider.
+- **manually** : when the login provider is not explicit nor implicit, a list is prompted and
```
ArmandGrillet: s/`a list is prompted and the user is expected to select`/`a list is prompted to let the user select`

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+    the user is expected to select a login provider manually.
+
+Once the login provider is selected, its relevant credentials are read from command-line flags
+and the user is being prompted for the missing ones (if any). The login provider's `client-method`
+is then triggered.
+
+Each login provider has one of these 4 login methods associated to it (`client-method`):
```
ArmandGrillet: I only see 3 login methods below.

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+
+- **dcos-usercredential-post-receive-authtoken**, **dcos-credential-post-receive-authtoken** :
+    POST username and password to the `start_flow_url` endpoint.
+- **dcos-servicecredential-post-receive-authtoken** : Generate a token from a service account
+    private key. POST username and the generated token to the `start_flow_url` endpoint.
+- **browser-prompt-authtoken** : Open the DC/OS cluster UI in a browser, at the page referenced
+    in `start_flow_url`. The user is then expected to continue the flow in the browser, eventually
```
ArmandGrillet: s/`The user is then expected to continue the flow in the browser, eventually they are redirected to a page with`/`We expect the user to follow the flow in the browser until being redirected to a page with`.

design/login.md
```diff
@@ -0,0 +1,102 @@
...
+- **manually** : when the login provider is not explicit nor implicit, a list is prompted and
+    the user is expected to select a login provider manually.
+
+Once the login provider is selected, its relevant credentials are read from command-line flags
+and the user is being prompted for the missing ones (if any). The login provider's `client-method`
+is then triggered.
+
+Each login provider has one of these 4 login methods associated to it (`client-method`):
+
+- **dcos-usercredential-post-receive-authtoken**, **dcos-credential-post-receive-authtoken** :
```
bamarni: @jgehrcke : one method is used in the `dcos-uid-password-ldap` provider definition while the other one is in the `dcos-uid-password` definition, apart from that is there any behavioral difference between the 2?

design/login.md
bamarni: 2 are on the same line, cf. my comment below

design/login.md
bamarni: The user might have as well nothing to select, if the scenario falls under the implicit scheme described below.

design/login.md
bamarni: Who's "we"? I've only used `the DC/OS CLI` in the document.

design/login.md
bamarni: I've left it out on purpose as this is not relevant to the design of this feature. If we were to say for eg. `this is not available in DC/OS open source`, that document will become inaccurate at some point.

design/login.md
bamarni: `Specify password (insecure).` sounds unclear, in my mind here `on the command line` emphasizes on why it is `insecure`.

br-lewis: I quickly looked around for examples from other command line utilities but it seems most will prompt for a password. I'd say wording it like `Specify password in plaintext (insecure)` might be better. Or maybe `Specify password inline in plaintext (insecure)` to be a little more explicit.

bamarni: Updated, I've also added back the `--password-file` option which I was planning to deprecate as suggested here : https://docs.google.com/document/d/17rR1BNEE9Ws27Faa5poPnpitqhT1ks843tCpSUD1Tu8/edit?disco=AAAAB3jFhV8

But the same argument could be said for the --password option, it most likely persists the password to the disk (in the shell history), so marking them as insecure sounds reasonable and avoids breaking users scripts in the future.

design/login.md
ArmandGrillet: The issue is with `the CLI must select`, the interface doesn't select a thing but the user does, no?

design/login.md
ArmandGrillet: Then `The user is then expected to follow the flow in the browser until being redirected to a page with`.

design/login.md
```diff
@@ -0,0 +1,103 @@
...
+- **saml-sp-initiated** : Log in using SAML 2.0.
+- **oidc-authorization-code-flow** : Log in using OpenID Connect authorization code flow.
+- **oidc-implicit-flow** : Log in using OpenID Connect implicit flow.
+
+A login happens through the `dcos auth login` command, which accepts the following flags:
+
+- **--provider** (string) : Specify the login provider ID to use.
+- **--username** (string) : Specify the username.
+- **--password** (string) : Specify password in plaintext (insecure).
+- **--password-file** (string) : Specify the path to a file that contains the password (insecure).
```
mhrabovcin: Why is reading a password from a file is considered insecure?

bamarni: The insecure part is not about the reading, but about having a password stored in plaintext in a file.

jongiddy: Although files can be leaked through bad practices, I'd consider files with 0600 permissions to be adequately secure.  I'd be more worried that the support for storing passwords in environment variable will encourage users to add their password to their `.profile` file, where we cannot detect that the password is exposed.

Why is there no support for interactive "getpass"-like support for entering the password?

bamarni: There is, these options are for scripting purposes. If they are not passed the user is prompted for a password.

design/login.md
```diff
@@ -0,0 +1,103 @@
...
+from the `dcos auth login` command.)
+
+In order to perform a login, the CLI must select a login provider to use. This selection can occur
+in 3 different ways:
+
+- **explicitly** : the user passed the `--provider` flag with a given login provider ID.
+- **implicitly** : a single login provider is available or the user passed some flags
+    which are specific to a single login provider.
+- **manually** : when the login provider is not explicit nor implicit, a list is prompted to let
+    the user select a login provider manually.
```
jongiddy: Is there a flag to prevent interactivity (e.g. the manual option here) for scripts?

design/login.md
```diff
@@ -0,0 +1,103 @@
...
+
+``` json
+{
+  "dcos-users": {
+    "authentication-type": "dcos-uid-password",
+    "client-method": "dcos-usercredential-post-receive-authtoken",
+    "config": {
+      "start_flow_url": "/acs/api/v1/auth/login"
+    },
+    "description": "Default DC/OS authenticator"
```
jgehrcke: Even if you got that from a real DC/OS cluster let's replace "authenticator" with "login provider".

design/login.md
bamarni: Do you mean a generic flag or one for this specific case? What I would do here in scripts is passing the --provider flag explicitly.

jongiddy: A command stopping for user interaction can sometimes be hard to debug, especially when it occurs in a batch process on a remote node (e.g. CI). It is nicer to have an option to ensure that if the script gets to a point where it needs interaction, it will fail with a non-zero exit code and an informative message instead.

So I think it would be generic. If the user forgets to set where the password is coming from, or what provider to use, then the script fails rather than hanging waiting for a non-existent interactive user.

However, this is particularly important for a case like this where a script may work for me (because I only have access to a single login provider) but fail for you (because you have access to multiple login providers).

bamarni:
> I forgot to bring that up in the meeting. We should probably replace all use of "username" with uid.

@jgehrcke : are you referring to the wording in our help messages / docs only, or even the `--username` flag?

design/login.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Log in to a DC/OS cluster
+
+Log in to a DC/OS cluster is the action of presenting some credentials to the cluster in order to
+get an authorization token in return (also known as `ACS token`).
+
+The DC/OS CLI supports 6 login providers types (`authentication-type`):
+
+- **dcos-uid-password** : Log in using a standard DC/OS user account (username and password).
+- **dcos-uid-password-ldap** : Log in using an LDAP user account (username and password).
```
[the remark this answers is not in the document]
bamarni: are you referring to the wording in our help messages / docs only, or even the --username flag?

PR #1232, opened 2018-05-03
Job Kill Requires Argument

kensipe:
Either job_run_id OR --all flag is required for `dcos job <job_id>`

https://jira.mesosphere.com/browse/DCOS_OSS-3302

### DIFF
--- a/python/lib/dcoscli/dcoscli/data/help/job.txt
+++ b/python/lib/dcoscli/dcoscli/data/help/job.txt
@@ -10,7 +10,7 @@ Description:
         dcos job remove <job-id> [--stop-current-job-runs]
         dcos job show <job-id>
         dcos job update <job-file>
-        dcos job kill <job-id> [<run-id>][--all]
+        dcos job kill <job-id> (<run-id>|--all)
         dcos job run <job-id> [--json]
         dcos job list [--json|--quiet]
         dcos job schedule add <job-id> <schedule-file>


python/lib/dcoscli/dcoscli/data/help/job.txt
```diff
@@ -10,7 +10,7 @@ Description:
         dcos job remove <job-id> [--stop-current-job-runs]
         dcos job show <job-id>
         dcos job update <job-file>
-        dcos job kill <job-id> [<run-id>][--all]
+        dcos job kill <job-id> (<run-id>|--all)
```
alenkacz: should we maybe also improve the message when someone calls it with just jobid? So that it does not spit out this 'NoneType' thing but says something meaningful?

alenkacz:
and the docs are also kind of misleading, because it looks like the thing is optional :/ https://docs.mesosphere.com/1.11/cli/command-reference/dcos-job/dcos-job-kill/

python/lib/dcoscli/dcoscli/data/help/job.txt
bamarni: By default it would print an `Invalid subcommand usage` error message with the help menu below. If you want a specific message this should be a check at the beginning of the command instead.

alenkacz: oh ok, so this file is used also for validation, cool ... then it should be enough, thanks!

bamarni:
run linux integration tests

kensipe:
thanks @bamarni !

PR #1236, opened 2018-05-09
WIP: Dispatch refactor

br-lewis:
This PR is currently a proposal for how to do command dispatch in a way that allows for autocompletion and for the inclusion and integration of externally available commands in the command tree.

Major points:
* This adds a `DcosCommand` interface which is implemented by `InternalCommand` and `ExternalCommand`. `InternalCommand`s hold a cobra `Command` and that `Command`'s children along with a function for customized tab completion. At runtime, it's determined if this call is trying to do tab completion or a normal run of the command and then the root `InternalCommand` will build either a normal command tree or an autocomplete command tree, both consist of cobra `Command`s but will have either the normal running code or an autocomplete function.
* `InternalCommand` holds the normal cobra `Command` in order to simplify command creation (not needing to replicate cobra's API) and allow autocomplete to do some introspection on it to pull out args and other things for autocompletion.
* `InternalCommand::autocomplete` is a function and not a command because we want default behavior that we don't need to replicate to every level of every command so `InternalCommand::AutocompleteCommand()` builds the cobra `Command` there and has a call to `autocomplete`. It also passes in `InternalCommand::command` as the first parameter instead of what cobra would normally do which is pass in the command being run (i.e. the autocomplete rather than the normal command).
* Under this method, you can't add a command's children to the cobra `Command` struct. This is because we won't have an easy way to associate the autocomplete functions with the correct `DcosCommand`. Instead you add them to the `DcosCommand` via the `AddSubCommand` method. If they're added to the `Command` struct the children will be duplicated.

I haven't added most of the current commands to the new tree, I figure it would be better to get some feedback before doing that. An example of Autocomplete can be found in `pkg/cmd/cluster/cluster_attach.go`.

Other things:
* `ExternalCommand` currently isn't implemented beyond some stub functions. But I don't think we'll need to do any major changes to the API here for that though.
* `InternalCommand::AutocompleteCommand` doesn't do flag completion yet.
* I don't think `pkg/subcommand` is the best place for the new command structs but I'm not sure of what a better place would be, any ideas?

=== Edit ===
I accidentally included a command for generating a bash completion script which also isn't finished nor is necessary for this PR. I think we're planning to create something like this anyway but I can pull these changes out of this PR if needed.


br-lewis:
From discussions that have taken place offline, we've decided that we won't be going with this method. It adds too much complexity to the code.

PR #1237, opened 2018-05-09
Add tests for commands

bamarni:
- Add a mockcluster package
- Create a fake environment
- Add a test for `dcos auth list-providers`


pkg/cmd/cluster/cluster_list.go
```diff
@@ -20,62 +20,82 @@ func newCmdClusterList(ctx api.Context) *cobra.Command {
...
+						item.Status = "AVAILABLE"
+						item.Version = version.Version
+					}
 					mu.Lock()
-					infos = append(infos, info)
-					mu.Unlock()
+					defer mu.Unlock()
+
+					// Order clusters by name, this guarantees a stable list in table output.
+					itemsLen := len(items)
```
ArmandGrillet: I find this loop slightly more comprehensible:
```
// Order clusters by name, this guarantees a stable list in table output.
for i := range items {
	if items[i].Name > item.Name {
		items = append(items[:i], append([]ClusterInfo{item}, items[i:]...)...)
		return
	}
}
```
I ran the tests with that loop and it passed.

bamarni: definitely :+1:  just updated it

PR #1238, opened 2018-05-10
Create CLI Style Guide

leemunroe:
This style guide is to help align internal and external CLI plugins going forward so the experience is consistent for DC/OS users.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,0 +1,212 @@
+# DC/OS CLI Style Guide
+
+## Intro
+The DC/OS CLI is the main tool used by developers and operators to interact with DC/OS cluster and services running on the clusters. The teams developing features and CLI plugins should build the CLI per this style guide so that we offer a consistent and usable experience to users of DC/OS.
+
+## Naming Commands
+* Generally top level commands are single nouns e.g. dcos job, dcos service
+* Top level commands are followed by verbs e.g. dcos job create, dcos cluster rename
+* Command names should always be a single, lowercase word without spaces, hyphens, underscores, or other word delimiters
+
+## Inputs
+
+### Actions
+
+* If no action provided then default to showing usage help output
+* Try to be consistent with other subcommands; only break convention if necessary
+  * Create - Creating a new object aka New, Add
+  * Delete - Deleting an object from the system aka Destroy
+  * Add - Adding an object that already exists but not creating a new object e.g. adding an existing user to a group
+  * Remove - Removing from an object but not deleting e.g. removing a user from a group
+  * Show - Show a description or definition aka Describe, Get
+  * List - List all objects
+
+### Flags
+* `--help` and `-h` should always output one level of help
+* `--version` should always show the version of the plugin
+* `--json` outputs JSON
+* `--force` and `-f` to force
+* `--quiet` and `-q` to silence output
+* `-v` and `-vv` for verbose outputs (for verbose and very verbose)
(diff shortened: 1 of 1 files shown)


STYLEGUIDE.md
```diff
@@ -0,0 +1,212 @@
+# DC/OS CLI Style Guide
+
+## Intro
+The DC/OS CLI is the main tool used by developers and operators to interact with DC/OS cluster and services running on the clusters. The teams developing features and CLI plugins should build the CLI per this style guide so that we offer a consistent and usable experience to users of DC/OS.
+
+## Naming Commands
+* Generally top level commands are single nouns e.g. dcos job, dcos service
+* Top level commands are followed by verbs e.g. dcos job create, dcos cluster rename
+* Command names should always be a single, lowercase word without spaces, hyphens, underscores, or other word delimiters
```
bamarni: We currently have `dcos auth list-providers`, in cases like this `dcos auth providers list` would be overkill as this is the only command and it is not meant to grow. Can we say that commands might contains a hyphen delimiter?

STYLEGUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+* `-v` and `-vv` for verbose outputs (for verbose and very verbose)
+* `--lines=N` to restrict amount of results
+
+## Arguments
+
+### Positional Arguments
+
+* A positional argument is a bare value, and its position in a list of arguments identifies it
+
+### Named Arguments
```
bamarni: I'm not familiar with this term, do you mean "Options" or "Flags"?

STYLEGUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+* A named argument is a (key, value) pair, where the key identifies the value
+* Flags may or may not accept named arguments
+* Flags should have smart defaults when no named argument is provided
+* Named arguments should be preceded by “=” i.e. not a space
```
bamarni: This is very common to accept `=` or a space interchangeably in CLIs, we currently support this as this is what libraries do by default.

STYLEGUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+### ✅ Do
+
+``` 
+$ dcos <command> --<flag>=<argument>
+```
+
+### 🚫 Don't
+
+```
+$ dcos <command> --<flag> <argument>
```
bamarni: As commented above, I think both are fine and should be accepted.

STYLEGUIDE.md
leemunroe: Hmm I pulled this from [here](https://pythonconquerstheuniverse.wordpress.com/2010/07/25/command-line-syntax-some-basic-concepts/) when looking to define the difference between arguments, options, flags. What would you suggest? You have more experience with these terms than me.

STYLEGUIDE.md
leemunroe: Interesting. In that case should we explicitly state that both should be supported?

STYLEGUIDE.md
bamarni: sounds good

STYLEGUIDE.md
bamarni: Arguments / flags is enough for the wording. Then a flag may or may not accept a value.

STYLEGUIDE.md
bamarni: yes makes sense

leemunroe:
@bamarni Made a few changes to address your concerns.

PR #1239, opened 2018-05-12
Add dcos cluster setup

bamarni:
- Refactor httpclient options
- [prompt] Add yes/no confirmation
- [login] Fix missing parameter in a log message
- Move back cli.Cluster to config.Cluster
- Support the /metadata endpoint in DCOSClient
- Add a Mesos client
- Add `dcos cluster setup`


pkg/cli/setup.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// Read cluster name from Mesos state summary.
+	if stateSummary, err := mesos.NewClient(httpClient).StateSummary(); err == nil {
+		cluster.SetName(stateSummary.Cluster)
+	} else {
+		// Fallback to cluster ID as cluster name.
+		cluster.SetName(metadata.ClusterID)
+	}
+
+	// Create the config for the given cluster and set it as the current.
```
ArmandGrillet: s/`set it as the current.`/`attach the cluster`.

pkg/config/manager.go
```diff
@@ -142,6 +142,24 @@ func (m *Manager) All() (configs []*Config) {
 	return
 }
 
+// Create saves a config to the disk under the given cluster ID folder.
+func (m *Manager) Create(config *Config, id string, caBundle []byte) error {
```
ArmandGrillet: `Save` would be more descriptive as a function name.

pkg/cli/setup.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+}
+
+// Register registers cluster setup flags.
+func (f *SetupFlags) Register(flags *pflag.FlagSet) {
+	flags.BoolVar(
+		&f.insecure,
+		"insecure",
+		false,
+		"Allow requests to bypass SSL certificate verification (insecure).",
```
mhrabovcin: `... TLS certificate verification ...`

pkg/cli/setup.go
```diff
@@ -0,0 +1,297 @@
...
+		&f.noCheck,
+		"no-check",
+		false,
+		"Do not check CA certficate downloaded from cluster (insecure). Applies to Enterprise DC/OS only.",
+	)
+	flags.StringVar(
+		&f.caBundlePath,
+		"ca-certs",
+		"",
+		"Specify the path to a list of trusted CAs to verify requests against.",
```
mhrabovcin: What does it mean a `path to a list`. Is it a directory which all files are accepted. Is it a path to a single file?

pkg/cli/setup.go
```diff
@@ -0,0 +1,297 @@
...
+	// Create the config for the given cluster and attach the cluster.
+	err = s.configManager.Save(cluster.Config(), metadata.ClusterID, flags.caBundle)
+	if err != nil {
+		return err
+	}
+	return s.configManager.Attach(cluster.Config())
+}
+
+// needsDCOSCABundle checks whether or not the cluster certificate is already trusted
+// by the sytem. This is done by making an HTTPS request to the cluster. This check
```
mhrabovcin: There is actually no guarantee that `needsDCOSCABundle` makes a TLS request. If the `httpClient` isn't configured to try secure URL it will correctly assume that it is not necessary to download DC/OS CA Bundle.

pkg/httpclient/client.go
```diff
@@ -14,58 +14,85 @@ import (
...
 	baseURL    string
-	timeout    time.Duration
 	baseClient *http.Client
-	logger     *logrus.Logger
+	ctx        context.Context
 }
 
 // Option is a functional option for an HTTP client.
-type Option func(*Client)
+type Option func(context.Context) context.Context
```
mhrabovcin: Why are you making this change? Why is it necessary for `context` to keep values that are later used to modify HTTP requests behavior (i.e. ACS token).

pkg/cli/setup.go
bamarni: Good point, it is a path to a single file, which might contain a single or multiple certificate authorities.

pkg/cli/setup.go
bamarni: Indeed, I'll adapt the comment. Also for optimization purpose it might make sense to check the URL scheme and return early with false if it's `http`, WDYT?

mhrabovcin: Will it work though if you're trying to setup a cluster in `disabled` security mode?

pkg/httpclient/client.go
mhrabovcin: Yes, my main point was about using the context object for options sharing. My perception of context is that it should enhance information going through request, but not make request behaviour dependent on the context.

bamarni: I can agree with that, introducing a dedicated `Options` struct should be cleaner.

pkg/cli/setup.go
bamarni: Changing to `Specify the path to a file with trusted CAs to verify requests against.`

pkg/cli/setup.go
bamarni: What might be problematic there?

bamarni:
@mhrabovcin : thanks for the review, I've addressed your comments and made some further improvements.

PR #1244, opened 2018-05-24
First draft of tab completion design doc

br-lewis:


design/tab_complete.md
```diff
@@ -0,0 +1,63 @@
...
+is somewhat nebulous at the moment. There are many aspects to it and the implementation will be
+dependent on what's most feasible with the command dispatch framework that we use.
+
+## Goals ##
+There are multiple different aspects of tab completion that may or may not make it into the CLI
+depending on how feasible they are to implement.
+
+*Must have:*
+* Support for bash.
+* Subcommand completion, e.g. `dcos cluster[TAB]` should give you `list attach remove rename`.
```
bamarni: missing space `dcos cluster [TAB]`

design/tab_complete.md
```diff
@@ -0,0 +1,63 @@
...
+at the internals of the commands we've built in order to extract subcommand and flag names. For additional
+pieces like argument completion or flag value completion, we're thinking of adding some metadata to the
+command/flag objects that the autocomplete command can read to determine what special behavior might be needed.
+
+The current problem that we're running into with Cobra comes from how it handles flags. Parsing flags is
+complicated and much of the code to handle it is in private functions within the Cobra package that also
+use fields private to Cobra. This makes it extremely difficult to differentiate a flag followed by a
+subcommand name and a flag that takes a value, e.g. with `dcos -v cluster list` we would need to check
+that `-v` doesn't take a value, therefore `cluster` must be an argument/subcommand. Because of this, it will
+be difficult to do robust argument completion or flag value completion.
```
bamarni: How about not using introspection in order to detect which flags / commands are available in the Go binary? But rather, define this in the code with the structure which makes the most sense for this feature. This would require to keep this structure in sync with the actual command / flags hierarchy, but it's a relatively straightforward task and would avoid the complications you're outlining.

design/tab_complete.md
br-lewis: The space wasn't left out particularly intentionally but also I don't think it needs to be there because it's perfectly valid to hit tab at that point and expect the same behavior.

bamarni: Not exactly, even though there is a single match, at this point the word to autocomplete is still `cluster`, one would have to press TAB twice in order to get the behavior you're describing.

design/tab_complete.md
br-lewis: This is an interesting thought and now I'm thinking that it makes more sense to abandon cobra and write our own command dispatch.

First, if we're going to maintain our own metadata about the commands and their flags, then really we would end up with a structure that's essentially the same as the completion structure we'll use for external commands so we should just always use the metadata if we go that route.

But this led to me thinking about how we'd implement that and by extension, how we'd deal with external commands. I think in order to parse that structure against the input we've received so far, we'll need to more or less reimplement how cobra does command parsing anyway. The metadata would need to encode what commands there are, the structure of the command tree, what flags each command can take, and whether or not those flags take arguments. We'd also need to be able to parse the given command string to find the appropriate command to get completions for. From there all we need to do is add in a field for the function that the command should run and flag argument info (what type it should be, default values) and we've more or less remade cobra.

Even if we maintained a flattened list of the commands instead of a tree, we'd still need to deal with flags, including flags on parent commands, unless we get rid of those but I think the subcommands might have some.

bamarni: If we don't go with the introspection approach then indeed we'd have to re-implement a subset of what cobra does, but as you said the good part about it is that we would have to do this anyways for plugin subcommands metadata, so there shoud be no wasted work here.

Then whether we move away from cobra to implement our own command dispatcher based on this metadata structure is another topic. I'd need to hear more about that to have a clear opinion, but I feel like a generic declarative structure is not as powerful as writing code and we'd probably face some difficulties if we were to go that route.

PR #1247, opened 2018-06-05
Added cluster link and unlink subcommands.

ArmandGrillet:
Also added a helper function to allow names when choosing a cluster, `IsUniqueCluster(name string)`.

### DIFF
--- a/pkg/clusterlinker/client.go
+++ b/pkg/clusterlinker/client.go
@@ -16,11 +16,8 @@ type Client struct {
 	logger *logrus.Logger
 }
 
-type LoginProvider struct {
-	ID   string `json:"id"`
-	Type string `json:"type"`
-}
-
+// LinkRequest is the request sent to the /v1/links endpoint to link the cluster
+// handled by the client to a cluster using its id, name, url, and login provider.
 type LinkRequest struct {
 	ID            string `json:"id"`
 	Name          string `json:"name"`
@@ -28,7 +25,14 @@ type LinkRequest struct {
 	LoginProvider `json:"login_provider"`
 }
 
-// NewClient creates a new DC/OS client.
+// LoginProvider representing a part of the message when sending a link request,
+// it gives information about the cluster to link to.
+type LoginProvider struct {
+	ID   string `json:"id"`
+	Type string `json:"type"`
+}
+
+// NewClient creates a new cluster linker client from a standard HTTP client.
 func NewClient(baseClient *httpclient.Client, logger *logrus.Logger) *Client {
 	return &Client{
 		http:   baseClient,
@@ -36,6 +40,7 @@ func NewClient(baseClient *httpclient.Client, logger *logrus.Logger) *Client {
 	}
 }
 
+// Link sends a link request to /v1/links using a given client.
 func (c *Client) Link(linkRequest *LinkRequest) error {
 	message, err := json.Marshal(linkRequest)
 
@@ -57,6 +62,7 @@ func (c *Client) Link(linkRequest *LinkRequest) error {
 	return nil
 }
 
+// Unlink sends an unlink request to /v1/links using a given client.
 func (c *Client) Unlink(linkedClusterID string) error {
 	resp, err := c.http.Delete("/cluster/v1/links/" + linkedClusterID)
 	if err != nil {
--- a/pkg/clusterlinker/client.go
+++ b/pkg/clusterlinker/client.go
@@ -1,0 +1,77 @@
+package clusterlinker
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+
+	"github.com/dcos/dcos-cli/pkg/dcos"
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/sirupsen/logrus"
+)
+
+// Client is able to detect available login providers and login to DC/OS.
+type Client struct {
+	http   *httpclient.Client
+	logger *logrus.Logger
+}
+
+type LoginProvider struct {
+	ID   string `json:"id"`
+	Type string `json:"type"`
+}
+
+type LinkRequest struct {
+	ID            string `json:"id"`
+	Name          string `json:"name"`
+	URL           string `json:"url"`
+	LoginProvider `json:"login_provider"`
+}
+
+// NewClient creates a new DC/OS client.
+func NewClient(baseClient *httpclient.Client, logger *logrus.Logger) *Client {
+	return &Client{
+		http:   baseClient,
+		logger: logger,
+	}
+}
+
+func (c *Client) Link(linkRequest *LinkRequest) error {
+	message, err := json.Marshal(linkRequest)
+
+	resp, err := c.http.Post("/cluster/v1/links", "application/json", bytes.NewReader(message))
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		var apiError *dcos.Error
+		if err := json.NewDecoder(resp.Body).Decode(&apiError); err != nil {
+			return errors.New("couldn't link")
+		}
+		return apiError
+	}
+
+	return nil
+}
+
+func (c *Client) Unlink(linkedClusterID string) error {
+	resp, err := c.http.Delete("/cluster/v1/links/" + linkedClusterID)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		var apiError *dcos.Error
+		if err := json.NewDecoder(resp.Body).Decode(&apiError); err != nil {
+			return errors.New("couldn't unlink")
+		}
+		return apiError
+	}
+
+	return nil
+}
--- a/pkg/cmd/cluster/cluster_link.go
+++ b/pkg/cmd/cluster/cluster_link.go
@@ -23,27 +23,31 @@ func newCmdClusterLink(ctx api.Context) *cobra.Command {
 				return err
 			}
 
+			var linkableCluster *config.Cluster
 			manager := ctx.ConfigManager()
-			linkableClusterConfig, err := manager.Find(args[0], false)
+			linkableClusterConfig, err = manager.Find(args[0], false)
 			if err != nil {
-				if err == config.ErrConfigNotFound {
-					msg := " is not set up in the CLI, would you like to do it now?"
-					err = ctx.Prompt().Confirm(args[0] + msg)
-					if err != nil {
-						return err
-					}
-
-					ctx.Setup(setupFlags, args[0])
-				} else {
+				if err != config.ErrConfigNotFound {
 					return err
 				}
+
+				msg := " is not set up in the CLI, would you like to do it now?"
+				err = ctx.Prompt().Confirm(args[0] + msg)
+				if err != nil {
+					return err
+				}
+
+				linkableCluster, err = ctx.Setup(setupFlags, args[0])
+				if err != nil {
+					return err
+				}
+			} else {
+				linkableCluster = config.NewCluster(linkableClusterConfig)
 			}
 
-			if attachedCluster.Config().Path() == linkableClusterConfig.Path() {
+			if attachedCluster.Config().Path() == linkableCluster.Config().Path() {
 				return errors.New("cannot link a cluster to itself")
 			}
-
-			linkableCluster := config.NewCluster(linkableClusterConfig)
 
 			client := login.NewClient(ctx.HTTPClient(linkableCluster), ctx.Logger())
 			rawProviders, err := client.Providers()
--- a/pkg/cmd/cluster/cluster_link.go
+++ b/pkg/cmd/cluster/cluster_link.go
@@ -33,9 +33,7 @@ func newCmdClusterLink(ctx api.Context) *cobra.Command {
 						return err
 					}
 
-					// We do not want to attach the linkable cluster.
-					attach := false
-					ctx.Setup(setupFlags, args[0], attach)
+					ctx.Setup(setupFlags, args[0])
 				} else {
 					return err
 				}
--- a/pkg/cmd/cluster/cluster_link.go
+++ b/pkg/cmd/cluster/cluster_link.go
@@ -1,0 +1,89 @@
+package cluster
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-cli/pkg/config"
+	"github.com/dcos/dcos-cli/pkg/login"
+	"github.com/spf13/cobra"
+)
+
+// newCmdClusterLink links the attached cluster to another one.
+func newCmdClusterLink(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:  "link",
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			attachedCluster, err := ctx.Cluster()
+			if err != nil {
+				return err
+			}
+
+			var linkableCluster config.Cluster
+			for _, cluster := range ctx.Clusters() {
+				if args[0] == cluster.ID() ||
+					args[0] == cluster.URL() ||
+					args[0] == cluster.Name() && ctx.IsUniqueCluster(cluster.Name()) {
+					linkableCluster = *cluster
+					break
+				}
+			}
+
+			if (config.Cluster{}) == linkableCluster {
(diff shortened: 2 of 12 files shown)


pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,57 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			attachedCluster, err := ctx.Cluster()
+			if err != nil {
+				return err
+			}
+
+			var linkedCluster config.Cluster
+			for _, cluster := range ctx.Clusters() {
+				if args[0] == cluster.ID() ||
+					args[0] == cluster.URL() ||
```
bamarni: In the current CLI, `dcos cluster unlink` doesn't accept an URL as argument (only a cluster name or id), I think it should stay like that.

The cluster linker API accepts an ID, and it might not be a configured cluster on the client side, so in case there is no cluster locally for a given ID, it should still make the DELETE request.

pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,57 @@
...
+				return errors.New("cannot unlink a cluster to itself")
+			}
+
+			client := ctx.HTTPClient(attachedCluster)
+			resp, err := client.Delete("/cluster/v1/links/" + linkedCluster.ID())
+			if err != nil {
+				return err
+			}
+
+			fmt.Println(resp)
```
bamarni: I just noticed we didn't add it to the style guide yet, but we are following the `no news is good news` UNIX motto across the CLI. No output with a zero exit code is enough to indicate success.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,101 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			attachedCluster, err := ctx.Cluster()
+			if err != nil {
+				return err
+			}
+
+			var linkableCluster config.Cluster
+			for _, cluster := range ctx.Clusters() {
+				if args[0] == cluster.ID() ||
+					args[0] == cluster.URL() ||
```
bamarni: Currently `dcos cluster link` only supports passing an URL, it requires the cluster to already be configured in the CLI though. Originally the idea was that, in order to link cluster A to cluster B, assuming A is already configured and attached, instead of having to: 
1) setup B
2) attach back to A
3) link A to B

One could just do `dcos cluster link B` using its URL, and this would trigger a setup "on-the-fly" before it eventually links A to B at the end of the flow. We didn't implement it as this required a lot of refactoring on the current CLI (to allow a setup on the fly) and thought we'd tackle this during this refactoring.

I hope we can all agree that the current implementation has a very bad UX, configured clusters should only be referred to using their name or ID, not URL. Here I think `dcos cluster link <url>` should trigger a cluster setup, and `dcos cluster link <id_or_name>` should look-up an already configured cluster.

@vinodkone : could you please confirm this one? I remember we already discussed this.

pkg/cmd/cluster/cluster_unlink.go
ArmandGrillet: In that case, why do we accept a cluster name? The URL is a unique argument that can easily be used to find which cluster should be unlinked, as having it accepted by the command requiring only one line I don't see why it shouldn't be allowed.

pkg/cmd/cluster/cluster_link.go
ArmandGrillet: I've added the new workflow.

pkg/cmd/cluster/cluster_unlink.go
bamarni: URLs are not identifiers for machines or clusters, but resources. They also specify how to access these resources.

A DC/OS cluster has an unbounded number of resources under it. Even though we agreed on a given resource to "refer" to the cluster (eg. `/`), there would still be an unbounded number of URLs to identify the cluster (eg. using different schemes, specifying the port or not). If we were to go that route, I'd argue that only the host should be used to identify the cluster, but even then, different hosts can point to the same machine.

On the other hand, a DC/OS cluster has a single ID and a single name. They are unambiguous and canonical ways to identify a cluster. This is what we use to refer to them in `dcos cluster attach`, `dcos cluster remove` and `dcos cluster rename`. If we use the same code here (with the config manager), this would also remove more than a line of code.

pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+
+			var linkedCluster config.Cluster
+			for _, cluster := range ctx.Clusters() {
+				if args[0] == cluster.ID() ||
+					(args[0] == cluster.Name() && ctx.IsUniqueCluster(cluster.Name())) {
+					linkedCluster = *cluster
+					break
+				}
+			}
```
bamarni: There is already a `Find` method to lookup a cluster config based on its ID / name, it is used in other cluster subcommands as well. This would remove the need for `IsUniqueCluster` in the context.

pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,49 @@
...
+
+			if (config.Cluster{}) == linkedCluster {
+				return errors.New("unable to retrieve cluster " + args[0])
```
bamarni: When no cluster is found, it should still proceed and make the unlink request, the link might still exist even though the cluster is not configured locally.

pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+
+			if attachedCluster.ID() == linkedCluster.ID() {
+				return errors.New("cannot unlink a cluster to itself")
+			}
+
+			client := ctx.HTTPClient(attachedCluster)
+			_, err = client.Delete("/cluster/v1/links/" + linkedCluster.ID())
+			if err != nil {
```
bamarni: The operation might have failed even though there is no err here (eg. a 404 is returned when the link doesn't exist).

How about a `FailOnErrorStatus()` option for the httpclient (that would error-out when the response status is 4XX / 5XX)? That would be helpful in other places too.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,133 @@
...
+				LoginProvider: LoginProvider{
+					ID:   provider.ID,
+					Type: provider.Type,
+				},
+			}
+			message, err := json.Marshal(linkRequest)
+
+			attachedClient := ctx.HTTPClient(attachedCluster)
+			_, err = attachedClient.Post("/cluster/v1/links", "application/json", bytes.NewReader(message))
+			if err != nil {
```
bamarni: Here as well we'd need to check the response status code.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,133 @@
...
+					return err
+				}
+
+				// We save the current configuration before starting the setup.
+				manager := ctx.ConfigManager()
+				conf, err := manager.Current()
+				if err != nil {
+					return err
+				}
+				ctx.Setup(setupFlags, args[0])
```
bamarni: We should check for a possible error here. Also, with this new use-case, how about not attaching the cluster at the end of the setup, but let the caller do it? This would remove the need for saving the current config before the setup, then getting the newly configured cluster and eventually attaching back to the previous one.

So `ctx.Setup()` would return a `(*Config, err)` or `(*Cluster, err)`.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,133 @@
...
+			if attachedCluster.ID() == linkableCluster.ID() {
+				return errors.New("cannot link a cluster to itself")
+			}
+
+			client := login.NewClient(ctx.HTTPClient(&linkableCluster), ctx.Logger())
+			rawProviders, err := client.Providers()
+			if err != nil {
+				return err
+			}
+			providers := rawProviders.Slice()
```
bamarni: Not all login providers are supported for a cluster link. Currently we only support (cf. the Python code) : 
```
                auth.AUTH_TYPE_DCOS_UID_PASSWORD,
                auth.AUTH_TYPE_DCOS_UID_PASSWORD_LDAP,
                auth.AUTH_TYPE_SAML_SP_INITIATED,
                auth.AUTH_TYPE_OIDC_AUTHORIZATION_CODE_FLOW]
```

bamarni:
I think it would make sense to have a cluster linker client in its own package, or as part of the Client from the `dcos` package, as the API will be used in other places too. `dcos cluster attach` and `dcos cluster list` also depend on this API (linked clusters appear in the list and it is possible to attach to a linked cluster, this would trigger a setup using the login provider registered in the link).

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			manager := ctx.ConfigManager()
+			linkableClusterConfig, err := manager.Find(args[0], false)
+			if err != nil {
+				return err
+			}
+
+			if config.Empty() == linkableClusterConfig {
+				// The cluster does not exist yet. Two possibilities:
```
bamarni: If the cluster does not exist, an error would have already been returned at this point. The content of this block should be moved to line 31 instead. We could also add a specific error to make things more robust : 
```
	linkableClusterConfig, err := manager.Find(args[0], false)
	if err != nil {
		if err != config.ErrConfigNotFound {
			return err
		}
		// proceed with cluster setup
	}
```

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,129 @@
...
+				// - The argument is an URL, we try to setup the cluster.
+				// - The argument is not an URL, we return an error.
+				_, err = url.ParseRequestURI(args[0])
```
bamarni: Does this support passing an URL without scheme? Currently we accept such arguments and assume `https://`. We should also move this validation logic to the Setup function not to duplicate the logic with the setup command.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,129 @@
...
+
+				msg := " is not set up in the CLI, would you like to do it now?"
+				err = ctx.Prompt().Confirm(args[0] + msg)
+				if err != nil {
+					return err
+				}
+
+				// We do not want to attach the linkable cluster.
+				attach := false
+				ctx.Setup(setupFlags, args[0], attach)
```
bamarni: `linkableClusterConfig` must be set to the newly configured cluster, thus it'd make sense to return it from the Setup function.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			if attachedCluster.Config() == linkableClusterConfig {
```
bamarni: This wouldn't compare equality as these are pointers. You could either compare their paths, or even skip this check and let the API return an error.

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,129 @@
...
+			attachedClient := ctx.HTTPClient(attachedCluster)
+			resp, err := attachedClient.Post("/cluster/v1/links", "application/json", bytes.NewReader(message))
+			if err != nil {
+				return err
+			}
+
+			defer resp.Body.Close()
+
+			if resp.StatusCode != 200 {
+				var apiError *login.Error
```
bamarni: Why using a `login.Error` here?

ArmandGrillet: Should I create a `link.Error`? It will contain the same as `login.Error` thus I'm thinking about creating a common `Error` struct for the error messages returned by the API.

bamarni: A common `api.Error` sounds good.

pkg/cmd/cluster/cluster_unlink.go
```diff
@@ -0,0 +1,30 @@
...
+		Use:  "unlink",
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			attachedCluster, err := ctx.Cluster()
+			if err != nil {
+				return err
+			}
+
+			manager := ctx.ConfigManager()
+			linkedClusterConfig, err := manager.Find(args[0], false)
```
bamarni: The error here is not used, how about logging it for debugging purpose?

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,103 @@
...
+			var filteredProviders []*login.Provider
+
+			// Not all login providers are supported for a cluster link.
+			for _, provider := range providers {
+				switch provider.Type {
+				case login.DCOSUIDPassword,
+					login.DCOSUIDPasswordLDAP,
+					login.SAMLSpInitiated,
+					login.OIDCAuthCodeFlow:
+					filteredProviders = append(filteredProviders, provider)
```
bamarni: nit, maybe a blank line before this line would improve readability.

pkg/setup/setup.go
```diff
@@ -103,9 +106,9 @@ func (s *Setup) Configure(flags *Flags, clusterURL string) error {
 	// Create the config for the given cluster and attach the cluster.
 	err = s.configManager.Save(cluster.Config(), metadata.ClusterID, flags.caBundle)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	return s.configManager.Attach(cluster.Config())
+	return cluster, s.configManager.Attach(cluster.Config())
```
bamarni: `opts.Attach` is not taken into account here

pkg/config/manager.go
```diff
@@ -30,6 +30,12 @@ type Manager struct {
 	dir       string
 }
 
+// ErrConfigNotFound means that the manager cannot find a config using a name/id.
+var ErrConfigNotFound = errors.New("no match found")
+
+// ErrTooManyConfigs means that there are more than one config with a given name.
```
bamarni: This is not necessarily `more than one config with a given name`:

`ErrTooManyConfigs means that there are more than one config results for a given search.`

pkg/cmd/cluster/cluster_setup.go
```diff
@@ -18,11 +16,10 @@ func newCmdClusterSetup(ctx api.Context) *cobra.Command {
...
 
-			// This supports passing a cluster URL without scheme, it then defaults to HTTPS.
-			if !strings.HasPrefix(clusterURL, "https://") && !strings.HasPrefix(clusterURL, "http://") {
-				clusterURL = "https://" + clusterURL
-			}
-			return ctx.Setup(setupFlags, clusterURL)
+			// We want to attach the cluster once it is setup.
+			attach := true
+			_, err := ctx.Setup(setupFlags, clusterURL, attach)
+			return err
```
bamarni: I find it clearer without the `attach` argument, it removes the need to pass a variable for readability and avoids the blank-assignment, while not being much more verbose : 
``` go
config, err := ctx.Setup(setupFlags, clusterURL)
if err != nil {
    return err
}
return ctx.ConfigManager().Attach(config)
```

pkg/cmd/cluster/cluster_link.go
```diff
@@ -0,0 +1,101 @@
...
+				if err == config.ErrConfigNotFound {
+					msg := " is not set up in the CLI, would you like to do it now?"
+					err = ctx.Prompt().Confirm(args[0] + msg)
+					if err != nil {
+						return err
+					}
+
+					// We do not want to attach the linkable cluster.
+					attach := false
+					ctx.Setup(setupFlags, args[0], attach)
```
bamarni: This could simply become `ctx.Setup(setupFlags, args[0])` if we leave the attachment responsibility to the caller.

pkg/clusterlinker/client.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// Client is able to detect available login providers and login to DC/OS.
+type Client struct {
+	http   *httpclient.Client
+	logger *logrus.Logger
+}
+
+type LoginProvider struct {
```
bamarni: Missing comment, there are also a few other places : 
```
pkg/clusterlinker/client.go:19:6: exported type LoginProvider should have comment or be unexported
pkg/clusterlinker/client.go:24:6: exported type LinkRequest should have comment or be unexported
pkg/clusterlinker/client.go:39:1: exported method Client.Link should have comment or be unexported
pkg/clusterlinker/client.go:60:1: exported method Client.Unlink should have comment or be unexported
pkg/dcos/error.go:3:6: exported type Error should have comment or be unexported
```

PR #1250, opened 2018-06-21
Added plugin system to DC/OS Go CLI.

ArmandGrillet:
Based on Brian's PR #1249, this adds Windows support and fixes regarding the architecture of the `plugin` package.

One thing we have noticed during the tests of this plugin system: the output of the logs on Windows is not the same as on macOS. This is due to Logrus and not our code thus we have not fixed the problem in this PR.

### DIFF
--- a/pkg/plugin/manager.go
+++ b/pkg/plugin/manager.go
@@ -1,0 +1,209 @@
+package plugin
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Manager retrieves the plugins available for the current cluster
+// by navigating into the filesystem. It requires a `plugin.yaml` file for
+// each plugin, and can create this file if it doesn't exist yet.
+type Manager struct {
+	Fs     afero.Fs
+	Logger *logrus.Logger
+	Dir    string
+	Stdout io.Writer
+	Stderr io.Writer
+	Stdin  io.Reader
+}
+
+// Package represents information of old plugins found in Universe packages.
+type Package struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Version     string `json:"version"`
+}
+
+// Plugins returns the plugins associated with the current cluster.
+func (m *Manager) Plugins() []*Plugin {
+	plugins := []*Plugin{}
+
+	pluginsDirHandle, err := m.Fs.Open(m.Dir)
+	if err != nil {
+		// If we fail to open the directory, we return an empty list.
+		return plugins
+	}
+	defer pluginsDirHandle.Close()
+
+	pluginDirs, err := pluginsDirHandle.Readdir(0)
+	if err != nil {
+		// If we fail to open the directory, we return an empty list.
+		return plugins
+	}
+
+	for _, pluginDir := range pluginDirs {
+		if pluginDir.IsDir() {
+			plugin := &Plugin{}
+
+			// Set plugin directories.
+			plugin.dir = filepath.Join(m.Dir, pluginDir.Name())
+			plugin.BinDir = filepath.Join(plugin.dir, "env", "bin")
+
+			if runtime.GOOS == "windows" {
+				exists, _ := afero.DirExists(m.Fs, plugin.BinDir)
+				if !exists {
+					plugin.BinDir = filepath.Join(plugin.dir, "env", "Scripts")
+				}
+			}
+
+			pluginFilePath := filepath.Join(plugin.dir, "env", "plugin.yaml")
+			data, err := afero.ReadFile(m.Fs, pluginFilePath)
+			if err == nil {
+				// We don't want to see the CLI failing if a single plugin is malformed.
+				// We thus log the error but continue if there is an issue at that step.
+				err = yaml.Unmarshal(data, plugin)
+				if err != nil {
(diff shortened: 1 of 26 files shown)


pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+			cmd := &command{
+				Name: commandName,
+			}
+
+			// We add the command information to the plugin file by shelling it out.
+			cmdExe := filepath.Join(plugin.binDir, binary.Name())
+			infoCmd, err := exec.Command(cmdExe, commandName, "--info").Output()
+			if err != nil {
+				log.Fatal(err)
```
bamarni: This would abort the script, we should also use logrus.

pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return plugins
+}
+
+// loadPluginFromPackage fills a Plugin structure from a Universe package.
+// This is based on the assumption that old packages were all downloaded
+// using the Universe and are thus Universe packages.
+func (m *Manager) loadPluginFromPackage(plugin *Plugin) error {
+	packageFilePath := filepath.Join(plugin.dir, "package.json")
+	data, err := ioutil.ReadFile(packageFilePath)
```
bamarni: `afero.ReadFile(m.fs, packageFilePath)`

pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
+package plugin
+
+import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"log"
```
bamarni: normally we shouldn't have to import the 2 packages above, as we use afero and logrus.

pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
...
+			plugin.dir = filepath.Join(m.Dir, pluginDir.Name())
+
+			if runtime.GOOS == "windows" {
+				plugin.binDir = filepath.Join(plugin.dir, "env", "Scripts")
+			} else {
+				plugin.binDir = filepath.Join(plugin.dir, "env", "bin")
+			}
+
+			pluginFilePath := filepath.Join(plugin.dir, "env", "plugin.yaml")
+			data, err := ioutil.ReadFile(pluginFilePath)
```
bamarni: this should use afero

pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+	for _, pluginDir := range pluginDirs {
+		if pluginDir.IsDir() {
+			plugin := &Plugin{}
+
+			// Set plugin directories.
+			plugin.dir = filepath.Join(m.Dir, pluginDir.Name())
+
+			if runtime.GOOS == "windows" {
+				plugin.binDir = filepath.Join(plugin.dir, "env", "Scripts")
```
bamarni: Should we support the `bin` dir on Windows too?

pkg/plugin/manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// We should not overwrite `plugin.yaml`, this method should only be used
+	// to create a minimal `plugin.yaml` if the user has an old plugin.
+	pluginFilePath := filepath.Join(plugin.dir, "env", "plugin.yaml")
+	if _, err := os.Stat(pluginFilePath); err == nil {
+		return errors.New(pluginFilePath + " already exists")
+	}
+
+	// Create the `plugin.yaml` file.
+	pluginFile, err := m.Fs.Create(pluginFilePath)
```
bamarni: The file is not closed. Instead of creating / writing to the file, you could use `afero.WriteFile(m.Fs, pluginFilePath, pluginYAML, 0644)`

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+
+			commands = append(commands, cmd)
+		}
+	}
+
+	return commands
+}
+
+// intoCobra creates a Cobra command used to call this command.
+func (c *command) intoCobra(ctx *cli.Context, dir string, exe string) *cobra.Command {
```
bamarni: It'd be preferable not to introduce dependencies on cobra in our packages outside of `cmd`.

ArmandGrillet:
@bamarni We thought about adding the feature now and having @br-lewis adding tests during the rest of the sprint in an upcoming PR. Does it sound OK?

bamarni:
@ArmandGrillet : Sounds good, as long as we consider this as a first prototype which will help us designing and experimenting with the feature. I mean we should still be fine with adjusting the plugin schema / implementation.

pkg/plugin/manager.go
ArmandGrillet: We have tested the current DC/OS CLI on Windows and the standard path for subcommands replace `bin` by `Scripts`.

br-lewis:
I don't think moving the plugin to cobra.Command conversion outside of plugin makes the code better. 

The plugin code is somewhat independent of cobra but I think it makes more sense for the code to convert a plugin to a command to belong to the code that's concerned with plugins since plugins aren't really useful except for their ability to be converted into a command, whether cobra or not and it should really be up to the plugin to know how it maps to a command.

bamarni:
@br-lewis : having it in our package means we expose a public method returning instances of cobra commands, cobra becomes part of our API. On the other hand if it's in `cmd`, cobra isn't exposed, it remains an implementation detail which we can change in the future without worrying of breaking users code.

While I agree that the current state of the code is not optimal (there is probably some shell-out logic which can be moved to the `plugin` package, so that the cobra commands content stays tiny), it at least doesn't have the aforementioned drawback.

ArmandGrillet:
Now squashed, ready for merge 🚀.

PR #1255, opened 2018-06-28
Plugin system testing

br-lewis:
This PR adds some basic tests for loading plugins.


pkg/plugin/plugin_test.go
```diff
@@ -0,0 +1,119 @@
...
+	logger, _ := test.NewNullLogger()
+
+	m := plugin.Manager{
+		Fs:     roFs(),
+		Dir:    subcommandsDir,
+		Logger: logger,
+	}
+
+	plugins := m.Plugins()
+	require.NotEmpty(t, plugins)
```
bamarni: `assert.Equal(t, 1, len(plugins))`

pkg/plugin/plugin_test.go
```diff
@@ -0,0 +1,119 @@
...
```
bamarni: `assert.Equal(t, 1, len(plugins))`

pkg/plugin/plugin_test.go
```diff
@@ -0,0 +1,119 @@
...
+	assert.Equal(t, 2, len(plugins))
+}
+
+func targetSubcommandDir(t *testing.T, name string) string {
+	wd, err := os.Getwd()
+	require.NoError(t, err)
+
+	var testdataDir string
+	if runtime.GOOS == "windows" {
+		testdataDir = filepath.Join(wd, "testdata_windows")
```
bamarni: `testdata` is a special convention (https://golang.org/cmd/go/#hdr-Test_packages) : 

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests.

Could you add the `_windows` suffixes to the fixtures folders inside `testdata` instead?

pkg/plugin/manager.go
```diff
@@ -188,14 +188,14 @@ func (m *Manager) loadPluginFromPackage(plugin *Plugin) error {
 func (m *Manager) persist(plugin *Plugin) error {
 	// We need an env directory for `plugin.yaml`.
 	envFilePath := filepath.Join(plugin.dir, "env")
-	if _, err := os.Stat(envFilePath); os.IsNotExist(err) {
+	if _, err := m.Fs.Stat(envFilePath); os.IsNotExist(err) {
```
bamarni: :+1:

pkg/plugin/plugin_test.go
br-lewis: How about `testdata/windows/` and `testdata/nix/` instead? I feel like mixing windows and non would make things a bit cluttered.

bamarni: Why `nix`? `unix` is more commonly used in that case (eg. https://github.com/golang/go/tree/master/src/path/filepath).

br-lewis: I've usually seen *nix to refer to all Unix-like OSs but I'm not very attached to that.

PR #1261, opened 2018-07-03
Supported linked clusters in `cluster list` and `cluster attach`.

ArmandGrillet:


pkg/clusterlinker/client.go
```diff
@@ -81,3 +85,33 @@ func (c *Client) Unlink(linkedClusterID string) error {
...
+// Links returns the links of a cluster.
+func (c *Client) Links() ([]*Link, error) {
+	resp, err := c.http.Get("/cluster/v1/links")
+	if err != nil {
+		return nil, errors.New("couldn't get linked clusters")
+	}
+
+	defer resp.Body.Close()
+
+	// Common error on clusters < 1.11 and OSS clusters.
```
bamarni: Not a big fan of such comments, as they might become outdated with the time, eg. this feature is open-sourced at some point.

pkg/clusterlinker/client.go
```diff
@@ -81,3 +85,33 @@ func (c *Client) Unlink(linkedClusterID string) error {
...
+	if resp.StatusCode == 404 {
+		return nil, errors.New("couldn't find linked clusters")
```
bamarni: Why would a specific error be returned on 404s here (when the cluster linker API is not available), and not in other methods? I'd skip this condition altogether.

pkg/cmd/cluster/cluster_attach.go
```diff
@@ -13,11 +18,62 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
...
 			manager := ctx.ConfigManager()
 			conf, err := manager.Find(args[0], false)
+
+			// If we have not found any config, we will look in the linked clusters.
+			if err != nil && err != config.ErrConfigNotFound {
+				return err
+			}
+
+			if conf != nil {
+				return manager.Attach(conf)
```
bamarni: If the search matches both a locally configured cluster and a linked cluster, this would attach to the local one. I'd assume this should instead fail with an error saying there are multiple matches.

pkg/cmd/cluster/cluster_attach.go
```diff
@@ -13,11 +18,62 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
...
+				if args[0] == linkedCluster.Name {
+					matchingLinkeClusters = append(matchingLinkeClusters, linkedCluster)
+				}
+				if args[0] == linkedCluster.ID {
+					return setupAndAttach(linkedCluster, ctx)
+				}
+				if strings.HasPrefix(linkedCluster.ID, args[0]) {
+					matchingLinkeClusters = append(matchingLinkeClusters, linkedCluster)
+				}
+			}
```
bamarni:
``` go
for _, linkedCluster := range linkedClusters {
	if args[0] == linkedCluster.ID {
		return setupAndAttach(linkedCluster, ctx)
	}
	if args[0] == linkedCluster.Name || strings.HasPrefix(linkedCluster.ID, args[0]) {
		matchingLinkedClusters = append(matchingLinkedClusters, linkedCluster)
	}
}
```

pkg/cmd/cluster/cluster_attach.go
```diff
@@ -13,11 +18,62 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
...
+				return err
+			}
+
+			clusterLinker := clusterlinker.NewClient(ctx.HTTPClient(currentCluster), ctx.Logger())
+			linkedClusters, err := clusterLinker.Links()
+			if err != nil {
+				ctx.Logger().Info(err)
+			}
+
+			var matchingLinkeClusters []*clusterlinker.Link
```
bamarni: matchingLinkedClusters

pkg/clusterlinker/client.go
ArmandGrillet: I did this because the error message otherwise is not helpful, an example when using `dcos cluster attach` with a cluster running on 110:
```
cli (DCOS_OSS-3695) $ ./dcos-go cluster link http://cluster.com
Error: invalid character '<' looking for beginning of value
```
This is because we do not have a `dcos.error` when DC/OS returns a 404.

pkg/cmd/cluster/cluster_attach.go
```diff
@@ -12,11 +17,63 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
...
+				}
+				if strings.HasPrefix(linkedCluster.ID, args[0]) {
+					matchingLinkedClusters = append(matchingLinkedClusters, linkedCluster)
+				}
+			}
+
+			// We act depending on the clusters we have found.
+			switch len(matchingLinkedClusters) {
+			case 0:
+				// No maching linked cluster, one matching cluster.
```
bamarni: `matching`, on above lines too

pkg/cmd/cluster/cluster_list.go
```diff
@@ -28,13 +30,42 @@ func newCmdClusterList(ctx api.Context) *cobra.Command {
 				currentClusterID = currentCluster.ID()
 			}
 
-			var clusters []*config.Cluster
-			if !onlyAttached {
-				clusters = ctx.Clusters()
-			} else if currentCluster != nil {
-				clusters = append(clusters, currentCluster)
+			if onlyAttached && currentCluster == nil {
+				return errors.New("no cluster is attached. Please run `dcos cluster attach <cluster-name>`")
```
bamarni: When there are no clusters configured, an empty file in `~/.dcos/dcos.toml` and I run `dcos cluster list --attached`, instead of seeing this message I have :
```
     NAME   ID      STATUS     VERSION  URL  
  *        .dcos  UNAVAILABLE  UNKNOWN
```

bamarni: https://jira.mesosphere.com/browse/DCOS-38927

PR #1262, opened 2018-07-03
Completion system that works with plugins

br-lewis:
This PR uses `cobra`'s built in completion functionality to create a completion script that works with our plugin system. We had to splice in a call to reload the completions on every complete call to make it automatically track attaching to different clusters or adding or removing plugins.

This doesn't yet have help instructions for how to use it but running
```
source /dev/stdin <<<"$(dcos completion <bash or zsh>)"
```
will work to bootstrap it.


br-lewis:
We've decided to just write straight bash for completion instead of trying to hack around cobra to get it to do what we want. https://github.com/dcos/dcos-cli/pull/1286

PR #1265, opened 2018-07-06
`dcos cluster list --attached` might display an invalid item

bamarni:
https://jira.mesosphere.com/browse/DCOS-38927


pkg/cmd/cluster/cluster_list_test.go
```diff
@@ -56,3 +56,19 @@ func TestClusterListTable(t *testing.T) {
...
+
+	conf := config.New(config.Opts{Fs: env.Fs})
+	conf.SetPath("dcos.toml")
+
+	ctx := mock.NewContext(env)
+	ctx.SetCluster(config.NewCluster(conf))
+
+	cmd := newCmdClusterList(ctx)
+	cmd.SetArgs([]string{"--attached"})
+	require.Error(t, cmd.Execute())
```
bamarni: Here I couldn't test the exact error output as it is currently mixed with help usage (we need to fix this default cobra behavior at some point).

We should really move this to an existing or new package at some point, at least it would facilitate testing.

PR #1268, opened 2018-07-09
Remove Jenkinsfiles

bamarni:
They are not being used anymore since the introduction of Travis and the creation of the core CLI repo.


bamarni:
@klueska : we don't have integration tests yet for the Go CLI. Given the lessons learned from the python integration tests, we really want to get this right and this requires to invest time in it.

We do however test commands, we are mocking DC/OS endpoints and then running tests against these, see for example the `dcos auth list-providers` tests (https://github.com/dcos/dcos-cli/blob/master/pkg/cmd/auth/auth_listproviders_test.go).

bamarni:
sure

PR #1271, opened 2018-07-18
Set the logging level outside of cobra

bamarni:
Having it in the pre run hook discards all the logs from the plugin manager, this is because the plugin manager is invoked while creating the cobra command tree. The hook is only invoked later while actually calling a command.

Declaring the flag outside of cobra and setting the log level directly fixes this problem.

https://jira.mesosphere.com/browse/DCOS_OSS-3822

### DIFF
--- a/cmd/dcos/main.go
+++ b/cmd/dcos/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/user"
 	"time"
@@ -11,7 +12,9 @@ import (
 	"github.com/dcos/dcos-cli/pkg/cmd"
 	"github.com/dcos/dcos-cli/pkg/dcos"
 	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
+	"github.com/spf13/pflag"
 )
 
 var version = "SNAPSHOT"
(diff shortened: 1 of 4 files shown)


cmd/dcos/main.go
```diff
@@ -35,6 +50,22 @@ func main() {
 	}
 }
 
+// logLevel returns the log level for the CLI.
```
ArmandGrillet: You can add that the verbosity is 0 by default, this is not very explicit in `var verbosity int`.

bamarni:
For testing, logrus doesn't expose the logger level (it is an unexported method), so I had to issue some logs and check them with the test hook.

cmd/dcos/main.go
```diff
@@ -26,13 +29,45 @@ func main() {
...
+}
+
+// run launches the DC/OS CLI with a given context and args.
+func run(ctx api.Context, args []string) error {
+	var verbosity int
+	var showVersion bool
+
+	// Register the version and verbose global flags.
+	globalFlags := pflag.NewFlagSet(args[0], pflag.ContinueOnError)
+	globalFlags.SetOutput(ioutil.Discard)
```
klueska: Why do we discard errors here?

cmd/dcos/main.go
```diff
@@ -26,13 +29,45 @@ func main() {
 		Fs:         afero.NewOsFs(),
 	})
 
-	if len(os.Args) == 2 && os.Args[1] == "--version" {
+	if err := run(ctx, os.Args); err != nil {
```
klueska: I don't know enough about how you propagate errors in the code base for logging purposes. Is it OK to throw away the error returned here and not log it? I assume you log it further down the call stack otherwise?

cmd/dcos/main.go
bamarni: This hybrid solution with a flagset outside of cobra is not the most robust. I've added some comments about it. Hopefully we'll clean this out when moving away from cobra at some point.

cmd/dcos/main.go
bamarni: Yeah this is handled by cobra, it will print errors by default when using the `RunE` method and `SilenceErrors` is not set to true.

bamarni:
Oh, I missed that Level is an exported field in the logger, let me simplify the tests...

bamarni:
Updated, I've kept the logger hook in the mock context as it can be useful if we want to assert on log entries in the future.

PR #1274, opened 2018-07-19
Update the plugin package to use TOML

bamarni:
This updates the plugin package to use TOML instead of YAML and refactors the code as well.

https://jira.mesosphere.com/browse/DCOS_OSS-3832

PR #1275, opened 2018-07-20
Auto-install dcos-enterprise-cli during cluster setup

bamarni:
Unless the --no-plugin flag is passed, `dcos cluster setup` will attempt
to install the dcos-enterprise-cli plugin. The dcos-enterprise-cli
plugin is only installed when the metadata endpoint explicitly specifies
that it is an enterprise variant.

The plugin resources are retrieved through the Cosmos client.

The plugin manager currently only supports installing a ZIP plugin from
the filesystem. Support for remote and bin resources will be added with
the future `dcos plugin add` command.

https://jira.mesosphere.com/browse/DCOS-20502

### DIFF
--- a/pkg/cosmos/client.go
+++ b/pkg/cosmos/client.go
@@ -1,0 +1,104 @@
+package cosmos
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/url"
+
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+)
+
+// Client is a client for Cosmos.
+type Client struct {
+	http *httpclient.Client
+}
+
+// NewClient creates a new Cosmos client.
+func NewClient(baseClient *httpclient.Client) *Client {
+	return &Client{
+		http: baseClient,
+	}
+}
+
+// PackageInfo contains information about a DC/OS package.
+// It is mirroring the JSON response of the `/package/describe`
+// endpoint for unmarshalling convenience.
+type PackageInfo struct {
+	Package struct {
+		PackagingVersion      string `json:"packagingVersion"`
+		Name                  string `json:"name"`
+		Description           string `json:"description"`
+		Version               string `json:"version"`
+		ReleaseVersion        int    `json:"releaseVersion"`
+		MinDCOSReleaseVersion string `json:"minDcosReleaseVersion"`
+		Framework             bool   `json:"framework"`
+		Maintainer            string `json:"maintainer"`
+		Resource              struct {
+			CLI struct {
+				Plugins map[string]map[string]*Plugin `json:"binaries"`
+			} `json:"cli"`
+		} `json:"resource"`
+	} `json:"package"`
+}
+
+// Plugin represents a CLI plugin resource.
+type Plugin struct {
+	Kind string `json:"kind"`
+	URL  string `json:"url"`
+}
+
+// DescribePackage returns information about the named package.
+func (c *Client) DescribePackage(name string) (*PackageInfo, error) {
+	reqBodyPayload := map[string]string{"packageName": name}
+
+	var reqBody bytes.Buffer
+	if err := json.NewEncoder(&reqBody).Encode(reqBodyPayload); err != nil {
+		return nil, err
+	}
+
+	req, err := c.http.NewRequest("POST", "/package/describe", &reqBody, httpclient.FailOnErrStatus(true))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set(
+		"Content-Type",
+		"application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1",
+	)
+	req.Header.Set(
+		"Accept",
+		"application/vnd.dcos.package.describe-response+json;charset=utf-8;version=v3",
+	)
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	var pkg PackageInfo
+	err = json.NewDecoder(resp.Body).Decode(&pkg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Workaround for a Cosmos bug leading to wrong schemes in plugin resource URLs.
+	// This happens on setups with TLS termination proxies, where Cosmos might rewrite
+	// the scheme to HTTP while it is actually HTTPS. The other way around is also possible.
+	// See https://jira.mesosphere.com/browse/COPS-3052 for more context.
+	//
+	// To prevent this we're rewriting such URLs with the scheme set in `core.dcos_url`.
+	httpClientBaseURL := c.http.BaseURL()
+	for _, plugins := range pkg.Package.Resource.CLI.Plugins {
+		for _, plugin := range plugins {
+			pluginURL, err := url.Parse(plugin.URL)
+			if err != nil {
+				continue
+			}
+			if pluginURL.Hostname() == httpClientBaseURL.Hostname() {
+				pluginURL.Scheme = httpClientBaseURL.Scheme
+				plugin.URL = pluginURL.String()
+			}
+		}
+	}
+	return &pkg, err
+}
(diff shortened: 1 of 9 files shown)


bamarni:
This also depends on #1274 for the runtime part.

pkg/cosmos/client.go
```diff
@@ -0,0 +1,95 @@
...
+			CLI struct {
+				Plugins map[string]map[string]*Plugin `json:"binaries"`
+			} `json:"cli"`
+		} `json:"resource"`
+	} `json:"package"`
+}
+
+// Plugin represents a CLI plugin resource.
+type Plugin struct {
+	Kind string `json:"kind"`
```
ArmandGrillet: `Kind` makes sense as it is used in https://docs.mesosphere.com/1.11/deploying-services/package-api/#/default/POST_package_describe :+1:

pkg/cosmos/client_test.go
```diff
@@ -0,0 +1,85 @@
...
+					URL:  "https://linux.example.com/dcos-test-cli.zip",
+				},
+			},
+			"darwin": map[string]*Plugin{
+				"x86-64": &Plugin{
+					Kind: "zip",
+					URL:  "http://darwin.example.com/dcos-test-cli.zip",
+				},
+			},
+			// Test case for https://jira.mesosphere.com/browse/COPS-3052.
```
ArmandGrillet: The COPS ticket is closed and cannot be opened by devs outside of Mesosphere, a description of the problem handling this test case check would be more useful.

pkg/cosmos/client.go
```diff
@@ -0,0 +1,95 @@
...
+	defer resp.Body.Close()
+
+	var pkg PackageInfo
+	err = json.NewDecoder(resp.Body).Decode(&pkg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Workaround for a Cosmos bug leading to wrong schemes in plugin resource URLs.
+	// See https://jira.mesosphere.com/browse/COPS-3052 for more context.
```
ArmandGrillet: The COPS ticket is closed and cannot be opened by devs outside of Mesosphere, a description of the problem handling this test case check would be more useful.

pkg/httpclient/client.go
```diff
@@ -24,13 +26,18 @@ type Option func(opts *Options)
...
-	CheckRedirect func(req *http.Request, via []*http.Request) error
+	Timeout         time.Duration
+	TLS             *tls.Config
+	Logger          *logrus.Logger
+	ACSToken        string
+	CheckRedirect   func(req *http.Request, via []*http.Request) error
+	FailOnErrStatus bool
 }
 
+type ctxKey int
```
ArmandGrillet: `ctxKey` and `ctxKeyFailOnErrStatus` are private but I have trouble seeing their purpose without commands, can you add some?

pkg/plugin/manager.go
```diff
@@ -34,6 +35,24 @@ type Package struct {
 	Version     string `json:"version"`
 }
 
+// Install installs a plugin.
+func (m *Manager) Install(name, path string) error {
```
ArmandGrillet: This is really clean 👏 👏 👏 .

pkg/setup/flags.go
```diff
@@ -39,6 +40,12 @@ func (f *Flags) Register(flags *pflag.FlagSet) {
 		false,
 		"Do not check CA certficate downloaded from cluster (insecure). Applies to Enterprise DC/OS only.",
 	)
+	flags.BoolVar(
+		&f.noPlugin,
+		"no-plugin",
+		false,
+		"Do not auto-install dcos-core-cli and dcos-enterprise-cli plugins.",
```
ArmandGrillet: A little more descriptive: `Do not install the core and enterprise plugins locally after setup`.

pkg/setup/setup.go
```diff
@@ -257,3 +272,63 @@ Do you trust it?`
...
+			return nil
+		}
+		s.logger.Debug(err)
+	}
+
+	s.logger.Error("Please run “dcos package install dcos-enterprise-cli” if you use a DC/OS Enterprise cluster.")
+	return nil
+}
+
+// installPlugin installs a plugin by its name. It gets the plugin download URL through Cosmos.
```
ArmandGrillet: s/`plugin download URL`/`plugin's download URL`

pkg/httpclient/client.go
bamarni: I've added some comments. Also as a side note the custom type is the recommended way to go in such cases, from https://golang.org/pkg/context/#WithValue : 
> Users of WithValue should define their own types for keys.

pkg/setup/flags.go
bamarni: I don't think it's better, the auto-installation would now be part of the setup, not after, it just happens to be the last step.

pkg/cosmos/client.go
bamarni: I've added more comments.

pkg/cosmos/client_test.go
bamarni: I removed this link and just put a simple comment as there is already a long description in the associated code.

bamarni:
Thanks for reviewing @ArmandGrillet! I've also added a bunch of comments in `installPlugin`.

bamarni:
@klueska : as discussed I've changed the code so that it doesn't have to hardcode the 3 minutes default timeout, see the last commit.

bamarni:
@ArmandGrillet : I had to rebase against the other PR so it discarded your review. #1279 and this PR have 6 commits in common. The other PR is probably easier to review first as it has less changes, then once merged this one will have a much shorter diff.

pkg/setup/flags.go
```diff
@@ -39,6 +40,12 @@ func (f *Flags) Register(flags *pflag.FlagSet) {
 		false,
 		"Do not check CA certficate downloaded from cluster (insecure). Applies to Enterprise DC/OS only.",
 	)
+	flags.BoolVar(
+		&f.noPlugin,
+		"no-plugin",
```
bamarni: There were some disagreement from @klueska on the naming here in our sync meeting, mainly because it's not explicit enough on what it is supposed to do.

IMO we shouldn't make flag names too long as it's hurting usability and doesn't fit well with other flags we have : 
```
      --ca-certs string        Specify the path to a file with trusted CAs to verify requests against.
  -h, --help                   help for setup
      --insecure               Allow requests to bypass TLS certificate verification (insecure).
      --name string            Specify a custom name for the cluster.
      --no-check               Do not check CA certficate downloaded from cluster (insecure). Applies to Enterprise DC/OS only.
      --password string        Specify the password on the command line (insecure).
      --password-env string    Specify an environment variable name that contains the password.
      --password-file string   Specify the path to a file that contains the password (insecure).
      --private-key string     Specify the path to a file that contains the service account private key.
      --provider string        Specify the authentication provider to use for login.
      --username string        Specify the username for login.
```

Here we can see that no flag is longer than 2 words, some flags are not that clear until we look at their description (eg. --insecure or --no-check).

I'm not saying that the current naming is the right way to name flags but at least it seems consistent. If we decide introduce a long flag here we should also move towards this approach on other flags too on the long run (eg. consider deprecating some unclear ones).

cc @leemunroe @AmrAbdelrazik

bamarni: There seems to be 2 schools of thoughts here. Below is a comparison of the same flags on different CLIs:

**curl**
```
--insecure       Allow insecure server connections when using SSL
--cacert <file>  CA certificate to verify peer against
```

**kubectl config set-cluster**
```
--insecure-skip-tls-verify   insecure-skip-tls-verify for the cluster entry in kubeconfig
--certificate-authority      Path to certificate-authority file for the cluster entry in kubeconfig
```

**curl** tends to prefer short flag names with clear help usages, while **kubectl** emphasizes on long and explicit flag names, while their help usages are not that informative.

Given that we only have to **learn once** what a flag is about but then **use it many times**, I think we should go with the **curl** approach. This is also in-line with the ones we already have (`--insecure` and `--ca-cert`).

PR #1279, opened 2018-07-26
Introduce the `dcos plugin add` command

bamarni:
- Add a FailOnErrStatus option to httpclient
- Add an Unzip helper function in fsutil
- Fix some debug logs in httpclient
- Unexport plugin manager's fields
- Add copy helpers in fsutil
- Add an Install method to the plugin manager
- Introduce the `dcos plugin add` command

Usage example : 
```
$ dcos plugin add https://downloads.dcos.io/cli/plugins/dcos-core-cli/linux/x86-64/dcos-core-cli-1.12.zip        
        
$ dcos plugin add https://downloads.mesosphere.io/cli/binaries/linux/x86-64/1.4.5/61d301f571fe26883b5122d567ac4b79bae7feb
bd2090c81b6cdda523659eb43/dcos-enterprise-cli                                                                                                            

$ dcos plugin list                                                                                                       
         NAME                         COMMANDS
  dcos-core-cli        job marathon node package service task
  dcos-enterprise-cli  backup license security
```


ArmandGrillet:
Could you cut the PR? The commits `Add a FailOnErrStatus option to httpclient` and `Unexport plugin manager's fields` seem to be out of the scope. Otherwise LGTM.

bamarni:
`FailOnErrStatus` is being used in the plugin manager, when downloading resources.

While unexporting the fields was not absolutely necessary for this PR it'd have been a waste of time to use the old way here and update the changes in a second PR. Removing just this commit now would break the code.

ArmandGrillet:
From Lee's yesterday email: "Bilal also demo'ed the work on `dcos plugin add …` and we spotted a couple of minor inconsistencies and were able to correct them on the spot.". Does this mean that this PR needs to be updated?

bamarni:
@ArmandGrillet : mainly the comments were that the argument is missing the help menu. I've opened #1281 as this the case for all commands and should rather be addressed globally. We also discussed about potentially adding a version in the plugin.toml and the plugin list table, but this is not directly related to this PR.

PR #1281, opened 2018-07-30
Add argument names in help menus

bamarni:
Current usage outputs :
```
Usage:
  dcos cluster setup [flags]
```
This changes them to print accepted argument names too :
```
Usage:
  dcos cluster setup <url> [flags]
```

### DIFF
--- a/pkg/cmd/cluster/cluster_rename.go
+++ b/pkg/cmd/cluster/cluster_rename.go
@@ -8,7 +8,7 @@ import (
 // newCmdClusterRename renames a cluster.
 func newCmdClusterRename(ctx api.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "rename",
+		Use:   "rename <cluster> <name>",
 		Short: "Rename a configured cluster",
 		Args:  cobra.ExactArgs(2),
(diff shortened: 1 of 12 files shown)


pkg/cmd/cluster/cluster_rename.go
```diff
@@ -8,7 +8,7 @@ import (
 // newCmdClusterRename renames a cluster.
 func newCmdClusterRename(ctx api.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "rename",
+		Use:   "rename <cluster> <new_name>",
```
ArmandGrillet: Shouldn't it be `<new-name>`? Our commands (e.g. `list-providers`) use dashes.

bamarni: I took it out from the current CLI help usage (`dcos cluster rename <name> <new_name>`), I will try to find if there is a preferred notation amongst CLIs in general.

We could also call it `<name>` here.

ArmandGrillet: Yes, let's use `<name>`.

bamarni: updated

PR #1283, opened 2018-08-02
Install the DC/OS Core CLI when setting up a cluster.

ArmandGrillet:


pkg/setup/setup.go
```diff
@@ -278,20 +278,27 @@ Do you trust it?`
 
 // installPlugins installs the dcos-core-cli and (if applicable) the dcos-enterprise-cli plugin.
 func (s *Setup) installPlugins(httpClient *httpclient.Client) error {
-	// TODO: install dcos-core-cli.
+	// Install dcos-core-cli.
+	coreInstall := make(chan error)
```
bamarni: `coreInstall` -> `coreInstallErr`

PR #1286, opened 2018-08-10
Bash complete

br-lewis:
This adds bash completion for a small subset of the commands currently available in the wrapper CLI along with a command that can output this completion script.


pkg/cmd/completion/bindata.go
```diff
@@ -0,0 +1,233 @@
+// Code generated by go-bindata. DO NOT EDIT. @generated
```
ArmandGrillet: According to `golint`, we have to edit this file 😅:
> pkg/cmd/completion/bindata.go:1:1: package comment should be of the form "Package completion ..."
We can either disable the linter for this package or add a comment of the excepted form at the first line.

br-lewis: We instead try the proposed workaround from https://github.com/golang/lint/issues/90 and have it output to something like `completiondata.gen.go` like the code generated by `goderive` does.

bamarni: The reason why goderive works is because they follow the Go convention for ignoring their files during linting (https://github.com/golang/go/issues/13560#issuecomment-288457920), while go-bindata is not and is unmaintainted anyways. I'm fine with whatever solution as long as we agree that it is temporary and should move away from go-bindata.

For the record, if we do not rely on backticks in the bash file (and there shouldn't be a strong reason for doing so), it should be pretty simple to generate a go file containing a raw string literal, see for example in the Go project how they generate these kind of files with static data (https://github.com/golang/go/blob/master/src/syscall/mksysnum_plan9.sh).

bamarni: It was working and they broke it actually : https://github.com/go-bindata/go-bindata/commit/e1b4998d06bcb1c2c20934a20bd6f81d1bc30823

I think the real project was https://github.com/jteeuwen/go-bindata originally, looking at the number of stars.

bamarni:
@br-lewis : there is a merge conflict

PR #1294, opened 2018-08-14
Inject CLI version when building the release binaries

bamarni:
Currently releases would only show SNAPSHOT. This updates the Jenkinsfile to use the commit hash ID for master builds and GIT tags for tag builds.


br-lewis:
Wouldn't this make the version that `dcos --version` outputs be a commit hash and not the CLI's current version scheme?

bamarni:
@br-lewis : There are 2 cases, tag builds and builds on pushes to master, commit hashes would only be for the latter.

Eg. of a tag build : https://downloads.dcos.io/binaries/cli/linux/x86-64/0.7.0/dcos

PR #1296, opened 2018-08-15
Rely on mitchellh/go-homedir instead of os/user

bamarni:
Integration tests on Windows highlighted a problem with the home directory detection, it happened because os/user doesn't work well with cross-compilation.

Once we upgrade to Go 1.11 we should be able to fix this simply with a build tag (https://tip.golang.org/doc/go1.11#os/user).

https://github.com/mitchellh/go-homedir#go-homedir
https://jira.mesosphere.com/browse/DCOS_OSS-3953


bamarni:

bamarni:

bamarni:
(The first comment didn't trigger them somehow..)

PR #1311, opened 2018-08-23
Cluster list json fix

br-lewis:
Changes `"id"` field from `cluster list --json` output to `"cluster_id"` to preserve compatibility with json from previous versions


bamarni:

PR #1313, opened 2018-08-24
Fixed plugin install during cluster setup

rgo3:
This commit fixes, that we potentially don't wait for the
enterprise plugin to be installed correctly if the install
for the core plugin fails. We now wait for both installs
to finish before proceeding.

https://jira.mesosphere.com/browse/DCOS_OSS-4003

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -283,17 +283,19 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
 		close(enterpriseInstallErr)
 	}()
 
-	err = s.installPlugin("dcos-core-cli", httpClient)
-	if err != nil {
-		// We don't return an error as the EE plugin is not as useful as the core plugin.
-		return fmt.Errorf("unable to install DC/OS core CLI plugin: %s", err)
-	}
-
+	// Install dcos-core-cli.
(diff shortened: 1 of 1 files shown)


rgo3:

pkg/setup/setup.go
```diff
@@ -283,17 +283,19 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
...
 	// The installation of the core and EE plugins happen in parallel.
 	// We wait for the installation of the enterprise plugin before returning.
-	err = <-enterpriseInstallErr
-	if err != nil {
-		return fmt.Errorf("unable to install DC/OS enterprise CLI plugin: %s", err)
+	errEnterprise := <-enterpriseInstallErr
+
+	switch {
+	case errCore != nil && errEnterprise != nil:
+		return fmt.Errorf("unable to install DC/OS core and enterprise CLI plugins:\n%s\n%s", errCore, errEnterprise)
```
bamarni: Multi-errors are tricky to handle, eg. there are some packages just on this topic (https://github.com/hashicorp/go-multierror). If we were to support these I think this should be done in a generic way while we redesign our error handling, for now here it is fine to just return the core error only, as long as we've waited for enterprise:
``` go
if errCore != nil {
    return errCore
}
return errEnterprise
```

rgo3:

PR #1316, opened 2018-08-28
Change the setup flow to install a default bundled plugin if Cosmos installation fails

br-lewis:
Prototype design for having the wrapper CLI bundle a core plugin that can be automatically installed if the user doesn't have access to Cosmos.


pkg/plugin/manager.go
```diff
@@ -88,6 +89,29 @@ func (m *Manager) Install(resource string, installOpts *InstallOpts) (err error)
 	return nil
 }
 
+// InstallDefaultPlugin installs the default core plugin bundled with the wrapper cli if something
+// went wrong with installing the plugin from Cosmos
```
bamarni: For consistency this should rather stay an implementation detail under the `setup` package, the plugin package has no dependency on cosmos.

pkg/setup/generate_bundle_stub.go
```diff
@@ -0,0 +1,14 @@
+//+build !withbundle
+
+package setup
+
+//go:generate go-bindata -pkg setup -o bundled_plugins_stub_linux.gen.go -ignore=\.gitignore -nometadata -dev -tags "!withbundle" -prefix "bundled_plugins" bundled_plugins/linux/
+//go:generate go-bindata -pkg setup -o bundled_plugins_stub_darwin.gen.go -ignore=\.gitignore -nometadata -dev -tags "!withbundle" -prefix "bundled_plugins" bundled_plugins/darwin/
+//go:generate go-bindata -pkg setup -o bundled_plugins_stub_windows.gen.go -ignore=\.gitignore -nometadata -dev -tags "!withbundle" -prefix "bundled_plugins" bundled_plugins/windows/
```
bamarni: I don't understand why we need to generate stuff here, as go-bindata relies on an `Asset` function, couldn't this file only define this "Asset" function and it'd return an error like `no core CLI is bundled in the binary`?

pkg/setup/setup.go
```diff
@@ -285,12 +289,22 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
...
 	// We wait for the installation of the enterprise plugin before returning.
 	errEnterprise := <-enterpriseInstallErr
 
 	if errCore != nil {
-		return errCore
+		// check that the version is 1.12 and if so, try to install the bundled plugin
+		if versionNumber(version.Version) == "1.12" {
+			err = s.installBundledPlugin()
+			if err != nil {
+				// We don't return an error as the EE plugin is not as useful as the core plugin.
```
bamarni: There are a lot of nested levels here, usually for readability I try to return early whenever possible in such cases : 
``` go
if versionNumber(version.Version) != "1.12" {
    return fmt.Errorf("unable to install DC/OS core CLI plugin")
}
err = s.installBundledPlugin()
// ...
```

pkg/setup/setup.go
```diff
@@ -285,12 +289,22 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
...
+				return fmt.Errorf("unable to install DC/OS core CLI plugin: %s", err)
```
bamarni: We use normal sentences for log messages, eg. `Unable to install DC/OS core CLI plugin.`

br-lewis: That isn't a log message though, it's an error message where you aren't supposed to write a normal sentence.

pkg/setup/generate_bundle_stub.go
br-lewis: Doing that gives us the ability to actually run the code and test this without needing to generate 3 real bundles and recompile. I suppose taking it out might be an improvement but then I think we should change our makefile to at least make it possible to easily generate a bundle without needing to use `GO_BUILD_TAGS` manually.

tests/integration/test_cluster.py
```diff
@@ -37,3 +37,17 @@ def test_empty_cluster_list():
     assert code == 0
     assert err == ''
     assert out == '[]\n'
+
+
+def test_cluster_setup_non_superuser(default_cluster_with_plugins):
+    username='nonsuperuser'
+    password='nonsuperpassword'
+    code, out, err = exec_cmd(['dcos', 'security', 'org', 'users',
```
bamarni: `exec_cmd` without assigning vars is enough here.

tests/integration/test_cluster.py
```diff
@@ -37,3 +37,17 @@ def test_empty_cluster_list():
...
+                               'create', username, '--password={}'.format(password)])
+    # assert code == 0
+
+    code, out, err = exec_cmd(['dcos', 'cluster', 'setup', default_cluster_with_plugins['dcos_url'],
+                               '--name={}'.format(default_cluster_with_plugins['name']),
```
bamarni: `--name` is not needed here

pkg/setup/setup.go
```diff
@@ -285,14 +291,27 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
...
 
 	if errCore != nil {
-		return errCore
+		// check that the version is 1.12 and if so, try to install the bundled plugin
+		if versionNumber(version.Version) != "1.12" {
+			return fmt.Errorf("unable to install DC/OS core CLI plugin: %s", errCore)
+		}
+		err = s.installBundledPlugin()
+		if err != nil {
+			// We don't return an error as the EE plugin is not as useful as the core plugin.
```
bamarni: the comment doesn't match the code

pkg/setup/setup.go
```diff
@@ -358,3 +377,41 @@ Do you trust it?`
...
+	// Write out the data into a temp directory so that it's in the real filesystem for buildPlugin
+	bundleTempDir, err := afero.TempDir(s.fs, os.TempDir(), "dcos-default-plugin")
+	if err != nil {
+		return err
+	}
+	bundleZipPath := path.Join(bundleTempDir, "dcos-core-cli.zip")
+	err = afero.WriteFile(s.fs, bundleZipPath, pluginData, 0644)
+	if err != nil {
+		return err
+	}
```
bamarni: Having a temp dir is not strictly needed, It'd be simpler to write to a temp file directly here :
``` go
f, err := afero.TempFile(s.fs, os.TempDir(), "dcos-core-cli")
if err != nil {
	return err
}
_, err = f.Write(ca)
if err != nil {
	return err
}
```

pkg/setup/setup.go
```diff
@@ -358,3 +377,41 @@ Do you trust it?`
...
+//
+// This will occur for two primary reasons:
+// 1. The cluster and the computer running setup are airgapped. By default the resources describing
+// where the core plugins are point to S3 which will be unreachable in an airgapped environment.
+// 2. The user running setup does not have permission to access Cosmos (dcos:adminrouter:package).
+//
+// Though a user could install the core plugin manually, this will prevent usability regressions until
+// other features of DC/OS are there to allow the normal path to handle all cases.
+func (s *Setup) installBundledPlugin() error {
+	pluginData, err := Asset(path.Join(runtime.GOOS, bundledCoreFilename))
```
bamarni: If we change the Makefile's `bundle` target to put the platform in the prefix, this could simply be `Asset("core.zip")`.

pkg/setup/setup.go
```diff
@@ -31,8 +32,11 @@ import (
 	"github.com/spf13/afero"
 )
 
+const bundledCoreFilename="core.zip"
```
bamarni: Not sure if we need this as it's used from a single place and is not really meant to be exposed.

Makefile
```diff
@@ -14,8 +14,15 @@ default:
 darwin linux windows: docker-image
 	$(call inDocker,env GOOS=$(@) go build \
 		-ldflags '-X $(PKG)/pkg/cli/version.version=$(VERSION)' \
+		-tags '$(GO_BUILD_TAGS)' \
 		-o build/$(@)/dcos$($(@)_EXE) ./cmd/dcos)
 
+.PHONY: bundle
+bundle: docker-image
```
bamarni: it sounds a bit generic, `corebundle`?

bamarni:
Looks almost ready! I've opened https://github.com/br-lewis/dcos-cli/pull/2 with minor updates and https://github.com/dcos/dcos-cli/pull/1325 regarding the CI failure on AppVeyor.

PR #1320, opened 2018-08-30
CLI will no longer return error return code if a cluster was successfully set up except but dcos-enterprise-cli fails to install.

br-lewis:


pkg/setup/setup.go
```diff
@@ -292,7 +292,11 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
 	if errCore != nil {
 		return errCore
 	}
-	return errEnterprise
+
+	if errEnterprise != nil {
+		s.logger.Error(fmt.Sprintf("cluster setup was successful but dcos-enterprise-cli could not be installed: %s", errEnterprise))
```
ArmandGrillet: Could the line be simplified using `s.logger.Errorf(format string, args ...interface{})` instead?

pkg/setup/setup.go
```diff
@@ -292,7 +292,11 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
 	if errCore != nil {
 		return errCore
 	}
-	return errEnterprise
+
+	if errEnterprise != nil {
+		s.logger.Errorf("cluster setup was successful but dcos-enterprise-cli could not be installed: %s", errEnterprise)
```
bamarni: The log message should start with a capital letter. "Cluster setup was sucessful" is implicit with the exit code so could be removed.

br-lewis: I added that to make it clear that while there was an error, the cluster was still set up properly. I know the error code will signal that but a normal user of the cluster probably won't check that especially if they're told something errored out.

That may not be the ideal place for something like that since it is a log message. An alternative could maybe be some final line printed to `ctx.Out()` which says essentially (setup was fine, enterprise failed) that and leaving the log line as just saying the enterprise cli install failed.

pkg/setup/setup.go
```diff
@@ -292,7 +292,11 @@ func (s *Setup) installDefaultPlugins(httpClient *httpclient.Client) error {
 	if errCore != nil {
 		return errCore
 	}
-	return errEnterprise
+
+	if errEnterprise != nil {
+		s.logger.Errorf("Cluster setup was successful but dcos-enterprise-cli could not be installed: %s.", errEnterprise)
```
bamarni: Can we print `In order to install the "dcos-enterprise-cli" plugin, make sure your user has the "dcos:adminrouter:package" permission and run "dcos package install dcos-enterprise-cli".`?

Then once we have instructions for air-gapped we'll also add a note about that.

br-lewis:
Closing because this change ended up included in https://github.com/dcos/dcos-cli/pull/1316

PR #1322, opened 2018-09-02
Accept global flags only at the root level

bamarni:
Currently `dcos {plugin} command -v` is supported, it sets the verbosity
to 1 but the `-v` flag is still passed to the subcommand.

This changes the global flags parsing to only accept them at the root
level, they also get stripped from the environment so that there is no
need to implement dummy flags in Cobra. This is done by using a custom
GlobalFlags struct instead of pflags.

https://jira.mesosphere.com/browse/DCOS_OSS-4026


bamarni:

bamarni:
Good point, I didn't put validation for this flag as it's a deprecated one so I tried to keep things simple. If users pass something wrong it will just be ignored.

PR #1327, opened 2018-09-06
Fix cluster attach when no cluster is attached

bamarni:
This is about fixing a bug in the `dcos cluster attach` command. When no cluster is attached yet it was failing.

This also adds a non-regression test for it.

https://jira.mesosphere.com/browse/DCOS_OSS-4047


bamarni:

PR #1328, opened 2018-09-07
Mention GNU standards in the CLI style guide

bamarni:
This moves the style guide under the design folder and adds links to GNU standards (based on POSIX), which are the ones we mostly follow in our CLI and plugins.

### DIFF
--- a/design/style.md
+++ b/design/style.md
@@ -1,9 +1,11 @@
 # DC/OS CLI Style Guide
 
 ## Intro
+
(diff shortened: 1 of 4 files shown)


design/style.md
```diff
@@ -22,22 +24,20 @@ The DC/OS CLI is the main tool used by developers and operators to interact with
   * Show - Show a description or definition aka Describe, Get
   * List - List all objects
 
-### Arguments/Flags
+### Arguments/Options
+
+CLI arguments should follow the [GNU standards][1] as much as possible.
```
ArmandGrillet: Inline links would be more coherent with what we currently do in our `.md` files.

bamarni: I added them to the bottom as the line was quite long otherwise.

PR #1330, opened 2018-09-11
Remove temp folder after installing a plugin

bamarni:
Since we switched from renaming to copying the plugin, we should remove the temp plugin as a cleanup procedure.

This became an issue on the DC/OS integration tests suite where the new CLI caused "No space left on device" errors.

https://jira.mesosphere.com/browse/DCOS-41483


bamarni:

PR #1331, opened 2018-09-12
Remove downloaded plugins after installation

bamarni:
Currently we only remove the temporary staging dir, we should also remove ZIP artifacts from the temp folder.


bamarni:

PR #1335, opened 2018-09-14
Support DCOS_ACS_TOKEN when setting-up a cluster

bamarni:
DCOS_ACS_TOKEN is supported in most commands except cluster setup.

This updates the setup command to skip login and use the ACS token from
this env var when it exists. This simplifies other teams scripts as they
sometimes get the ACS token by their own mean and expect the CLI to be
configured with it.

On the other hand we didn't want to add an "--acs-token" flag to the
command as we don't encourage this behaviour.

https://jira.mesosphere.com/browse/DCOS-41803

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -94,14 +94,16 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
 		httpOpts = append(httpOpts, httpclient.TLS(tlsConfig))
 	}
 
-	// Login to get the ACS token.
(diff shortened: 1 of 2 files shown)


bamarni:

takirala:
Can one use this for 1.9 cluster ?

ArmandGrillet:
@takirala No, the new DC/OS CLI is currently compatible only with DC/OS >= 1.10.

takirala:
When would the url for downloading a cli binary that has this fix be ready ?

bamarni:
@takirala : it requires an approval to be merged, once it is it takes a few minutes to get uploaded to S3.

PR #1337, opened 2018-09-17
Persist the token when doing a setup with DCOS_ACS_TOKEN

bamarni:
This is a follow-up from e830f8f98790b5807b53e187c96a690138d0a5d7.

It fixes the setup command to properly persist the ACS token when doing a setup with the DCOS_ACS_TOKEN env var. It updates the integration test accordingly.


bamarni:

bamarni:
@takirala FYI

PR #1338, opened 2018-09-17
Support browser-prompt-authtoken login client method

bamarni:
It was mistaken with browser-prompt-oidcidtoken-get-authtoken,
which requires the user to continue the flow in the browser
and then paste a login token to the terminal, as opposed to
an ACS token.

This commit clarifies the login design overview by adding that 5th login
client method and updates the code accordingly.

This was discovered while working on DCOS-41848.

https://jira.mesosphere.com/browse/DCOS_OSS-4134

### DIFF
--- a/design/login.md
+++ b/design/login.md
@@ -42,7 +42,7 @@ reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
 {
   "dcos-oidc-auth0": {
     "authentication-type": "oidc-implicit-flow",
-    "client-method": "browser-prompt-authtoken",
+    "client-method": "browser-prompt-oidcidtoken-get-authtoken",
     "config": {
       "start_flow_url": "/login?redirect_uri=urn:ietf:wg:oauth:2.0:oob"
     },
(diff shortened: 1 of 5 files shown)


bamarni:

design/login.md
```diff
@@ -42,7 +42,7 @@ reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
 {
   "dcos-oidc-auth0": {
     "authentication-type": "oidc-implicit-flow",
-    "client-method": "browser-prompt-authtoken",
+    "client-method": "browser-prompt-oidcidtoken-get-authtoken",
```
bamarni: This client method exists in our old Python CLI : https://github.com/dcos/dcos-core-cli/blob/a550652477175f48708a361fd0d65e21d243cded/python/lib/dcoscli/dcoscli/auth/main.py#L177

pkg/login/flow.go
```diff
@@ -119,45 +120,64 @@ func (f *Flow) selectProvider(providers Providers) (*Provider, error) {
...
 
+			// With methodBrowserAuthToken, the user is passing the ACS token from the browser
+			// to the terminal directly. Send a HEAD request with an appropriate Authorization
+			// header to a well-known path in order to verify the token.
+			var resp *http.Response
+			resp, err = f.client.sniffAuth(token)
+			if err != nil {
+				return "", err
+			}
+			if resp.StatusCode == 401 {
```
bamarni: Here I'm trying the token and fail on an explicit 401, the old CLI is checking on a 200 or 403 code (https://github.com/dcos/dcos-core-cli/blob/master/python/lib/dcos/dcos/auth.py#L295), is that preferable?

design/login.md
```diff
@@ -94,14 +94,21 @@ Once the login provider is selected, its relevant credentials are read from comm
...
-    in `start_flow_url`. The user is then expected to continue the flow in the browser, eventually
-    they are redirected to a page with a token to copy-paste from the browser to their terminal.
-    POST this token to `/acs/api/v1/auth/login`.
+- **browser-prompt-authtoken** : Open the browser at the page referenced in `start_flow_url`. The user
+    is then expected to continue the flow in the browser, eventually they are redirected to a page with
+    a login token to copy-paste from the browser to their terminal. POST this token to
+    `/acs/api/v1/auth/login`.
+- **browser-prompt-oidcidtoken-get-authtoken** : Open the browser at the page referenced in
+    `start_flow_url`. The user is then expected to continue the flow in the browser, eventually
+    they are redirected to a page with the authentication token to copy-paste from the browser
```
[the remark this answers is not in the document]
bamarni: good point, I'll make a follow-up PR

pkg/login/flow.go
janisz: I think we should add condition for other codes as well. What if we got 5xx

PR #1340, opened 2018-09-18
Pass cluster specific environment variables to plugins 

bamarni:
Follow-up PR from the plugin design update: https://github.com/dcos/dcos-cli/pull/1341

When a plugin is invoked with a cluster currently attached, a given set
of environment variables are passed.

- `DCOS_URL`: The base URL of the DC/OS cluster without a trailing slash
(eg. https://dcos.example.com).

- `DCOS_ACS_TOKEN`: The authentication token for the current user.

- `DCOS_TLS_INSECURE`: Indicates whether the CLI is configured with an
insecure TLS setup. This is set to 1 when the DCOS_URL scheme is http://
or core.ssl_verify is set to false.

- `DCOS_TLS_CA_PATH`: Unless DCOS_TLS_INSECURE=1, this env var indicates the
path to the CA bundle to verify server certificates against. When it is
not set, the system's CA bundle must be used.

https://jira.mesosphere.com/browse/DCOS_OSS-4101

### DIFF
--- a/pkg/cmd/dcos.go
+++ b/pkg/cmd/dcos.go
@@ -8,12 +8,13 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/dcos/dcos-cli/api"
 	"github.com/dcos/dcos-cli/pkg/cmd/auth"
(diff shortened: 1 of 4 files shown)


bamarni:

bamarni:

PR #1341, opened 2018-09-19
Add a plugin design doc

bamarni:
This adds the plugin design doc from:

https://docs.google.com/document/d/1UN3rPhmtdpy2dz1jn3XobTJlOxQtXujbW3ExvBkmi14/edit

It removes some background section to fit more into the style of the
design pages in this repo, and updates some sections to reflect some
recent changes that were made.

The second commit is a design update to the plugin system in order to pass
additionnal env variables to plugins whenever a cluster is configured.

The main goal being to make it easier for plugins to interact with
DC/OS, as they currently have to read the config TOML file by themselves
or invoke back the `dcos` binary.

### DIFF
--- a/design/plugin.md
+++ b/design/plugin.md
@@ -21,8 +21,7 @@ In its normalized form, an unpacked plugin is a directory with the following str
 An unpacked plugin has an env directory containing a bin directory which contains all the executables in
 the plugin and a plugin.toml file which defines information about the plugin.
 
-The bin folder is just a convention but in the next section we’ll see that it can be changed to something
-else.
+The bin directory is just a convention, it can be changed to something else as we'll see in the next section.
 
 A plugin may contain multiple executables, each of which can hold multiple top level commands. Which
 executables hold which top level commands is defined in the plugin.toml file. The file contains command
--- a/design/plugin.md
+++ b/design/plugin.md
@@ -108,6 +108,16 @@ For example, if a `dcos-hello` executable in a plugin has a `hello` top level co
 
 Executables contained within plugins are called synchronously from the CLI by spawning a new child process. The CLI waits for the child process to complete. The CLI makes available its own executable path to the child via an ENV variable `DCOS_CLI_EXECUTABLE_PATH`.
 
+When a cluster is attached, the CLI will also pass the following ENV variables:
+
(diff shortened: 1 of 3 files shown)


design/plugin.md
```diff
@@ -0,0 +1,131 @@
...
+
+This command lists installed plugins, with their name, version, description and the subcommands they provide. It also accepts a --json flag.
+
+## Calling a plugin
+
+When a plugin is called it will have the top level command name as its second argument regardless of whether the plugin only has one executable.
+
+For example, if a `dcos-hello` executable in a plugin has a `hello` top level command, it would be called like `env/bin/dcos-hello hello [...]`.
+
+Executables contained within plugins are called synchronously from the CLI by spawning a new child process. The CLI waits for the child process to complete. The CLI makes available its own executable path to the child via an ENV variable `DCOS_CLI_EXECUTABLE_PATH`.
```
br-lewis: The description of `DCOS_CLI_EXECUTABLE_PATH` probably belongs in the list below.

design/setup.md
```diff
@@ -102,3 +103,29 @@ and the CLI is attached to it. A cluster config looks like this:
...
+
+These plugins download URLs are retrieved through Cosmos, where they are registered as packages.
+
+### dcos-core-cli
+
+The [core plugin](https://github.com/dcos/dcos-core-cli) contains subcommands such as marathon, job, node,
+package, service, task.
+
+When the core plugin can't be installed (eg. insufficient Cosmos permission or air-gapped environment),
+it then fallbacks to installing it from the DC/OS CLI binary itself, which bundles a core plugin.
```
br-lewis: "fallbacks" -> "falls back".

We should make clear that which version of the core CLI it has bundled will be dependent on the version of the wrapper CLI you happen to be using and that it won't check the version of the core it has against the DC/OS cluster.

design/plugin.md
bamarni: The list below is for env variables which are only set when a cluster is attached.

bamarni:
I've addressed the comments, I mainly wanted to add the second commit here as it's needed for https://jira.mesosphere.com/browse/DCOS_OSS-4101 however we didn't have the design for the current system yet.

design/plugin.md
br-lewis: Ah I see now.

PR #1342, opened 2018-09-19
Plugin completion

br-lewis:
This add the necessary changes to the Go code and completion shell script to support plugin completion scripts.

It also has some whitespace and shellcheck tweaks that don't impact functionality


pkg/cmd/plugin/plugin_list.go
```diff
@@ -28,6 +31,18 @@ func newCmdPluginList(ctx api.Context) *cobra.Command {
...
+			} else if commands {
+				for _, plugin := range plugins {
+					for _, command := range plugin.Commands {
+						fmt.Fprintf(ctx.Out(), "%s ", command.Name)
+					}
+				}
+				return nil
+			} else if completionDirs {
+				for _, plugin := range plugins {
+					fmt.Fprintf(ctx.Out(), "%s ", plugin.CompletionDir())
```
bamarni: Any reason for using spaces instead of newlines? It seems more common for quiet outputs.

bamarni: `var=($(dcos plugin list --completion-dirs))` should work with newlines

br-lewis: I did it this way because shellcheck complains about that syntax. There's more info here
https://github.com/koalaman/shellcheck/wiki/SC2207
and I wouldn't be against fixing it but then we do start to run into problems of doing things in a bash3 or bash4 way.

Though really I haven't been too worried about that so far and I've been running bash4...

bamarni: ok, spaces are fine here as it's a hidden/internal command, however we will most likely have newlines in other commands (eg. dynamic autocompletion with `dcos marathon app list -q`).

pkg/plugin/plugin.go
```diff
@@ -14,3 +17,13 @@ type Command struct {
...
 }
+
+// Dir gives the path to the plugin's env directory on the filesystem.
+func (p *Plugin) Dir() string {
+	return p.dir
+}
+
+// CompletionDir returns the absolute path to the directory holding the plugin's completion script files
+func (p *Plugin) CompletionDir() string {
+	return filepath.Join(p.Dir(), "completion")
```
bamarni: The completion command requires `bash` as an argument, this gives us the flexibility to support other shell completion in the future. Can we reflect this here too? If we were to support zsh in the future for example where would we put the scripts.

pkg/cmd/plugin/plugin_list.go
br-lewis: This is a good point. How do we want to handle differences in bash3 vs bash4? I'm not even sure what all of them are but at least for this case maybe I can make a function that'll take in a command and return a list and have that handle the different methods based on version.

pkg/plugin/plugin.go
br-lewis: Should we go with `completion/<shell>` or something more like `completion.<shell>/`?

bamarni: I like `completion/<shell>` :+1:

pkg/cmd/plugin/plugin_list.go
bamarni: > How do we want to handle differences in bash3 vs bash4?

They say `# For bash 3.x+`, so the code with IFS should also work on bash 4 I assume?

Otherwise, I don't think there is anything intrinsically wrong in using the `var=()` syntax, from what they say : 

> You are doing unquoted command expansion in an array. This will invoke the shell's sloppy word splitting and glob expansion.

If a command output has a well-known limited set of characters (eg. a list of UUIDs), this will never cause the problem they mention.

pkg/plugin/plugin.go
br-lewis: This was done in the script instead since it didn't make much sense to add some kind of shell flag to `dcos plugin list --completion-dirs` and each shell will know which shell it is.

PR #1343, opened 2018-09-24
Rename DCOS_ACS_TOKEN to DCOS_CLUSTER_SETUP_ACS_TOKEN

bamarni:
DCOS_ACS_TOKEN is already used and planned to be used for passing
information to plugins. If a plugin does a cluster setup it would then
cause issues as the flow would try to use the current cluster's ACS
token (due to the default child process env var inheritance).

It makes sense to reduce the scope of this env var to the cluster setup
command. It remains an undocumented env var for internal usage at
Mesosphere.

https://jira.mesosphere.com/browse/DCOS_OSS-4176


bamarni:

PR #1344, opened 2018-09-27
Make browser-prompt_authtoken check for 200/403 status codes

bamarni:
Follow-up from https://github.com/dcos/dcos-cli/pull/1338

This changes the browser-prompt_authtoken client method to expect a 200
or 403 status code in case of a valid token instead of accepting
anything else than a 401. Otherwise it should also have been checking
for 50X status codes.

This is the same behavior as the Python CLI:

https://github.com/dcos/dcos-core-cli/blob/8a51a0d05ad3ece50d1041bede5a7638f3ed13a6/python/lib/dcos/dcos/auth.py#L295

### DIFF
--- a/pkg/login/flow.go
+++ b/pkg/login/flow.go
@@ -160,18 +160,21 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
 				break
 			}
 
-			// With methodBrowserAuthToken, the user is passing the ACS token from the browser
-			// to the terminal directly. Send a HEAD request with an appropriate Authorization
-			// header to a well-known path in order to verify the token.
+			// With methodBrowserAuthToken, the user is passing the authentication token from the browser
+			// to the terminal directly. Send a HEAD request with an appropriate Authorization header to a
+			// well-known path in order to verify the authentication token.
 			var resp *http.Response
 			resp, err = f.client.sniffAuth(token)
 			if err != nil {
 				return "", err
 			}
-			if resp.StatusCode == 401 {
-				err = errors.New("invalid auth token")
-			} else {
+			switch resp.StatusCode {
+			case 200, 403:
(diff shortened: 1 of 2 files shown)


pkg/login/flow.go
```diff
@@ -168,10 +168,13 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
 			if err != nil {
 				return "", err
 			}
-			if resp.StatusCode == 401 {
-				err = errors.New("invalid auth token")
-			} else {
+			switch resp.StatusCode {
+			case 200, 403:
```
jgehrcke: When do we expect a 403 emitted by whom?

What is `acsToken`? Is it the DC/OS authentication token?

The DC/OS authentication token is never emitted within a 403 response.

bamarni: > When do we expect a 403 emitted by whom?
I do not know, this is what we currently have in the Python code. The resource we use to verify the auth token is the same that we use to challenge auth on OSS clusters (`/pkgpanda/active.buildinfo.full.json`).

This is the `browser-prompt-authtoken` so we already have the token even though a 403 response is returned.

pkg/login/flow.go
```diff
@@ -168,10 +168,13 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
...
 				acsToken = token
+			case 401:
+				err = errors.New("invalid auth token")
```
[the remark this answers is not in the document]
bamarni: This code path here is only about the `browser-prompt-authtoken` method.

pkg/login/flow.go
bamarni: I've update the comment to use `authentication token` instead of `ACS token`

bamarni:

PR #1345, opened 2018-09-27
Introduce the concept of canonical cluster URL

bamarni:
This updates the "setup" design page to introduce the concept of
canonical cluster URL instead of trying to detect the cluster security
mode.

This proposal is meant to tackle the following error path:

https://jira.mesosphere.com/browse/DCOS-22237

This is also currently problematic with our soak clusters redirections.

### DIFF
--- a/design/setup.md
+++ b/design/setup.md
@@ -40,25 +40,24 @@ HTTPS URL:
 The setup command also accepts all the flags from the `dcos auth login` command as it triggers a login
 flow during the setup.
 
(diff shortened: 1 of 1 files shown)


bamarni:
@jongiddy : thanks for the feedback, I've put these codes as this is the default behavior of the go httpclient, this can be customized, though it'd require looping and making all requests manually instead of delegating it to the stdlib. If there is nothing obviously wrong in our case for following 303, I'd rather follow these too.

PR #1347, opened 2018-09-28
Create a fake core plugin for the CLI to use to populate the help list with core cli commands

br-lewis:
This uses the bundled core plugin and afero's MemMapFs to create a temporary fake core plugin which is used to populate the help menu with the core cli commands when not attached to the cluster.

This does have some oddities.
* If a user has a cluster attached but no Core CLI, the commands won't be shown.
* Autocompletion may sort of work to complete the top level command name but nothing under it.
* If one of those commands is called when no cluster is attached, it'll print `Error: no cluster attached`. This also applies to any plugin command as plugins can't be global at the moment.

Besides updating the reference to the new package, this doesn't touch how we do normal bundled installation.


ArmandGrillet:
I'm OK with the three oddities described in the PR.

Makefile
```diff
@@ -30,6 +30,10 @@ core-download:
 	wget https://downloads.dcos.io/cli/plugins/dcos-core-cli/1.12/darwin/x86-64/dcos-core-cli.zip -O build/darwin/core.zip
 	wget https://downloads.dcos.io/cli/plugins/dcos-core-cli/1.12/windows/x86-64/dcos-core-cli.zip -O build/windows/core.zip
 
+.PHONY: install
+install: generate
```
bamarni: For consistency with the "go build" target, can we make this one not depend on "generate" and accept the build tags from the env var too?

pkg/corecli/extract.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"github.com/dcos/dcos-cli/pkg/fsutil"
+	"github.com/dcos/dcos-cli/pkg/plugin"
+	"github.com/pelletier/go-toml"
+	"github.com/spf13/afero"
+	"path/filepath"
+)
+
+// TempPlugin will read the bundled corecli into a memmapfs and return a fake plugin from that data.
+func TempPlugin() (*plugin.Plugin, error) {
```
bamarni: Can we leave this hidden as an implementation detail instead of exposing a new package? See https://golang.org/doc/go1.4#internalpackages

pkg/corecli/extract.go
```diff
@@ -0,0 +1,62 @@
...
+	pluginData, err := Asset("core.zip")
+	if err != nil {
+		return "", err
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	// Write out the data into a temp directory so that it's in the real filesystem for buildPlugin
```
bamarni: Is it in the real filesystem or in the mem map fs? More generally does anything gets written to the real filesystem with these changes?

br-lewis: I copied this from the code in `setup` and forgot to remove this comment. It doesn't reach into the real filesystem (though this function would if passed an Fs that does).

Makefile
```diff
@@ -19,9 +19,9 @@ darwin linux windows: docker-image
 
 .PHONY: core-bundle
 core-bundle: docker-image
-	$(call inDocker,go-bindata -pkg setup -o pkg/setup/corecli_linux.gen.go -nometadata -tags "corecli" -prefix "build/linux" build/linux/core.zip)
-	$(call inDocker,go-bindata -pkg setup -o pkg/setup/corecli_darwin.gen.go -nometadata -tags "corecli" -prefix "build/darwin" build/darwin/core.zip)
-	$(call inDocker,go-bindata -pkg setup -o pkg/setup/corecli_windows.gen.go -nometadata -tags "corecli" -prefix "build/windows" build/windows/core.zip)
+	$(call inDocker,go-bindata -pkg corecli -o pkg/corecli/corecli_linux.gen.go -nometadata -tags "corecli" -prefix "build/linux" build/linux/core.zip)
+	$(call inDocker,go-bindata -pkg corecli -o pkg/corecli/corecli_darwin.gen.go -nometadata -tags "corecli" -prefix "build/darwin" build/darwin/core.zip)
+	$(call inDocker,go-bindata -pkg corecli -o pkg/corecli/corecli_windows.gen.go -nometadata -tags "corecli" -prefix "build/windows" build/windows/core.zip)
```
bamarni: should this be changed with the internal folder?

bamarni:

ArmandGrillet:
`test_dcos_help` need to be updated.

bamarni:

PR #1348, opened 2018-10-01
Added handling of redirects when setting up a cluster.

rgo3:
With this commit we add the detection of the canonical cluster URL by
following redirects. For more information see: #1345.

https://jira.mesosphere.com/browse/DCOS_OSS-4099


pkg/setup/setup.go
```diff
@@ -153,6 +162,25 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
...
+func (s *Setup) detectCanonicalClusterURL(clusterURL string, httpOpts []httpclient.Option) (string, error) {
+	httpClient := httpclient.New(clusterURL, httpOpts...)
+	req, err := httpClient.NewRequest("HEAD", "/", nil)
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return "", err
```
bamarni: Maybe we could make the error message more informative here, as it's the first one users should see in case of an unreachable cluster for example.

pkg/setup/setup.go
```diff
@@ -85,7 +85,16 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
...
 		httpclient.Logger(s.logger),
-		httpclient.NoFollow(),
+	}
+
+	// Make sure we follow redirects.
+	canonicalClusterURL, err := s.detectCanonicalClusterURL(cluster.URL(), httpOpts)
+	if err != nil {
+		return nil, err
+	}
+	if !strings.Contains(canonicalClusterURL, cluster.URL()) {
```
bamarni: Why using contains here and not a `!=`?

rgo3: Because I return `resp.Request.URL.String()` which will return the domain with an `/` at the end which often is not included by user-input. the `.Contains` function will cover both cases.

rgo3:

rgo3:

cmd/dcos/main.go
```diff
@@ -66,14 +66,14 @@ func logrusLevel(errout io.Writer, verbosity int, logLevel string) logrus.Level
...
+	case "info":
 		fmt.Fprintln(errout, "The --log-level flag is deprecated. Please use the -v flag.")
 		return logrus.InfoLevel
-	case "error", "critical":
+	case "error", "critical", "warning":
 		fmt.Fprintf(errout, "The --log-level=%s flag is deprecated. It is enabled by default.\n", logLevel)
 	}
-	// Without the verbose flag, default to error level.
-	return logrus.ErrorLevel
+	// Without the verbose flag, default to warning level.
```
ArmandGrillet: This should not be part of the PR or at least not the same commit.

PR #1352, opened 2018-10-04
Experimental auto-installation of plugins / Auto-extraction of the bundle core CLI

bamarni:
## Auto-install plugins based on an experimental env var

This impletements the behavior designed in the following commit: cdfb740

It removes the `--no-plugin` option if favor of an experimental env var called `DCOS_CLI_EXPERIMENTAL_AUTOINSTALL_PLUGINS`.

When this var is set, the CLI will autoinstall the default plugins (core and, if applicable, enterprise).

As it's an experimental feature we're not trying to handle errors gracefully, any error in that step will lead to a non-zero exit code (even if it's only related to the enterprise plugin).

https://jira.mesosphere.com/browse/DCOS_OSS-4186

## Auto-extract the bundled core CLI

This implements what has been agreed in [design/corecli.md](https://github.com/dcos/dcos-cli/blob/master/design/corecli.md).

The first time a user invokes a core command, an informational message is printed and the core CLI bundle gets auto-extracted.

https://jira.mesosphere.com/browse/DCOS_OSS-4187


pkg/cmd/dcos.go
```diff
@@ -175,6 +190,18 @@ func newPluginCommand(ctx api.Context, cmd plugin.Command) *cobra.Command {
 	}
 }
 
+// extractCorePlugin extracts the bundled core plugin into the plugins folder.
+func extractCorePlugin(ctx api.Context, cluster *config.Cluster) (*plugin.Plugin, error) {
+	ctx.Logger().Warn("Extracting “dcos-core-cli“...")
```
ArmandGrillet: The first quotes seem wrong.

pkg/internal/corecli/corecli.go
```diff
@@ -59,3 +57,16 @@ func extractPlugin(fs afero.Fs) (string, error) {
 
 	return pluginFile.Name(), nil
 }
+
+// InstallPlugin installs the core plugin bundled with the CLI.
```
ArmandGrillet: The scope of the function doesn't seem limited to the core plugin, is it?

tests/integration/test_corecli.py
```diff
@@ -0,0 +1,44 @@
+import json
+import os
+
+from .common import exec_cmd, default_cluster, default_cluster_with_plugins  # noqa: F401
+
+import pytest
+
+
+@pytest.mark.skipif(os.environ.get('DCOS_TEST_CORECLI') is None, reason="no core CLI bundle")
```
ArmandGrillet: Not `DCOS_CLI_TEST_CORECLI`? `CLI` would then be redundant but the variable would have a name aligned with `DCOS_CLI_EXPERIMENTAL_AUTOINSTALL_PLUGINS`.

pkg/cmd/dcos.go
bamarni: I used the one from here : https://github.com/dcos/dcos-cli/blob/master/pkg/setup/setup.go#L310

However we use normal quotes a bit below : https://github.com/dcos/dcos-cli/blob/master/pkg/setup/setup.go#L335

I don't have a strong opinion on using the one or the other, whatever we decide should be added to the style guide, it's not directly related to this PR though.

pkg/internal/corecli/corecli.go
bamarni: You mean the name `InstallPlugin`? The package name `corecli` should already indicate that functions in it are related to the core plugin.

tests/integration/test_corecli.py
bamarni: This is consistent with https://github.com/dcos/dcos-cli/blob/master/tests/launch_cluster.py#L56-L59. That being said I wouldn't mind changing these to add `CLI`, but it should be done separately.

pkg/cmd/dcos.go
ArmandGrillet: Please use the one from https://github.com/dcos/dcos-cli/blob/master/pkg/setup/setup.go#L335 and fix https://github.com/dcos/dcos-cli/blob/master/pkg/setup/setup.go#L310 in this PR or another to follow US standards 
quotation marks over the world

bamarni: I've updated it to use normal quotes on this line only, if you want to standardize quoting in our style guide please follow-up with a separate jira as it's not directly related to this PR.

ArmandGrillet: Sure, thanks!

ArmandGrillet:
@bamarni The commits seem good for a rebase and merge, what do you think?

bamarni:
Let's run the tests and wait until tomorrow at least to see if there are more reviews

bamarni:

PR #1358, opened 2018-10-10
Disable cgo

bamarni:
Because we're cross-compiling from a Linux jenkins builder and cgo is
disabled when cross-compiling, our Darwin and Windows CLIs don't rely on
cgo while the Linux one might.

Explicitly passing CGO_ENABLED=0 makes sure it is also disabled on
Linux.


bamarni:

PR #1359, opened 2018-10-10
Skip certificate validation when using --insecure

bamarni:
When using the `--insecure` option, the CLI should not verify server certificates. This should also be the case when the certificate is for another host than the one in use.

https://jira.mesosphere.com/browse/DCOS_OSS-4292


bamarni:

bamarni:

Issue #1351, opened 2018-10-04
dcos cluster setup fails for clusters without authentication


jerith:
### What version of DC/OS + DC/OS CLI are you using (`dcos --version`)?

```
dcoscli.version=20f80a50d62367f12bb6542eefa10eb5ef51da04
dcos.version=1.12.0-beta1
dcos.commit=a55cf6cd18bea6961e2cc3c957ac8d0ee47583d5
dcos.bootstrap-id=f6e3cce0ac2024518b3dfac5879423ef75db79ac
```

### What operating system and version are you using?

Debian 8.6 (jessie)

### What did you do?

`dcos cluster setup http://10.0.91.11` to configure the CLI for a 1.12.0-beta1 test cluster.

### What did you expect to see?

Successful CLI configuration for the given cluster.

### What did you see instead?

CLI configuration failed with `Error: expected status code 401, got 200`

When run with `-v`:
```
$ dcos -v cluster setup http://10.0.91.11
Setting up the cluster...
Falling back to the WWW-Authenticate challenge.
Error: expected status code 401, got 200
```

I think the problem is in this code: https://github.com/dcos/dcos-cli/blob/20f80a50d62367f12bb6542eefa10eb5ef51da04/pkg/setup/setup.go#L110-L119

If there is no ACS token, the CLI tries to login to get one. Because the cluster has no authentication configured, it doesn't return the 401 auth challenge response that the CLI expects.

As a workaround, I provided an arbitrary ACS token to skip the login and was able to configure my cluster:
```
$ DCOS_CLUSTER_SETUP_ACS_TOKEN=blah dcos cluster setup http://10.0.91.11
$ dcos cluster list
        NAME                      ID                    STATUS      VERSION            URL
  *  seed-stack  24d8583c-38ba-4091-adb7-fe11a3b4d3db  AVAILABLE  1.12.0-beta1  http://10.0.91.11
```

jgehrcke:
Thank you for the report!

> Because the cluster has no authentication configured

Did you use the `oauth_enabled` configuration parameter for this?

jgehrcke:
Thanks for the quick reply. Sorry for the inconvenience! 

>  (We don't really want to juggle oauth stuff for test clusters that we rebuild regularly and that aren't accessible from the public internet anyway.)

Makes sense.

I have created [a JIRA ticket](https://jira.mesosphere.com/browse/DCOS_OSS-4235) which tracks patching the CLI to address the issue you have observed.

[...]

PR #1365, opened 2018-10-17
Fix handling for clusters with `oauth_enabled: false`

br-lewis:
This makes it so that `dcos cluster setup` of a cluster without authentication will not fail. `dcos auth login` and `dcos auth list-providers` will fail however but considering that a cluster without auth has no use for either of these, that seems like an acceptable tradeoff.

### DIFF
--- a/design/login.md
+++ b/design/login.md
@@ -51,9 +51,9 @@ reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
 }
 ```
 
-If the `Head` request returns a 200 response, the cluster has no authorization. Setup will succeed and
-the cluster will be usable but any calls to `dcos auth login` or `dcos auth list-providers` will result
-in an error.
+If the `Head` request returns a 200 response, authentication is disabled on this cluster. Setup will
+succeed and the cluster will be usable but any calls to `dcos auth login` or
+`dcos auth list-providers` will result in an error.
 
 ## Log in to a DC/OS cluster
 
--- a/design/login.md
+++ b/design/login.md
@@ -51,6 +51,10 @@ reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
 }
 ```
 
+If the `Head` request returns a 200 response, the cluster has no authorization. Setup will succeed and
+the cluster will be usable but any calls to `dcos auth login` or `dcos auth list-providers` will result
+in an error.
+
 ## Log in to a DC/OS cluster
 
 "Login" is the process of presenting some credentials to a DC/OS cluster within an HTTP request which is then,
--- a/pkg/login/login.go
+++ b/pkg/login/login.go
@@ -13,6 +13,8 @@ import (
 )
 
 var (
+	// ErrAuthDisabled is the error returned when attempting to get authentication providers
+	// from a cluster without authentication.
 	ErrAuthDisabled = errors.New("authentication disabled")
 )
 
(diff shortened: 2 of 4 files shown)


br-lewis:
Addresses https://github.com/dcos/dcos-cli/issues/1351

pkg/setup/setup.go
```diff
@@ -124,7 +124,7 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
 		httpClient := httpclient.New(cluster.URL(), httpOpts...)
 		var err error
 		acsToken, err = s.loginFlow.Start(flags.loginFlags, httpClient)
-		if err != nil {
+		if err != nil && err != login.ErrAuthDisabled {
```
bamarni: Should we warn when authentication is disabled? Is that considered insecure?

br-lewis:
@mesosphere-ci run all integration tests

design/login.md
```diff
@@ -51,6 +51,10 @@ reading the `WWW-Authenticate` header DC/OS returns in the 401 response:
 }
 ```
 
+If the `Head` request returns a 200 response, the cluster has no authorization. Setup will succeed and
```
jgehrcke: "the cluster does not require authentication"  would be much more correct than "the cluster has no authorization"

pkg/setup/setup.go
```diff
@@ -124,7 +124,9 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
 		httpClient := httpclient.New(cluster.URL(), httpOpts...)
 		var err error
 		acsToken, err = s.loginFlow.Start(flags.loginFlags, httpClient)
-		if err != nil {
+		if err == login.ErrAuthDisabled {
+			s.logger.Warn("This cluster does not have authorization enabled. Skipping.")
```
bamarni: > This cluster does not have authentication enabled. Skipping login.

pkg/login/login.go
```diff
@@ -86,7 +93,9 @@ func (c *Client) challengeAuth() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if resp.StatusCode != 401 {
+	if resp.StatusCode == 200 {
+		return "", ErrAuthDisabled
+	} else if resp.StatusCode != 401 {
```
bamarni: This is a very small nit, but why using else when the previous statement returns? As the previous `if err != nil` condition returns too, it'd be more consistent to write : 
``` go
if err != nil {
    return "", err
} else if resp.StatusCode == 200 {
    return "", ErrAuthDisabled
} else if resp.StatusCode != 401 {
    return "", fmt.Errorf("expected status code 401, got %d", resp.StatusCode)
}
```

I think it's clearer to avoid "else / else if" altogether when the previous block returns, though the benefit is more visible for else blocks (as there is then no extra indentation level).

bamarni:
It'd be good to also have a unit test for this.

pkg/setup/setup.go
jgehrcke: Thanks, @bamarni. It's important to distinguish authentication and authorization. I still however prefer something in the lines of "This cluster does not require authenticated requests".

bdonnahue:
So what version is this fix going into / when will it be available?

br-lewis:
It's part of 0.7 which was released yesterday with DC/OS 1.12.

PR #1366, opened 2018-10-19
Build the CLI with 1.12-patch.0

bamarni:
The version in the Makefile is set to 1.12-patch.0 and it can be configured through a CORE_VERSION env var.

https://jira.mesosphere.com/browse/DCOS-40541


bamarni:

PR #1367, opened 2018-10-19
Add test and make small cosmetic changes to no auth fix

br-lewis:
This adds a test for and addresses a couple cosmetic issues from https://github.com/dcos/dcos-cli/pull/1365

### DIFF
--- a/pkg/login/login.go
+++ b/pkg/login/login.go
@@ -92,12 +92,16 @@ func (c *Client) challengeAuth() (string, error) {
 	resp, err := c.sniffAuth("")
 	if err != nil {
 		return "", err
-	} else if resp.StatusCode == 200 {
+	}
+
+	switch resp.StatusCode {
+	case 200:
 		return "", ErrAuthDisabled
-	} else if resp.StatusCode != 401 {
+	case 401:
+		return resp.Header.Get("WWW-Authenticate"), nil
+	default:
 		return "", fmt.Errorf("expected status code 401, got %d", resp.StatusCode)
 	}
-	return resp.Header.Get("WWW-Authenticate"), nil
 }
 
 // sniffAuth sends an HTTP request with a given ACS token to a well-known resource.
--- a/pkg/login/login.go
+++ b/pkg/login/login.go
@@ -92,8 +92,7 @@ func (c *Client) challengeAuth() (string, error) {
 	resp, err := c.sniffAuth("")
 	if err != nil {
 		return "", err
-	}
-	if resp.StatusCode == 200 {
+	} else if resp.StatusCode == 200 {
 		return "", ErrAuthDisabled
 	} else if resp.StatusCode != 401 {
 		return "", fmt.Errorf("expected status code 401, got %d", resp.StatusCode)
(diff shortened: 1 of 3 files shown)


pkg/login/login_test.go
```diff
@@ -57,6 +57,13 @@ func TestProviders(t *testing.T) {
 				w.Header().Add("WWW-Authenticate", fixture.authChallenge)
 				w.WriteHeader(401)
 			})
+		} else {
+			mux.HandleFunc("/pkgpanda/active.buildinfo.full.json", func(w http.ResponseWriter, req *http.Request) {
+				assert.Equal(t, "HEAD", req.Method)
+				assert.Equal(t, "", req.Header.Get("Authorization"))
+
+				w.WriteHeader(200)
+			})
```
bamarni: It would make sense to specifically test for a `ErrAuthDisabled` error, otherwise the test would succeed even without your fix. Maybe that'd be simpler to have a separate test function for this case.

pkg/setup/setup.go
```diff
@@ -125,7 +125,7 @@ func (s *Setup) Configure(flags *Flags, clusterURL string, attach bool) (*config
 		var err error
 		acsToken, err = s.loginFlow.Start(flags.loginFlags, httpClient)
 		if err == login.ErrAuthDisabled {
-			s.logger.Warn("This cluster does not have authorization enabled. Skipping.")
+			s.logger.Warn("This cluster does not have authorization enabled. Skipping login.")
```
bamarni: `authorization` > `authentication`

pkg/login/login.go
```diff
@@ -92,8 +92,7 @@ func (c *Client) challengeAuth() (string, error) {
 	resp, err := c.sniffAuth("")
 	if err != nil {
 		return "", err
-	}
-	if resp.StatusCode == 200 {
+	} else if resp.StatusCode == 200 {
```
bamarni: Do you find it more readable like this? I meant it the other way around, or even using a switch statement. See here for example how it's done to conditionally return based on the status code : https://github.com/dcos/dcos-core-cli/blob/1cbe3e815a54a7e1d43d8236690c504ab9765c55/pkg/metronome/client.go#L80

But anyways, that's fine as is too, not a big deal.

pkg/login/login_test.go
br-lewis: I originally did that but figured it might make some sense to integrate into that function. If I back out those changes though what is https://github.com/dcos/dcos-cli/pull/1367/files#diff-3abfe1b27be64ad5227dbe318509dedeL22 supposed to be checking?

Would it be better to add another check
```go
if fixture.authChallenge == "" {
    require.Error(t, err, ErrAuthDisabled)
}
```?

bamarni: https://github.com/dcos/dcos-cli/pull/1367/files#diff-3abfe1b27be64ad5227dbe318509dedeL22 would be like testing against a non DC/OS cluster.

If you think that makes sense as part of this function that's fine too.

pkg/setup/setup.go
bamarni: A comment from JP on the other PR : 
> I still however prefer something in the lines of "This cluster does not require authenticated requests".

PR #1374, opened 2018-10-31
[design] Plugin auto-install proposal

bamarni:
The goal of this design proposal is to improve the dcos-core-cli and
dcos-enterprise-cli auto-installation in order to be able to remove
the core CLI bundling and make the auto-installation a non-experimental
feature.

It relies on a fallback mechanism to support most of the cases,
before eventually failing with a link to a web-page where the user can
manually get the download URLs.


design/setup.md
```diff
@@ -127,3 +124,23 @@ This is determined through the [DC/OS variant](https://jira.mesosphere.com/brows
...
+- `stability` is set to `testing` when the CLI is a dev build, otherwise defaults to `releases`
+- `platform` can either be `linux`, `darwin`, or `windows`
+- `dcos-version` is the major and minor version of the DC/OS cluster (eg. `1.13`)
+
+Failing to download the plugins from their canonical URL usually means that the CLI user operates from an
+air-gapped environment. In that case the CLI falls back to downloading the plugins through Cosmos, where they are registered as packages named `dcos-core-cli` and `dcos-enterprise-cli`.
+
+Failing to download the plugins from Cosmos usually means that the CLI user doesn't have the necessary
+permissions in order to interact with Cosmos. In that case the setup command fails with an error message
+indicating a web-page the user can go to in order to download the plugins manually.
```
rgo3: Is it sufficient enough to just tell users the location of the plugins? Do they know how to add them?
Otherwise LGTM.

bamarni:
My main concerns thinking about this again are : 

- how to handle updates of these 2 plugins, because they can come from 2 sources, it woyld either be with `dcos package install` or `dcos plugin add` (or possibly adding another command..), yesterday @ArmandGrillet suggested that the way to update them would be to run cluster setup again, which I think is a decent solution.

- how to deal with dc/os dev versions? Eg. At the moment DC/OS 1.13 is in development, we have no 1.13 core plugin for it except the testing one (built on each commit).

bamarni:
Also we had a discussion with @takirala yesterday and it seems like the cosmos permission issue should be very easy to solve, it'd only be about updating adminrouter's configuration to not require permissions on some endpoints (in our case the describe endpoint).

PR #1376, opened 2018-11-07
Add an 'index.html' file to represent the index of all CLI artifacts.

klueska:
This PR adds files to the scripts folder for publishing a list of all CLI artifacts to S3. The artifacts themselves are scraped from `downloads.dcos.io/cli/releases` and published as an `artifacts.json` file that is then read and rendered by a small `html/css/javascript` app.

A live version of the index can be seen here:
https://downloads.dcos.io/cli/index.html

Two different script are used to publish the necessary artifacts for this index.

The first is `publish_index.py` which publishes all of the necessary files required to render the index. This script should not need to be ran very often, and is only necessary if changes are made to any of the files used to implement the rendering of the index.

The second is `publish_artifacts.py` which publishes a JSON file listing all of the artifacts that should be displayed in the index. This script should be ran every time we publish a new artifact to S3 which we want to publish as a new release. This applies to both CLI releases themselves, as well as plugins.

In a subsequent commit we should add the calling of `publish_artifacts.py` to our CI as a step in publishing any of our release artifacts.


bamarni:
I might not be up-to-date here, what is the status of this PR and what are the next steps? eg. how to automate the build of the web page, where will it be documented.

klueska:
I think the script itself is ready to be merged. We need to update the Jenkins tooling to call it, and republish it whenever a new artifact is pushed to S3. We also need to update our documentation on the DC/OS docs to point to it so people can go to it and know where to download things from.

ArmandGrillet:
Added https://jira.mesosphere.com/browse/DCOS_OSS-4537 and https://jira.mesosphere.com/browse/DCOS_OSS-4538 to our sprint in order to track the upcoming work to do. IMO this PR is ready to be rebased and merged.

PR #1377, opened 2018-11-15
Update "requests" dependency

bamarni:
Github is showing a security warning about the current version being used.


bamarni:
I just used the example snippet they had on the warning description, now updated to `==`.

PR #1379, opened 2018-11-15
Handled case when one locally and linked cluster found when attaching.

ArmandGrillet:
The issue before:
```
➜  dcos-cli (master) ✔ build/darwin/dcos cluster list
           NAME                          ID                     STATUS      VERSION                                        URL
  *  ronzalez-fakjz4f  f3a326bc-134d-47ab-bbf0-6260c857e7c6  AVAILABLE    1.12.0      https://34.212.25.125
     ronzalez-gs0b04n  27e76b1f-f63d-4f54-8d21-ae6c38d13a9b  AVAILABLE    1.12.0      https://34.222.140.41
➜  dcos-cli (master) ✔ build/darwin/dcos cluster attach ronzalez-gs0b04n
Error: multiple matches found
```
This was because ronzalez-gs0b04n xas also a cluster linked to ronzalez-fakjz4f.

### DIFF
--- a/pkg/cmd/cluster/cluster_attach.go
+++ b/pkg/cmd/cluster/cluster_attach.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"path/filepath"
 	"strings"
 
 	"github.com/dcos/dcos-cli/api"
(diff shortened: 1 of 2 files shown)


bamarni:
Could you add a unit test to prevent regression?

ArmandGrillet:
As the code is not in a function and the use case quite specific, I think we should rather have an integration test for this. There are currently no integration tests for `cluster attach` and I think this should be part of a future PR.

bamarni:
We only have a single cluster during integration test so it wouldn't be possible to have an integration test for that. There is a mock package which can be used to simulate a dc/os cluster, it has support for mocking cluster links too. Let me know if you need help for writing it, if we don't do it as part of this PR we'll probably never get back to it.

bamarni:

PR #1383, opened 2018-11-19
Add support for relative DCOS_DIR

bamarni:
Prior to this patch, setting a relative DCOS_DIR would cause the CLI to
fail when invoking a plugin starting from the second invocation. It
happened because the CLI prepends the DC/OS dir to a command path when
it sees it as a relative path. When a DC/OS dir is relative, that
operation would happen at every plugin invocation and grow the path
indefinitely.

This updates the DCOSDir method to prepend the current directory in case
DCOS_DIR is a relative path. It also makes it return an explicit error
in case anything goes wrong.

https://jira.mesosphere.com/browse/DCOS_OSS-4405

### DIFF
--- a/pkg/cmd/cluster/cluster_attach.go
+++ b/pkg/cmd/cluster/cluster_attach.go
@@ -17,7 +17,10 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
 		Short: "Attach the CLI to a cluster",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			manager := ctx.ConfigManager()
+			manager, err := ctx.ConfigManager()
+			if err != nil {
+				return err
+			}
 
 			// We try to find a Config matching the argument given.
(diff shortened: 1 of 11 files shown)


bamarni:

bamarni:

pkg/cmd/cluster/cluster_attach.go
```diff
@@ -17,7 +17,10 @@ func newCmdClusterAttach(ctx api.Context) *cobra.Command {
 		Short: "Attach the CLI to a cluster",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			manager := ctx.ConfigManager()
+			manager, err := ctx.ConfigManager()
```
ArmandGrillet: Why are we sometimes calling the manager `manager` and sometimes `configManager`? Could we normalize the naming?

bamarni: There is no specific reason for this, on the above I just left it as it was called.

PR #1386, opened 2018-11-23
Updated login flow to fix provider selection when uid/password given.

ArmandGrillet:
This resolves two issues seen in https://jira.mesosphere.com/browse/DCOS-45239.

First, if a user provides a uid and a password, we default to the provider
'DCOSUIDPassword'. This makes the behavior of the CLI the same as before.

Second, we have updated the error message when our prompt does not work
due to an input that we cannot parse.

### DIFF
--- a/pkg/login/flow.go
+++ b/pkg/login/flow.go
@@ -86,6 +86,8 @@ func (f *Flow) Start(flags *Flags, httpClient *httpclient.Client) (string, error
 func (f *Flow) selectProvider(providers Providers) (*Provider, error) {
 	// Explicit provider selection.
 	if f.flags.providerID != "" {
+		// TODO: If a provider ID, a username, and a password are given
+		// we should return an error. This might break some use cases.
 		provider, ok := providers[f.flags.providerID]
 		if ok {
 			return provider, nil
@@ -96,6 +98,13 @@ func (f *Flow) selectProvider(providers Providers) (*Provider, error) {
 	// Extract login provider candidates for implicit or manual selection.
 	var providerCandidates []*Provider
 	for _, provider := range providers.Slice() {
+		// If both username and password are passed, we default to dcos-uid-password.
+		// The provider does not matter in the actual request we do to the cluster,
+		// the IAM might delegate the credential validation to a directory backend via LDAP.
+		if provider.ID == DCOSUIDPassword && f.flags.username != "" && f.flags.password != "" {
+			return provider, nil
+		}
+
 		if f.flags.Supports(provider) {
 			providerCandidates = append(providerCandidates, provider)
 		} else {
--- a/pkg/prompt/prompt.go
+++ b/pkg/prompt/prompt.go
@@ -72,7 +72,7 @@ func (prompt *Prompt) Select(msg string, choices interface{}) (int, error) {
 	choice := prompt.Input(fmt.Sprintf("(%d-%d): ", 1, choicesLen))
 	i, err := strconv.Atoi(choice)
 	if err != nil {
-		return 0, err
+		return 0, errors.New("unable to parse selected input")
 	}
 	if i < 1 || i > choicesLen {
 		return 0, fmt.Errorf("choice %d doesn't exist", i)


ArmandGrillet:

pkg/login/flow.go
```diff
@@ -86,6 +86,8 @@ func (f *Flow) Start(flags *Flags, httpClient *httpclient.Client) (string, error
 func (f *Flow) selectProvider(providers Providers) (*Provider, error) {
 	// Explicit provider selection.
 	if f.flags.providerID != "" {
+		// TODO: If a provider ID, a username, and a password are given
+		// we should return an error. This might break some use cases.
```
jgehrcke: For the record:  or at least emit a warning that the provider ID is ignored / not used.

bamarni: Why would that be needed? I thought this patch was about keeping backwards compatibility with the previous CLI, which wasn't behaving like that.

bamarni:
This needs an entry in the changelog.

pkg/login/flow.go
jgehrcke: > Why would that be needed?

I discussed this with Armand, it's a good question and I am happy to write down the reasoning here.

As of today the provider is ID is meaningless in the context of a uid/password-based login (if you question this let's discuss this, but assume this to be true for the remaining paragraphs).

If the command line _interface_ however suggests to a user that it is indeed relevant (by accepting it as an option) then this is a misleading user experience. In the worst case it might mislead a user into providing credentials (username and password) to the DC/OS IAM which it should _never_ see, such as credentials for a single sign-on provider.

That is why the CLI should not accept a provider ID option when it is not _needed_. 

If this would break backwards compatibility then the second-best option is to emit a warning, maybe  deprecation warning (with later strict enforcement).

bamarni: > As of today the provider ID is meaningless in the context of a uid/password-based login.

I'd understand this if DC/OS always returned a single uid/password-based login provider in `/acs/api/v1/auth/providers`. Then the provider ID wouldn't matter because it is always unique.

My main question here is how do you expect the CLI to behave exactly when there are multiple uid/password-based login providers? Should it pick one of these randomly? Should it ignore their `start_flow_url` config and use a hardcoded one?

Also if DC/OS can return multiple uid/password-based login providers while they are effectively the same, why isn't DC/OS returning a single one?

jgehrcke: > My main question here is how do you expect the CLI to behave exactly when there are multiple uid/password-based login providers?

When using the `client-method` called `dcos-usercredential-post-receive-authtoken` then the provider ID does not matter. It should not be asked from the human user; and not be sent to the IAM.

> Also if DC/OS can return multiple uid/password-based login providers while they are effectively the same, why isn't DC/OS returning a single one?

To inform the operator that LDAP login delegation is activated. This provider is there, and it is active, that's why it is shown. I think it's totally OK that there is an additional constraint that the user agent cannot explicitly choose it.

> My main question here is how do you expect the CLI to behave exactly when there are multiple uid/password-based login providers? 

Key insight here is that the client _method_ `dcos-usercredential-post-receive-authtoken` does not allow for choosing a specific provider. `dcos-usercredential-post-receive-authtoken` effectively defines fully how the corresponding login HTTP request needs to be constructed (and from there we _know_ that the provider ID does _not_ get encoded in the login HTTP request).

That is why we can quickly appreciate that the CLI does not need to choose one; should not even choose one.

Let's consider the CLI to be a black box. For making this point really clear let's assume that I really don't care about the internals of this black box. There is an input end (human user uses it) and an output end (an HTTP request is sent to a DC/OS cluster).

What can I say about input and output?

- In the input end it is important to me that the human user does not enter a provider ID when they enter uid and password, because the provider ID is conceptually meaningless and because of the security issue I described in a comment above.
- In the output end it is important that the CLI constructs the login HTTP request as defined by the `dcos-usercredential-post-receive-authtoken` method which does not allow for encoding the provider ID In the HTTP request.

Now, based on _that_ I can say something really important about the _inner_ workings of the black box: it does absolutely not need to choose a provider ID.

If you still _want_ to choose one in this black box then I recommend taking a machine learning-based approach; or let the user play a little minigame :-).

-----------

I have just realized that we have a bug in the IAM where it would sometimes emit

```
dcos-credential-post-receive-authtoken
```
instead of

```
dcos-usercredential-post-receive-authtoken
```

`dcos-credential-post-receive-authtoken` and `dcos-usercredential-post-receive-authtoken` mean the same! We have a JIRA ticket to fix that.

PR #1387, opened 2018-11-26
Use an absolute path in TestClusterAttachBothConfiguredAndLinked

bamarni:
This updates the test to use an absolute path when creating cluster
fixtures, otherwise they are missing the DCOS_DIR prefix and the test
fails.


ArmandGrillet:
I'll backport this + the support for relative DCOS_DIR for 1.12.

PR #1388, opened 2018-11-26
[1.12] Updated login flow to detect when dcos-users provider is available.

ArmandGrillet:
In a previous commit, we used the provider ID to detect a specific
provider  available for a cluster. This was wrong, the value
`DCOSUIDPassword` we are looking for being the provider type.

We should have caught the issue with an integration test but we
had not added one due to the wish to fix the DC/OS EE CI ASAP.


ArmandGrillet:
@janisz The PR is now "ready". Please only review my PRs if you are assigned as a reviewer or if I asked for reviews explicitly on another medium, otherwise the lack of context can make us both lose time.

bamarni:
@jgehrcke : we've updated #1386 and added a test.

PR #1389, opened 2018-11-30
Add a script to generate a universe resource

bamarni:
The generate_universe_resource.py script can be used to generate
resource files for universe. It takes as argument a plugin name and a
version and uses them to download the plugins from their canonical URLs
and generate sha256 checksums accordingly.

https://jira.mesosphere.com/browse/DCOS_OSS-1797


bamarni:
@janisz : we opted for python for integration tests and CI in general, only the CLI itself and unit tests are written in Go.

PR #1391, opened 2018-12-04
Enabled bash completions to be compatible with zsh.

rgo3:


pkg/cmd/completion/completion.sh
```diff
@@ -9,6 +9,35 @@ __dcos_debug() {
     fi
 }
 
+# Since _get_comp_words_by_ref is not available in zsh we need a ported version that can be sourced
```
ArmandGrillet: The comment can be cut so that each line has roughly the same length, e.g.
```
# _get_comp_words_by_ref is not available in zsh, we use 
# a ported version that can be sourced when using zsh.
```

PR #1394, opened 2018-12-07
[0.7.x] Enabled bash completions to be compatible with zsh.

rgo3:


pkg/cmd/completion/completion.sh
```diff
@@ -9,6 +9,34 @@ __dcos_debug() {
     fi
 }
 
+# Since _get_comp_words_by_ref is not available in zsh
+# we need a ported version that can be sourced when using zsh.
+if [[ -n ${ZSH_VERSION:+set} ]] && ! type _get_comp_words_by_ref >/dev/null 2>&1; then
```
bamarni: @br-lewis had designed the CLI command to accept a shell argument, eg. `dcos completion bash`. To enable zsh completion, one should run `dcos completion zsh` and get the appropriate script, this would remove the need for runtime checks.

br-lewis: Doing a separate zsh script that gets rid of this check would result in two mostly identical scripts. I think a zsh script should be a totally different script to support native zsh completion features instead one that's entirely bash completion.

bamarni: There is no need to duplicate scripts necessarily, but my main point is that we should document `dcos completion zsh`, even though it returns the same script as bash. In the future if we get to add proper zsh completion it will then return the appropriate one instead of sticking to bash.

rgo3: @bamarni opened #1396 to make sure we avoid any confusion around this.

PR #1400, opened 2018-12-12
[0.7.x] Fixed emitting empty `uid` value in login credentials.

rgo3:
With this commit we fix that the CLI only emits the `uid` value in the
login credentials when explicitly setting it. This is needed to fix a
bug when using the `browser-prompt-oidcidtoken-get-authtoken` method.

https://jira.mesosphere.com/browse/DCOS_OSS-4599


ArmandGrillet:

PR #1411, opened 2019-01-04
Add a script verifying CLI artifacts

bamarni:
This adds a script which verifies that CLI artifacts are properly
deployed.

It scans the tags in dcos/dcos-cli and gets the latest one for each
minor version: 0.5.x, 0.6.x, and 0.7.x. It then downloads CLI artifacts
for each supported platform / DC/OS version and runs `dcos --version` on
them. The output is parsed and the script fails if `dcoscli.version`
contains an unexpected value.

It also checks the testing endpoint, which should have as a version the
latest commit of the master branch.

The "latest" endpoints are commented for now because they are not live
yet (it should happen on the next 0.7.x release).

The build can be manually run from here :

https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-cluster-ops/job/mesosphere-dcos-cli/job/verify-artifacts/


ci/verify-artifacts.py
```diff
@@ -0,0 +1,90 @@
...
+
+if platform == 'win32':
+    platform = 'windows'
+    ext = '.exe'
+
+g = Github(os.environ.get("GITHUB_TOKEN"))
+
+dcos_cli_repo = g.get_repo("dcos/dcos-cli")
+
+latest_0_5 = "0.5.0"
```
ArmandGrillet: Could we use PyGithub to get the three latest big releases and then loop over it so that we don't have to update the script once we release a CLI 0.8?

bamarni: While looking at it, I think it'd make it too complex and not readable to dynamically get the latest 3 and map them to DC/OS versions, especially given that 0.8 should be the latest version pinned to a specific DC/OS version, I've just added a block to uncomment.

PR #1417, opened 2019-01-10
[backport] Make sure the Auth token is empty when challenging auth

bamarni:
When challenging auth we must ensure the auth token is empty, otherwise
we might get a 200 and the command will fail with "Authentication
disabled".

https://jira.mesosphere.com/browse/DCOS_OSS-4660


bamarni:

PR #1431, opened 2019-01-30
Fix cluster attach when no cluster is attached

bamarni:
This is a follow-up from 9b85cc927874a51539ccc8525e084689f32369cb,
the patch only worked when no cluster is attached and there is a single
cluster configured...

This fixes the case when there are multiple clusters without any of them
being attached and one runs `dcos cluster attach <cluster>`.

https://jira.mesosphere.com/browse/DCOS-47191


ArmandGrillet:

PR #1444, opened 2019-03-03
 Make the plugin runtime safe for concurrent usage

bamarni:
It was previously reading/writing/overwriting files in-place without any
synchronization mechanism.

This commit introduces a `tmp` directory inside the cluster's dir, which
is used as a staging directory for writing files before they are
eventually moved atomically to their final location.

https://jira.mesosphere.com/browse/DCOS_OSS-4843

### DIFF
--- a/tests/integration/test_corecli.py
+++ b/tests/integration/test_corecli.py
@@ -1,9 +1,12 @@
 import json
 import os
+import sys
 
-from .common import exec_cmd, default_cluster  # noqa: F401
+from concurrent import futures
 
 import pytest
+
+from .common import exec_cmd, default_cluster  # noqa: F401
 
 
 @pytest.mark.skipif(os.environ.get('DCOS_TEST_CORECLI') is None, reason="no core CLI bundle")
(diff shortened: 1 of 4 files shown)


bamarni:

ArmandGrillet:
`test_extract_core_concurrently` is failing 😕

bamarni:
It is failing on windows because their `rename` works differently than UNIX. There seems to be packages for the windows use-case (https://github.com/natefinch/atomic), but it requires more investigation.

ArmandGrillet:
Would it make sense to not run the test on Windows? The concurrent plugin runtime wasn't safe on Windows before anyway, right?

bamarni:

bamarni:

tests/integration/test_corecli.py
```diff
@@ -30,6 +33,26 @@ def test_extract_core(default_cluster):
...
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason='Not yet concurrent-safe on Windows (DCOS_OSS-4843)')
+def test_extract_core_concurrently(default_cluster):
+    code, out, err = exec_cmd(['dcos', 'plugin', 'remove', 'dcos-core-cli'])
+    assert code == 0
+    assert out == ''
+    assert err == ''
+
+    with futures.ThreadPoolExecutor() as pool:
+        cmds = [pool.submit(exec_cmd, ['dcos', 'node']) for _ in range(50)]
```
rgo3:
```suggestion
        cmds = [pool.submit(exec_cmd, ['dcos', 'node', 'list']) for _ in range(50)]
```
I would suggest using this line, since `dcos node` is deprecated? Would require a small change to the backport.

bamarni:

bamarni:

PR #1446, opened 2019-03-05
Fix autocompletion bug for dynamic completions

rgo3:
This PR adds a function which can be used to check if a command is in a set of subcommands. This check is useful when adding dynamic completions, so that dynamic arguments will not be interpreted as a subcommand. That enables further completions for a command even if dynamic arguments have been printed on the command-line.


pkg/cmd/completion/completion.sh
```diff
@@ -9,6 +9,23 @@ __dcos_debug() {
     fi
 }
 
+
+# checks if an array of subcommands contains a subcommand.
+# usage: _contains_subcommand <arrays_of_subcommands> <cmd>
+_contains_subcommand () {
```
br-lewis: I don't think this is too important but it doesn't follow the naming scheme for helper functions. I'm not sure how common of a name this might be but I figured it's better to namespace these under `__dcos` since these functions will be global within a shell.

Also, it seems like you're not using this function anywhere.

pkg/cmd/completion/completion.sh
```diff
@@ -9,6 +9,23 @@ __dcos_debug() {
...
+    local possible_cmds="$1[@]"
```
br-lewis: I'm not sure this works. It looks like you're trying to use https://wiki.bash-hackers.org/syntax/arrays#indirection which isn't a bad idea but that's not the syntax to accomplish it. Though the syntax given there makes shellcheck real unhappy.

```
asdf() {
    local foo="$1[@]"
    echo "${foo[@]}"
}

bar=("a" "b" "c" "d")
asdf bar
```
Gives
```
❯ bash test.sh
bar[@]
```

rgo3: Yeah something is not working yet. I'll fix that 🤦‍♂️

pkg/cmd/completion/completion.sh
rgo3: Ok, I can update the naming scheme. 
Regarding the use of this function: As of now none of the functions where we have dynamic completions need this fix. Going forward we can use this function when adding more commands that support dynamic argument completion.

PR #1466, opened 2019-04-30
Improve error messages for `dcos cluster setup` args error

bamarni:
This makes it clearer that the command expects a single argument which
is a cluster URL.

https://jira.mesosphere.com/browse/DCOS-52029


ArmandGrillet:

PR #1478, opened 2019-06-12
When installing the CLI we want to use the vendored deps

makkes:
## High-level description

The `install` target didn't use the vendored dependencies but would rather download all deps to the `GOPATH`. This PR fixes this.

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

- [DCOS_OSS-<number>](https://jira.mesosphere.com/browse/DCOS_OSS-<number>) Foo the Bar so it stops Bazzing.

## Checklist for all PRs

- [x] Added a comprehensible changelog entry to `CHANGELOG.md` or explain why this is not a user-facing change: only relevant for devs
- [ ] Updated completion script if applicable
- [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: this is a change to the dev setup
- [ ] Made a [documentation PR](https://github.com/mesosphere/dcos-docs-site) if these changes need to be reflected on https://docs.mesosphere.com/latest/cli/
- [ ] Created backport PRs if needed:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ core-download:
 
 .PHONY: install
 install:
-	go install ./cmd/dcos
+	go install -mod=vendor ./cmd/dcos
 
 .PHONY: test
 test: vet
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ VERSION?=$(shell git rev-parse HEAD)
 CORE_VERSION?=1.13-patch.x
 CORE_STABILITY?=testing
 
+export GOFLAGS := -mod=vendor
+export GO111MODULE := on
+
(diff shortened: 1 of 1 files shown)


makkes:
> In another project we use `GOFLAGS` to make sure it's picked up by all relevant commands (https://github.com/dcos/dcos-core-cli/blob/4bacda1a064f701bf137323855010e80918d404b/Makefile#L9), but that's fine as is too.

Even better. Shall I introduce that here, too?

bamarni:
Yeah that'd be nice :+1:

PR #1483, opened 2019-06-27
Added 'dcos config keys' command.

ArmandGrillet:
## High-level description

Adds a `dcos config keys` subcommand and uses it for autocompletion.

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

- [DCOS-53202](https://jira.mesosphere.com/browseDCOS-53202) No dcos config keys subcommand to get all the possible keys without having to check the documentation.

## Checklist for all PRs

- [X] Added a comprehensible changelog entry to `CHANGELOG.md` or explain why this is not a user-facing change:
- [X] Updated completion script if applicable
- [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
- [ ] Made a [documentation PR](https://github.com/mesosphere/dcos-docs-site) if these changes need to be reflected on https://docs.mesosphere.com/latest/cli/
- [ ] Created backport PRs if needed:

### DIFF
--- a/pkg/cmd/config/config_keys.go
+++ b/pkg/cmd/config/config_keys.go
@@ -1,0 +1,38 @@
+package config
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-cli/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+// newCmdConfigKeys returns the keys that a user can set in a configuration file.
+func newCmdConfigKeys(ctx api.Context) *cobra.Command {
+	var quietOutput bool
+	cmd := &cobra.Command{
+		Use:   "keys",
+		Short: "Print all the keys that can be set in a configuration file",
+		Args:  cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			configKeys := config.Keys()
+			var sortedKeys []string
+			for k := range configKeys {
+				sortedKeys = append(sortedKeys, k)
+			}
+			sort.Strings(sortedKeys)
+			for _, key := range sortedKeys {
+				if quietOutput {
+					fmt.Fprintln(ctx.Out(), key)
+				} else {
+					fmt.Fprintln(ctx.Out(), key+" : "+configKeys[key])
+				}
+
+			}
+		},
+	}
+	cmd.Flags().BoolVarP(&quietOutput, "quiet", "q", false, "Only print config keys")
+	return cmd
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -96,6 +96,27 @@ func Empty() *Config {
 	return New(Opts{})
 }
 
+// Keys returns the possible config keys. TODO: make all the keys constants.
+func Keys() map[string]string {
+	return map[string]string{
+		keyACSToken:          "the DC/OS authentication token",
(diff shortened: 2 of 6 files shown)


pkg/cmd/config/config_keys.go
```diff
@@ -0,0 +1,65 @@
...
+				"core.reporting",
+				"core.prompt_login",
+				"cluster.name",
+				"job.url",
+				"job.service_name",
+				"marathon.url",
+				"package.cosmos_url",
+			}
+
+			configMap := map[string]string{
```
bamarni: Could this map be returned by the config package (`config.Keys()`)? As we already have constants there.

pkg/cmd/config/config_keys.go
```diff
@@ -0,0 +1,65 @@
...
+				"core.ssh_proxy_ip":     "whether to use a fixed ssh proxy host (Bastion) for node SSH access",
+				"core.reporting":        "whether to report usage events to Mesosphere",
+				"core.prompt_login":     "whether to prompt the user to log in when token expired, otherwise automatically initiate login",
+				"cluster.name":          "human readable name of cluster",
+				"job.url":               "API URL for talking to the Metronome scheduler",
+				"job.service_name":      "the name of the metronome cluster",
+				"marathon.url":          "base URL for talking to Marathon, overwrites the value specified in 'core.dcos_url'",
+				"package.cosmos_url":    "base URL for talking to Cosmos, overwrites the value specified in 'core.dcos_url'",
+			}
+			for _, key := range keys {
```
bamarni: you can loop through the map, that removes the need for the slice

ArmandGrillet: Ordering is not offered when looping over a map, that's why I prefer to loop over a slice so that the keys are alphabetically organized. I could order the keys at runtime to give the same output.

bamarni: Ah I see, yeah let's order at runtime from the map, that'd make it less likely that we forget a place when adding new keys. Also if it's returned from the config package it's easier to keep the list in sync.

pkg/config/config.go
```diff
@@ -96,6 +96,27 @@ func Empty() *Config {
 	return New(Opts{})
 }
 
+// Keys returns the possible config keys.
+func Keys() map[string]string {
+	return map[string]string{
+		"core.dcos_acs_token":   "the DC/OS authentication token",
```
bamarni: Ideally this would use the constants strings, to reduce the chance of having typos in the map keys.

pkg/cmd/config/config_keys.go
```diff
@@ -0,0 +1,39 @@
...
+				if quietOutput {
+					fmt.Fprintln(ctx.Out(), key)
+				} else {
+					fmt.Fprintln(ctx.Out(), key+" : "+configKeys[key])
+				}
+
+			}
+		},
+	}
+	cmd.Flags().BoolVarP(&quietOutput, "quiet", "q", false, "Does not print what a key represents")
```
bamarni:
```suggestion
	cmd.Flags().BoolVarP(&quietOutput, "quiet", "q", false, "Only print config keys")
```

ArmandGrillet:
@mesosphere-ci run all integration tests

ArmandGrillet:
@mesosphere-ci run all integration tests

PR #1485, opened 2019-08-05
Add `dcos cluster open`

bamarni:
It can be used to open the currently attached cluster UI in the browser.

As it's a very simple command I thought I'd propose it as a PR.

TODO if accepted:
- update autocompletion
- update docs


pkg/cmd/cluster/cluster_open.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+// newCmdClusterOpen opens the current cluster UI in the user browser.
+func newCmdClusterOpen(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "open",
+		Short: "Open the current cluster UI in the browser",
+		Args:  cobra.NoArgs,
```
bamarni: We could also accept an optional argument here, eg. `dcos cluster open soak114`

ArmandGrillet: Yes I think it would make sense to have this option.

bamarni:
Updated and added autocompletion

PR #1492, opened 2019-08-30
Loosen the DC/OS version check

bamarni:
Currently it only accepts DC/OS 1.XX. This uses a library instead of a
regex for more flexibility.

It also avoids false-positive by just printing a warning if we can't
parse the version for some reason. It might then still be possible for
the setup flow to install the default plugins, through Cosmos for example.


bamarni:

bamarni:

PR #1497, opened 2019-09-19
Return plugin after installation

bamarni:
This returns the installed plugin when the Install() method is called.
This is useful in some cases where the caller needs to do additional
work with a plugin once it's installed, eg. print the new subcommands.

This also updates the `dcos plugin add` command to print the new
commands whenever a plugin is installed.

### DIFF
--- a/pkg/cmd/dcos.go
+++ b/pkg/cmd/dcos.go
@@ -132,7 +132,7 @@ func updateCorePlugin(ctx api.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


bamarni:

bamarni:

bamarni:

bamarni:
@makkes : maybe you've seen this message as part of `dcos cluster setup`, though when running `dcos plugin add` it was never displayed, I thought it'd be nice to also show it there. There were no regression, it's just a slight UX improvement which I think doesn't deserve an entry in the changelog.

bamarni:

bamarni:

PR #1499, opened 2019-09-23
Pass plugin config as environment variables

bamarni:
This makes it possible for a plugin to retrieve its config values
through environment variables when it's invoked.

For example, assuming a command `foo` in a given plugin. When the
configuration file contains the `bar` config for the `foo` section,
it will be passed to it as the `DCOS_FOO_BAR` environment variable.

### DIFF
--- a/design/plugin.md
+++ b/design/plugin.md
@@ -122,6 +122,15 @@ When a cluster is attached, the CLI will also pass the following ENV variables:
                       verify server certificates against. When it is not set, the system's CA bundle
                       must be used.
 
+The CLI will also pass the subcommand config as additional environment variables with the following format:
+
+    DCOS_{COMMAND}_{CONFIG}={VALUE}
+
+Where `COMMAND` is the uppercase subcommand name, if the command contains hyphens, they are replaced by underscores.
+And `CONFIG` is the uppercase config key in the command section.
+
+For instance, when the `hello` subcommand has a config value (`dcos config set hello.world foo`), it will be passed when the subcommand is invoked (as `DCOS_HELLO_WORLD=foo`).
+
 The child process has access to the standard out/standard error of the CLI process. It is not guaranteed to have access to any other file descriptors the parent may have opened.
 
 The child process is not started in a shell. If the command is not startable for any reason an error message is printed (not on the path, executable does not exist, binary incompatibility, etc).
--- a/design/plugin.md
+++ b/design/plugin.md
@@ -112,6 +112,8 @@ For example, if a `dcos-hello` executable in a plugin has a `hello` top level co
 
 Executables contained within plugins are called synchronously from the CLI by spawning a new child process. The CLI waits for the child process to complete. The CLI makes available its own executable path to the child via an ENV variable `DCOS_CLI_EXECUTABLE_PATH`.
 
+The CLI version is passed in the `DCOS_CLI_VERSION` ENV variable.
+
 When a cluster is attached, the CLI will also pass the following ENV variables:
 
 - `DCOS_URL`: The base URL of the DC/OS cluster without a trailing slash (eg. `https://dcos.example.com`).
--- a/pkg/cmd/dcos.go
+++ b/pkg/cmd/dcos.go
@@ -25,6 +25,7 @@ import (
 	"github.com/dcos/dcos-cli/pkg/cmd/completion"
 	configcmd "github.com/dcos/dcos-cli/pkg/cmd/config"
 	plugincmd "github.com/dcos/dcos-cli/pkg/cmd/plugin"
+	"github.com/dcos/dcos-cli/pkg/config"
 	"github.com/dcos/dcos-cli/pkg/internal/cosmos"
 	"github.com/dcos/dcos-cli/pkg/plugin"
 )
@@ -151,48 +152,21 @@ func updateCorePlugin(ctx api.Context) error {
 
 // invokePlugin calls the binary of a plugin, passing in the arguments it's been given.
 func invokePlugin(ctx api.Context, cmd plugin.Command, args []string) error {
-	executablePath, err := os.Executable()
-	if err != nil {
-		return err
-	}
 	execCmd := exec.Command(cmd.Path, args...)
 	execCmd.Stdout = ctx.Out()
 	execCmd.Stderr = ctx.ErrOut()
 	execCmd.Stdin = ctx.Input()
 
-	execCmd.Env = append(os.Environ(), "DCOS_CLI_EXECUTABLE_PATH="+executablePath)
-	execCmd.Env = append(execCmd.Env, "DCOS_CLI_VERSION="+version.Version())
-
-	switch ctx.Logger().Level {
-	case logrus.DebugLevel:
-		execCmd.Env = append(execCmd.Env, "DCOS_VERBOSITY=2", "DCOS_LOG_LEVEL=debug")
-	case logrus.InfoLevel:
-		execCmd.Env = append(execCmd.Env, "DCOS_VERBOSITY=1", "DCOS_LOG_LEVEL=info")
+	executablePath, err := os.Executable()
+	if err != nil {
+		return err
 	}
-
-	// Pass cluster specific env variables when a cluster is attached.
-	if cluster, err := ctx.Cluster(); err == nil {
-		execCmd.Env = append(execCmd.Env, "DCOS_URL="+cluster.URL())
-		execCmd.Env = append(execCmd.Env, "DCOS_ACS_TOKEN="+cluster.ACSToken())
-
-		insecure := cluster.TLS().Insecure || strings.HasPrefix(cluster.URL(), "http://")
-		if insecure {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_INSECURE=1")
-		} else if cluster.TLS().RootCAsPath != "" {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_CA_PATH="+cluster.TLS().RootCAsPath)
-		}
-
-		// Create env entries based on the subcommand config.
-		if cmdConfig, ok := cluster.Config().ToMap()[cmd.Name].(map[string]interface{}); ok {
-			for key, val := range cmdConfig {
-				execCmd.Env = append(execCmd.Env, fmt.Sprintf(
-					"%s=%v",
-					cmdConfigEnvKey(cmd.Name, key),
-					val,
-				))
-			}
-		}
+	cluster, err := ctx.Cluster()
+	if err != nil && err != config.ErrNotAttached {
+		return err
 	}
+	execCmdEnv := pluginEnv(executablePath, cmd.Name, ctx.Logger().Level, cluster)
+	execCmd.Env = append(os.Environ(), execCmdEnv...)
 
 	err = execCmd.Run()
 	if err != nil {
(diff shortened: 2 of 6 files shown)


design/plugin.md
```diff
@@ -122,6 +122,10 @@ When a cluster is attached, the CLI will also pass the following ENV variables:
                       verify server certificates against. When it is not set, the system's CA bundle
                       must be used.
 
+The CLI will also pass the subcommand config as additional environment variables. For instance, when the
+`hello` subcommand has some config values (eg. `dcos config set hello.world foo`), they will be passed when
+the subcommand is invoked (eg. `DCOS_HELLO_WORLD=foo`).
```
makkes: I was thinking if we should make it a little more clear how `hello.world` is transformed to `DCOS_HELLO_WORLD`, i.e. it's transformed to all-uppercase, dots are replaced with underscores and the prefix `DCOS_` is prepended.

pkg/cmd/dcos.go
```diff
@@ -179,6 +179,21 @@ func invokePlugin(ctx api.Context, cmd plugin.Command, args []string) error {
...
+			for key, val := range cmdConfig {
+				entry := fmt.Sprintf(
+					"DCOS_%s_%s=%v",
+					strings.ToUpper(cmd.Name),
+					strings.ToUpper(key),
+					val,
+				)
+				execCmd.Env = append(execCmd.Env, entry)
+			}
+		}
```
janisz: I'd like to extract above code to a function that calculates env and add unit tests for it

```diff
diff --git a/pkg/cmd/dcos.go b/pkg/cmd/dcos.go
index ac4c967..22d2a87 100644
--- a/pkg/cmd/dcos.go
+++ b/pkg/cmd/dcos.go
@@ -159,25 +159,43 @@ func invokePlugin(ctx api.Context, cmd plugin.Command, args []string) error {
 	execCmd.Stderr = ctx.ErrOut()
 	execCmd.Stdin = ctx.Input()
 
-	execCmd.Env = append(os.Environ(), "DCOS_CLI_EXECUTABLE_PATH="+executablePath)
+	execCmd.Env = append(execCmd.Env, calculateEnv(ctx, executablePath, cmd)...)
 
+	err = execCmd.Run()
+	if err != nil {
+		// Because we're silencing errors through Cobra, we need to print this separately.
+		ctx.Logger().Debug(err)
+		// When the plugin command exits with a non-zero code, the main CLI process
+		// exit code should be the same. ExitCode() returns -1 if the process
+		// hasn't exited or was terminated by a signal thus we have to check.
+		//
+		// See https://jira.mesosphere.com/browse/DCOS_OSS-4399
+		if exitCode := execCmd.ProcessState.ExitCode(); exitCode > 0 {
+			os.Exit(exitCode)
+		}
+
+	}
+	return err
+}
+
+func calculateEnv(ctx api.Context, executablePath string, cmd plugin.Command) (env []string) {
+	env = append(os.Environ(), "DCOS_CLI_EXECUTABLE_PATH="+executablePath)
 	switch ctx.Logger().Level {
 	case logrus.DebugLevel:
-		execCmd.Env = append(execCmd.Env, "DCOS_VERBOSITY=2", "DCOS_LOG_LEVEL=debug")
+		env = append(env, "DCOS_VERBOSITY=2", "DCOS_LOG_LEVEL=debug")
 	case logrus.InfoLevel:
-		execCmd.Env = append(execCmd.Env, "DCOS_VERBOSITY=1", "DCOS_LOG_LEVEL=info")
+		env = append(env, "DCOS_VERBOSITY=1", "DCOS_LOG_LEVEL=info")
 	}
-
 	// Pass cluster specific env variables when a cluster is attached.
 	if cluster, err := ctx.Cluster(); err == nil {
-		execCmd.Env = append(execCmd.Env, "DCOS_URL="+cluster.URL())
-		execCmd.Env = append(execCmd.Env, "DCOS_ACS_TOKEN="+cluster.ACSToken())
+		env = append(env, "DCOS_URL="+cluster.URL())
+		env = append(env, "DCOS_ACS_TOKEN="+cluster.ACSToken())
 
 		insecure := cluster.TLS().Insecure || strings.HasPrefix(cluster.URL(), "http://")
 		if insecure {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_INSECURE=1")
+			env = append(env, "DCOS_TLS_INSECURE=1")
 		} else if cluster.TLS().RootCAsPath != "" {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_CA_PATH="+cluster.TLS().RootCAsPath)
+			env = append(env, "DCOS_TLS_CA_PATH="+cluster.TLS().RootCAsPath)
 		}
 
 		// Create env entries based on the subcommand config. For example if `hello.world = "foo"`
@@ -191,26 +209,11 @@ func invokePlugin(ctx api.Context, cmd plugin.Command, args []string) error {
 					strings.ToUpper(key),
 					val,
 				)
-				execCmd.Env = append(execCmd.Env, entry)
+				env = append(env, entry)
 			}
 		}
 	}
-
-	err = execCmd.Run()
-	if err != nil {
-		// Because we're silencing errors through Cobra, we need to print this separately.
-		ctx.Logger().Debug(err)
-		// When the plugin command exits with a non-zero code, the main CLI process
-		// exit code should be the same. ExitCode() returns -1 if the process
-		// hasn't exited or was terminated by a signal thus we have to check.
-		//
-		// See https://jira.mesosphere.com/browse/DCOS_OSS-4399
-		if exitCode := execCmd.ProcessState.ExitCode(); exitCode > 0 {
-			os.Exit(exitCode)
-		}
-
-	}
-	return err
+	return env
 }
 
 // help menu template that follow UX styleguide.
```

pkg/config/config.go
```diff
@@ -99,17 +99,20 @@ func Empty() *Config {
 // Keys returns the possible config keys. TODO: make all the keys constants.
```
janisz: How about solving this TODO?

bamarni: I've left a comment for the keys below who don't have constants, they shouldn't belong to this package in the first place so we didn't create constants for them.

pkg/cmd/dcos.go
bamarni: I've added unit tests in the latest commits.

pkg/cmd/dcos.go
```diff
@@ -150,36 +152,21 @@ func updateCorePlugin(ctx api.Context) error {
...
-		execCmd.Env = append(execCmd.Env, "DCOS_ACS_TOKEN="+cluster.ACSToken())
-
-		insecure := cluster.TLS().Insecure || strings.HasPrefix(cluster.URL(), "http://")
-		if insecure {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_INSECURE=1")
-		} else if cluster.TLS().RootCAsPath != "" {
-			execCmd.Env = append(execCmd.Env, "DCOS_TLS_CA_PATH="+cluster.TLS().RootCAsPath)
-		}
+	cluster, err := ctx.Cluster()
+	if err != nil && err != config.ErrNotAttached {
```
janisz: Previously we ignored this error? Why you changed this behavior?

pkg/cmd/dcos_test.go
```diff
@@ -49,3 +52,27 @@ Use "dcos [command] --help" for more information about a command.
...
+	require.Contains(t, env, "DCOS_URL=https://dcos.example.com")
+	require.Contains(t, env, "DCOS_ACS_TOKEN=abc")
+	require.Contains(t, env, "DCOS_HELLO_WORLD=foo")
+	require.NotContains(t, env, "DCOS_HALLO_WORLD=foo")
+}
+
+func TestCmdConfigEnvKey(t *testing.T) {
+	require.Equal(t, "DCOS_HELLO_WORLD", cmdConfigEnvKey("hello", "world"))
+	require.Equal(t, "DCOS_HELLO_WORLD_FOO", cmdConfigEnvKey("hello-world", "foo"))
+	require.Equal(t, "DCOS_HELLO_AROUND_THE_WORLD", cmdConfigEnvKey("hello", "around_the_world"))
```
janisz: How about parametrized test, I'm not strong on this since this test looks simple enough but @drewvanstone loves parametrized tests :)

```go
func TestCmdConfigEnvKey(t *testing.T) {
	testCases := []struct {
		expected  string
		cmdName  string
		configKey string
	}{
		{"DCOS_HELLO_WORLD", "hello", "world"},
		{"DCOS_HELLO_WORLD_FOO", "hello-world", "foo"},
		{"DCOS_HELLO.WORLD_FOO", "hello.world", "foo"},
		{"DCOS_HELLO_AROUND_THE_WORLD", "hello", "around_the_world"},
	}
	for _, tc := range testCases {
		t.Run(tc.expected, func(t *testing.T) {
			assert.Equal(t, tc.expected, cmdConfigEnvKey(tc.cmdName, tc.configKey))
		})
	}
}
```

pkg/cmd/dcos_test.go
```diff
@@ -49,3 +52,27 @@ Use "dcos [command] --help" for more information about a command.
...
+
+	env := pluginEnv("/path/to/me", "hello", logrus.DebugLevel, cluster)
+
+	require.Contains(t, env, "DCOS_CLI_EXECUTABLE_PATH=/path/to/me")
+	require.Contains(t, env, "DCOS_CLI_VERSION="+version.Version())
+	require.Contains(t, env, "DCOS_VERBOSITY=2")
+	require.Contains(t, env, "DCOS_URL=https://dcos.example.com")
+	require.Contains(t, env, "DCOS_ACS_TOKEN=abc")
+	require.Contains(t, env, "DCOS_HELLO_WORLD=foo")
+	require.NotContains(t, env, "DCOS_HALLO_WORLD=foo")
```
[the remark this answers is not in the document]
bamarni: I've used ElementsMatch instead of Contains. Regarding `assert` vs `require`, we currently use `require` everywhere on the goroutine running a test. If we want to change this it should be done globally, though assert should be used with care, always making sure the test won't panic afterwards.

pkg/cmd/dcos.go
bamarni: This shouldn't happen in practice but I think it's a valid error to catch.

design/plugin.md
```diff
@@ -122,6 +124,15 @@ When a cluster is attached, the CLI will also pass the following ENV variables:
...
                       must be used.
 
+The CLI will also pass the subcommand config as additional environment variables with the following format:
+
+    DCOS_{COMMAND}_{CONFIG}={VALUE}
+
+Where `COMMAND` is the uppercase subcommand name, if the command contains hyphens, they are replaced by underscores.
+And `CONFIG` is the uppercase config key in the command section.
+
+For instance, when the `hello` subcommand has a config value (`dcos config set hello.world foo`), it will be passed when the subcommand is invoked (as `DCOS_HELLO_WORLD=foo`).
```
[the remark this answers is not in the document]
bamarni: It's tested with `require.Equal(t, "DCOS_HELLO_WORLD", cmdConfigEnvKey("hello", "world"))`

pkg/cmd/dcos_test.go
bamarni: We have test tables in some unit tests, I just thought it's overkill in this case.

design/plugin.md
bamarni: I've updated the doc with more information.

pkg/cmd/dcos_test.go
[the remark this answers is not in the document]
bamarni: This should be changed globally at the same time for consistency, see for example https://github.com/dcos/dcos-core-cli/pull/364/commits/ba52e5e3bf57fc9850ae536644aaaa6fb8378810

bamarni:

PR #1502, opened 2019-10-11
Discard the local login server on non-desktop environment

bamarni:
For the `dcos-oidc-auth0` login provider, when the user operates the CLI
from a container or a remote machine, the local webserver will not run
on the same machine as the browser. Later on the browser will try to
send the token to eg. `localhost:43383`, which won't be reachable.

When this situation arises, we had designed the UI to fallback to
printing the login token for copy-pasting, but once the CLI has
successfully initiated the local web server, it doesn't accept input
from STDIN at all, so the user has no chance to copy-paste the token.

This commit updates the CLI to fallback to reading from STDIN when the
browser can't be opened.

Note: as an alternative, it would have been possible to concurrently
wait for data on both the local web server and STDIN, but the current
fix is simpler and covers the container / remote use-cases as expected.

https://jira.mesosphere.com/browse/DCOS_OSS-5591

### DIFF
--- a/pkg/login/flow.go
+++ b/pkg/login/flow.go
@@ -167,29 +167,39 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
 				if provider.ID == DCOSOIDCAuth0 {
 					loginServer, err = loginserver.New(startFlowURL)
 					if err != nil {
-						return "", err
+						f.logger.Error(err)
+					} else {
+						startFlowURL = loginServer.StartFlowURL()
+						go loginServer.Start()
+						defer loginServer.Close()
 					}
-					startFlowURL = loginServer.StartFlowURL()
-					go loginServer.Start()
 				}
 
 				if err := f.openBrowser(startFlowURL); err != nil {
-					return "", err
-				}
+					// Being unable to open the browser is not critical, a message was prompted
+					// to the user with the URL they should go to in order to start the login flow.
+					f.logger.Info(err)
 
-				if loginServer != nil {
+					// However we close and don't use the web server in this scenario, as the CLI might
+					// be running on a non-desktop environment so the we server would be stuck waiting.
+					// Falling back to reading from STDIN is safer here.
+					//
+					// See https://jira.mesosphere.com/browse/DCOS_OSS-5591
+					loginServer.Close()
+				} else if loginServer != nil {
 					token = <-loginServer.Token()
-				}
 
-				if token == "" {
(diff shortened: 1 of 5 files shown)


pkg/open/open.go
```diff
@@ -21,3 +21,11 @@ func NewOsOpener(logger *logrus.Logger) *OsOpener {
 		logger: logger,
 	}
 }
+
+// OpenerFunc is a func adapter for the Opener interface.
+type OpenerFunc func(string) error
```
janisz: This is used only in test https://github.com/dcos/dcos-cli/pull/1502/files#diff-48a7e37bb2ef223e6225547541997ea1R58 so it can be placed in `_test.go` file

pkg/login/flow.go
```diff
@@ -167,29 +167,39 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
...
+					f.logger.Info(err)
 
-				if loginServer != nil {
+					// However we close and don't use the web server in this scenario, as the CLI might
+					// be running on a non-desktop environment so the we server would be stuck waiting.
+					// Falling back to reading from STDIN is safer here.
+					//
+					// See https://jira.mesosphere.com/browse/DCOS_OSS-5591
+					loginServer.Close()
+				} else if loginServer != nil {
```
janisz: Do we have a test for this condition?

pkg/login/flow.go
```diff
@@ -167,29 +167,39 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
 				if provider.ID == DCOSOIDCAuth0 {
 					loginServer, err = loginserver.New(startFlowURL)
 					if err != nil {
-						return "", err
+						f.logger.Error(err)
```
janisz: Previously we returned at this point. Why this change?

bamarni: This lets use fallback to reading from stdin in that case.

pkg/login/flow.go
bamarni: Not right now, the login server is tested individually already, I'll see if I can add coverage here as well.

pkg/open/open.go
bamarni: Even though it is only used in a test right now, it is meant to be exposed to other projects using this package (to provide a custom open mechanism, or for their own tests).

pkg/login/flow.go
bamarni: I've added a test

bamarni:

PR #1503, opened 2019-10-17
Add timeout option for cluster setup

janisz:
This option will be saved in cluster `dcos.toml` and used by `ctx.HTTPClient(...)` to create http client.

See: https://github.com/dcos/dcos-cli/blob/b4a7d91628ecec93c4f4dc4ab7cb9ae7b4ccb29f/pkg/cli/context.go#L156-L175

### DIFF
--- a/constants/timeouts.go
+++ b/constants/timeouts.go
@@ -1,0 +1,6 @@
+package constants
+
+import "time"
+
+const HTTPTimeout = 3 * time.Minute
+const DialTimeout = 10 * time.Second
--- a/pkg/cmd/cluster/cluster_setup.go
+++ b/pkg/cmd/cluster/cluster_setup.go
@@ -1,9 +1,6 @@
 package cluster
 
 import (
-	"errors"
-	"fmt"
-
 	"github.com/dcos/dcos-cli/api"
 	"github.com/dcos/dcos-cli/pkg/setup"
 
@@ -16,15 +13,8 @@ func newCmdClusterSetup(ctx api.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "setup <url>",
 		Short: "Set up the CLI to communicate with a cluster",
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
-				cmd.SilenceUsage = false
-				return errors.New("missing cluster URL")
-			}
-			if len(args) > 1 {
-				cmd.SilenceUsage = false
-				return fmt.Errorf("received %d arguments %s, expects a single cluster URL", len(args), args)
-			}
 			clusterURL := args[0]
 			_, err := ctx.Setup(setupFlags, clusterURL, true)
 			return err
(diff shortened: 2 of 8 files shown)


pkg/setup/flags.go
```diff
@@ -52,6 +54,12 @@ func (f *Flags) Register(flags *pflag.FlagSet) {
 		"",
 		"Specify a custom name for the cluster.",
 	)
+	flags.DurationVar(
+		&f.timeout,
+		"timeout",
+		5 * time.Second,
+		"Specify HTTP timeout (default 5s).",
```
drewvanstone: Probably should say something like, `"Specify HTTP timeout in seconds (default 5 seconds)." Think that would be a little clearer that this option is in seconds

janisz: It's not limited to seconds only it could be any go duration string

pkg/setup/flags.go
```diff
@@ -16,6 +16,7 @@ type Flags struct {
 	insecure     bool
 	loginFlags   *login.Flags
 	fs           afero.Fs
+	noTimeout    bool
```
makkes: I'd rather call this `noTimeouts` (plural) to make it clear to the user that this flag unsets all of the application-level timeout values.

pkg/setup/flags.go
```diff
@@ -52,6 +53,12 @@ func (f *Flags) Register(flags *pflag.FlagSet) {
 		"",
 		"Specify a custom name for the cluster.",
 	)
+	flags.BoolVar(
+		&f.noTimeout,
+		"no-timeout",
+		false,
+		"Specify if HTTP timeout should be disabled. Useful for slow remote connections.",
```
drewvanstone: 👍

constants/timeouts.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
+import "time"
+
+const HTTPTimeout = 3 * time.Minute
+const DialTimeout = 10 * time.Second
```
bamarni: As these constants are some defaults for the `httpclient` package, they'd fit better in the package itself.

pkg/cmd/cluster/cluster_setup.go
```diff
@@ -16,15 +13,8 @@ func newCmdClusterSetup(ctx api.Context) *cobra.Command {
...
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
-				cmd.SilenceUsage = false
-				return errors.New("missing cluster URL")
-			}
-			if len(args) > 1 {
-				cmd.SilenceUsage = false
-				return fmt.Errorf("received %d arguments %s, expects a single cluster URL", len(args), args)
-			}
```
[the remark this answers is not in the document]
janisz: Fix is here https://github.com/dcos/dcos-cli/pull/1505/files#diff-3a414621eefcabda78980d53f77e5737R18-R26

constants/timeouts.go
janisz: We can't put them there to prevent circular dependecy
```go
constants/timeouts.go: const HTTPTimeout = 3 * time.Minute
pkg/config/cluster.go: return constants.HTTPTimeout
pkg/httpclient/client.go: Timeout: constants.HTTPTimeout,
pkg/setup/setup.go: cluster.SetTimeout(constants.HTTPTimeout)

```

PR #1507, opened 2019-10-23
Reduce function complexity limit

bamarni:
This fixes a regression introduced in #1502, which can cause a
panic on non-desktop environments with browser-based login flows
other than auth0.

https://jira.mesosphere.com/browse/DCOS-60349

### DIFF
--- a/pkg/login/flow.go
+++ b/pkg/login/flow.go
@@ -32,6 +32,7 @@ type Flow struct {
 	prompt      *prompt.Prompt
 	logger      *logrus.Logger
 	opener      open.Opener
+	loginServer *loginserver.LoginServer
 	flags       *Flags
 	interactive bool
 }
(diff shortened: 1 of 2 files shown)


bamarni:

pkg/login/flow_test.go
```diff
@@ -97,34 +97,49 @@ func TestTriggerMethodOIDCWithLoginServer(t *testing.T) {
 
 func TestTriggerMethodOIDCWithoutLoginServer(t *testing.T) {
 
-	// We're simulating a container environment where the browser can't be opened.
-	// See https://jira.mesosphere.com/browse/DCOS_OSS-5591
-	opener := open.OpenerFunc(func(_ string) error {
-		return errors.New("couldn't open browser, I'm just a container")
-	})
+	fixtures := []struct {
```
drewvanstone: As a matter of convention, I've generally seen this struct be called `tests`, and then use `tt` or `tc` as the singular in the for range loop. Python prefers fixtures, but in Go it generally is a `tests` struct.

pkg/login/flow_test.go
```diff
@@ -97,34 +97,49 @@ func TestTriggerMethodOIDCWithLoginServer(t *testing.T) {
...
+			// https://jira.mesosphere.com/browse/DCOS-60349
+			shibbolethLoginProvider(),
+		},
+	}
 
-	expectedLoginToken := "dummy_login_token"
-	expectedACSToken := "dummy_acs_token"
+	for _, fixture := range fixtures {
+		expectedLoginToken := "dummy_login_token"
+		expectedACSToken := "dummy_acs_token"
```
drewvanstone: These can probably be added to the the above `fixtures` / `tests` struct. Makes it easier to add additional tests and modify them in the future.

pkg/login/flow.go
```diff
@@ -157,46 +158,20 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
...
-					// Thus we raise the attempt counter to 3.
-					attempt = 3
-				}
+				f.initBrowserFlow(provider)
 			}
 
-			if token == "" {
+			if f.loginServer != nil {
+				token = <-f.loginServer.Token()
+				f.loginServer.Close()
```
janisz: What happen if we call `close` twice?

pkg/login/flow.go
```diff
@@ -157,46 +158,20 @@ func (f *Flow) triggerMethod(provider *Provider) (acsToken string, err error) {
...
+
+				// When the token comes from the local webserver there is no need to retry. In case of an
+				// error, retrying isn't needed (it can't be a typo, no copy-pasting was involved).
+				// Thus we raise the attempt counter to 3.
+				attempt = 3
```
janisz: This is weird. I'd prefer break here

janisz: OK, now I got it. This is a way to not repeat loop, not to break it. Am I correct?

janisz: Do we have a test for it?

bamarni:
@makkes : I've updated the commit, it was a nil pointer dereference.

pkg/login/flow.go
bamarni: I don't think this is currently tested.

janisz:
> @janisz : indeed it'd be nice to move some code to other methods as it grew quite big. I've started to move some part of it. As the loginWithBrowser method should not attempt to login multiple times in case it uses the login server and the token is incorrect, it cannot be extracted as in your diff.

There is no test or comment explaining it :(

bamarni:
> There is no test or comment explaining it :(

https://github.com/dcos/dcos-cli/blob/253a400266d1d5816b2e4524046fc691fd302033/pkg/login/flow.go#L192-L194

janisz:
> there is no need to retry
But still it we can retry.

bamarni:
I'll follow-up in another PR

PR #1508, opened 2019-10-24
Fix nil pointer dereference

bamarni:
This fixes a regression introduced in #1502, which can cause a
panic on non-desktop environments with browser-based login flows
other than auth0.

https://jira.mesosphere.com/browse/DCOS-60349

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -35,7 +35,7 @@ enable = [
 line-length = 150
 
 [linters-settings.gocyclo]
-min-complexity = 22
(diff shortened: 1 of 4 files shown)


.golangci.toml
```diff
@@ -35,7 +35,7 @@ enable = [
 line-length = 150
 
 [linters-settings.gocyclo]
-min-complexity = 22
+min-complexity = 23
```
janisz: :-1:

bamarni:

PR #1513, opened 2019-11-20
Check TLS config when creating HTTP Client

janisz:
We do not check if TLS is set up properly. When we can't read certificate or it was invalid we silently switched to insecure mode. This PR changes this behavior. Whenever HTTP Client is created it
checks for errors and returns one if necessary. The only exception are `list` and `setup` commands. List will allow user to list all clusters but call to attach will fail. `setup` uses `insecureHttpClient` by default – to download certs.

https://github.com/dcos/dcos-cli/blob/51eee402180b99bdf56ebe8356963b98807b8861/pkg/setup/setup.go#L251-L254

Fixes: https://jira.mesosphere.com/browse/DCOS-61005

### DIFF
--- a/cmd/dcos/main.go
+++ b/cmd/dcos/main.go
@@ -30,9 +30,7 @@ func main() {
 				"True (will then use CA certificates from certifi), " +
 				"or False (will then send insecure requests).\n"
 			fmt.Fprint(env.ErrOut, msg)
-			os.Exit(1)
 		}
-		fmt.Fprintf(env.ErrOut, "ERROR: %s\n", err)
 		os.Exit(1)
 	}
 }
--- a/cmd/dcos/main.go
+++ b/cmd/dcos/main.go
@@ -18,8 +18,11 @@ import (
 )
 
 func main() {
-	if err := run(cli.NewOsEnvironment()); err != nil {
+	env := cli.NewOsEnvironment()
+	if err := run(env); err != nil {
+		fmt.Fprintf(env.ErrOut, "ERROR: %s\n", err)
 		os.Exit(1)
+
 	}
 }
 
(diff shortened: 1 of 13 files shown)


cmd/dcos/main.go
```diff
@@ -18,8 +18,11 @@ import (
 )
 
 func main() {
-	if err := run(cli.NewOsEnvironment()); err != nil {
+	env := cli.NewOsEnvironment()
+	if err := run(env); err != nil {
+		fmt.Fprintf(env.ErrOut, "ERROR: %sz", err)
```
makkes:
```suggestion
		fmt.Fprintf(env.ErrOut, "ERROR: %v", err)
```
or
```suggestion
		fmt.Fprintf(env.ErrOut, "ERROR: %s", err)
```

pkg/config/cluster.go
```diff
@@ -69,7 +79,7 @@ func (c *Cluster) TLS() TLS {
 		return TLS{
 			Insecure:    true,
 			RootCAsPath: tlsVal,
-		}
+		}, fmt.Errorf("can't read %s: %s", tlsVal, err)
```
ArmandGrillet: This error starts with "can't" but the one below starts with "cannot", let's have both of them starting with "cannot" as it's the most common form in our error messages.

janisz: Fixed it with specific error. User will get clear message with instruction what to do

janisz:
:wave: @drewvanstone @br-lewis @makkes @bamarni @rgo3  Can you review, please?

cmd/dcos/main.go
```diff
@@ -12,13 +12,25 @@ import (
...
-	if err := run(cli.NewOsEnvironment()); err != nil {
+	env := cli.NewOsEnvironment()
+	err := run(env)
+	if err != nil {
+		if _, ok := err.(*config.SSLError); ok {
+			msg := "Error: An SSL error occurred. To configure your SSL settings, please " +
+				"run: 'dcos config set core.ssl_verify <value>'\n" +
+				"<value>: Whether to verify SSL certs for HTTPS or path to certs. " +
+				"Valid values are a path to a CA_BUNDLE, " +
+				"True (will then use CA certificates from certifi), " +
```
bamarni: certifi is a library that was used in the python codebase, it does not apply anymore.

makkes: Oops, I just merged that. @janisz care to fix the message in a new PR?

janisz: https://github.com/dcos/dcos-cli/pull/1514

PR #1536, opened 2020-06-09
Pin golangci-lint version

janisz:
We need to pin linter version to prevent failures introduced by changes in in
See https://ci.appveyor.com/project/janisz/dcos-cli/builds/33406479


janisz:
Fixed by https://github.com/dcos/dcos-cli/pull/1542

PR #1549, opened 2020-07-20
Skip authorization on nightly build

janisz:
Refs: https://github.com/dcos/dcos-core-cli/blame/5edeb7e8f5add1311c8f5c6d312c977e1e1de4f6/ci/build.groovy#L13-L21

### DIFF
--- a/ci/integration-tests.groovy
+++ b/ci/integration-tests.groovy
@@ -38,7 +38,7 @@ pipeline {
       }
 
       steps {
-        user_is_authorized([] as String[], '8b793652-f26a-422f-a9ba-0d1e47eb9d89', '#dcos-cli-ci')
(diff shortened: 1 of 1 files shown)


pierrebeitz:
https://github.com/mesosphere/sec-ci-libs/pull/6/files#diff-04c6e90faac2675aa89e2176d2eec7d8R136-R138 says: 

> It accepts the following, positional arguments:
> 1) A list of branches for which all all CI runs are permitted by default (e.g. `master`).

wouldn't that mean that this change *introduces* checking (instead of skipping something)?

janisz:
@pierrebeitz How about merging this to check if build will run over the weekend. If not we will revert this commit?

PR #1550, opened 2020-08-14
Accept ACS token passed in env to auth login

janisz:
`dcos cluster setup` accepts ACS token set in `DCOS_CLUSTER_SETUP_ACS_TOKEN`
environment variable. This PR will add the same behavior to `dcos auth login`.

Fixes: https://jira.d2iq.com/browse/COPS-6421


pkg/cmd/auth/auth_login.go
```diff
@@ -22,9 +22,13 @@ func newCmdAuthLogin(ctx api.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			acsToken, err := ctx.Login(flags, httpClient)
-			if err != nil {
-				return err
+			// Login to get the ACS token, unless it is already present as an env var.
+			acsToken, _ := ctx.EnvLookup("DCOS_CLUSTER_SETUP_ACS_TOKEN")
+			if acsToken == "" {
```
pierrebeitz, line 27: i think it would be more semantically correct to use the second return from the line above for that check.

i've seen this written like the following in some of our codebases: 

```
if acsToken, ok := ctx.EnvLookup("DCOS_CLUSTER_SETUP_ACS_TOKEN"); !ok {
	acsToken, err = ctx.Login(flags, httpClient)
	if err != nil {
		return err
	}
}
```

what do you think of this style? (i'm not a go-person, so i ask those kinds of dumb questions :D )

pierrebeitz (approved):
👍

pkg/cmd/auth/auth_login.go
janisz: I moved `acsToken, _ := ctx.EnvLookup("DCOS_CLUSTER_SETUP_ACS_TOKEN")` out from if condition because I need `acsToken` later. Variable declaration in condition is scoped only to this condition and it's not accessible after it.

pierrebeitz: argh! today a go-noob learned. 👍

PR #1552, opened 2020-08-19
Update Go to 1.15

janisz:
Currently DC/OS CLI keeps backward compatibility for macOS 10.10 Yosemite (2014) after updating Go version we will drop it and support only macOS 10.12 Sierra or later.

It's now the industry standard to deprecate TLS 1.0 and 1.1.
See: https://tools.ietf.org/html/draft-ietf-tls-oldversions-deprecate-00


janisz:
This is blocked by https://github.com/golang/go/issues/40893

Makefile
```diff
@@ -62,7 +62,8 @@ else
   define inDocker
     docker run \
       -e GOFLAGS \
-      -e GO111MODULE \
+      -e GOFLAGS -e GO111MODULE -e HOME=/tmp \
```
pierrebeitz, line 65: duplicate GOFLAGS?

pierrebeitz (approved):

Makefile
janisz: Good catch

pierrebeitz (approved):

